PR #16, opened 2025-12-27
feat: add static compiled assets generation for cross-platform builds

flintttan:
## Summary
- Add generate-compiled-assets.ts script to generate platform-specific asset imports
- Integrate asset generation into build-executable.ts workflow
- Update assets.ts to use compiled static assets instead of bun:bundle runtime dependency
- Fix Dockerfile to use full bun image instead of slim variant

## Problem
This change solves the issue where cross-platform compilation fails due to runtime dependencies on the `bun:bundle` module, which is not available in compiled executables.

## Solution
Generate static asset files at build time, avoiding runtime dependencies and enabling proper cross-platform compilation.

## Test plan
- [x] Build executable for current platform
- [x] Verify assets are properly generated
- [x] Test compiled executable runs without errors
- [x] Verify Docker image builds successfully

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,4 +1,4 @@
-FROM oven/bun:1.3.4 AS deps
+FROM oven/bun:1.3.5 AS deps
 
 WORKDIR /home/bun/app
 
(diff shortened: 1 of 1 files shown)


tiann:
I believe PR is unnecessary now. I've resolved the Docker issue. If you have any other concerns, please feel free to open a new one.

PR #126, opened 2026-01-30
fix(windows): enhance claude executable detection and spawn strategy

KaolaMiao:
- Improve automatic path detection for Claude CLI on Windows
- Optimize spawn logic: use 'node' for JS, no-shell for EXE, shell for CMD
- Fix shell quoting for hooks on Windows
- Suppress misleading UI errors during session switching
- Add PWA compatibility tags

### DIFF
--- a/docs/tools/bun-gemini-optimization/install-bun-gemini.ps1
+++ b/docs/tools/bun-gemini-optimization/install-bun-gemini.ps1
@@ -17,8 +17,11 @@ if ($null -ne $bunCmd) {
 }
 
 $geminiCmd = Get-Command gemini -ErrorAction SilentlyContinue
+$originalGemini = if ($geminiCmd) { $geminiCmd.Source } else { $null }
+
 if ($null -ne $geminiCmd) {
     Write-Host "  Gemini CLI 已安装" -ForegroundColor Green
+    Write-Host "  原始路径: $originalGemini" -ForegroundColor Gray
     try {
         Write-Host "  测试当前启动时间..." -ForegroundColor Cyan
         $currentTime = Measure-Command { & gemini --experimental-acp --help 2>$null | Out-Null }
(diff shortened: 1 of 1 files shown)


cli/src/modules/common/hooks/generateHookSettings.ts
```diff
@@ -56,9 +61,9 @@ function buildHookSettings(command: string, hooksEnabled?: boolean): HookSetting
         ]
     };
 
-    if (hooksEnabled !== undefined) {
-        hooks.enabled = hooksEnabled;
-    }
+    // if (hooksEnabled !== undefined) {
```
tiann: Is the intentional?

KaolaMiao:
Hi @tiann,

This is my first open source contribution, so please let me know if I've done anything incorrectly.

This PR fixes Claude Code detection on Windows by:
- Adding PowerShell fallback when The syntax of this command is:

WHERE [/R dir] [/Q] [/F] [/T] pattern...

Description:
    Displays the location of files that match the search pattern.
    By default, the search is done along the current directory and
    in the paths specified by the PATH environment variable.

Parameter List:
    /R       Recursively searches and displays the files that match the
             given pattern starting from the specified directory.

    /Q       Returns only the exit code, without displaying the list
             of matched files. (Quiet mode)

    /F       Displays the matched filename in double quotes.

    /T       Displays the file size, last modified date and time for all
             matched files.

    pattern  Specifies the search pattern for the files to match.
             Wildcards * and ? can be used in the pattern. The
             "$env:pattern" and "path:pattern" formats can also be
             specified, where "env" is an environment variable and
             the search is done in the specified paths of the "env"
             environment variable. These formats should not be used
             with /R. The search is also done by appending the
             extensions of the PATHEXT variable to the pattern.

     /?      Displays this help message.

  NOTE: The tool returns an error level of 0 if the search is
        successful, of 1 if the search is unsuccessful and
        of 2 for failures or errors.

Examples:
    WHERE /?
    WHERE myfilename1 myfile????.*
    WHERE $windir:*.* 
    WHERE /R c:\windows *.exe *.dll *.bat  
    WHERE /Q ??.??? 
    WHERE "c:\windows;c:\windows\system32:*.dll"
    WHERE /F /T *.dll  command fails
- Handling Windows-specific executable paths
- Improving error messages for Windows users

Tested on Windows 10/11 with both CMD and PowerShell.

If the changes are too large or need to be split, I'm happy to adjust. Just let me know what works best.

Thanks!

KaolaMiao:
Closing this PR as it has been superseded by PR #138, which contains only the Windows-related improvements without the Gemini Bun optimization features.

PR #155, opened 2026-02-03
feat(cli): support slash commands from installed plugins

Fdcode6:
This PR adds support for slash commands provided by installed plugins (e.g. Claude plugins). It automatically scans the ~/.claude/plugins/installed_plugins.json file, registers commands with their plugin namespace prefix, and ensures the web UI correctly renders them. Verified working with the superpowers plugin.


tiann:
@tiann review again

PR #246, opened 2026-03-05
fix(web): stabilize slash autocomplete in new sessions

jlovec:
## Summary
- Fix slash autocomplete race in newly opened sessions where initial remote command fetch may fail
- Add slash-context compensation trigger so recovery does not rely on leaving/re-entering slash mode
- Keep in-flight dedupe and cooldown protections to avoid refetch storms
- Scope slash command cache key by sessionId + agentType
- Add tests for merge filtering and cooldown boundary behavior in useSlashCommands

## Test plan
- [x] bun run --cwd web test
- [x] bun run --cwd web typecheck
- [ ] Manual: verify first `/` in a newly opened session can recover after initial fetch failure
- [ ] Manual: verify slash suggestions remain stable across session/flavor switches


jlovec1024:
按功能拆分后已由新 PR 承接：#255（Web 斜杠自动补全）与 #254（spawn 稳定性修复）。为保持审查范围清晰，此 PR 关闭。

Issue #272, opened 2026-03-10
Support queued pending messages while a session is still running


webwww123:
## Summary
Please support a **pending / queued send** workflow similar to Codex / Claude Code:
- if I send a new message while the agent is still working on the previous turn,
- the new message should be queued and automatically sent after the current turn completes.

## Actual behavior
The web UI currently blocks extra sends while a send is already pending / resolving.

## Expected behavior
- Allow multiple user messages to be queued in order.
- Show that they are queued / pending.
- Automatically flush the queue when the previous turn is done.

## Why this matters
This is very useful when supervising long-running coding sessions. I often know the next instruction before the current response finishes.

PR #274, opened 2026-03-10
feat(web): queue user messages while a session is running

webwww123:
## Summary
- allow users to queue additional messages in the web UI while a session is still running
- render queued user messages immediately with a `Queued` status
- automatically send the next queued message after the current run is no longer active

## Why
Issue #272 asks for Codex / Claude Code style pending sends, where I can type the next instruction before the current run finishes.

## Scope
This PR is intentionally web-only:
- no hub protocol changes
- no production deployment changes
- queue state lives in the current web session

## Validation
- `docker run --rm -v /root/sandbox/hapi-pr:/workspace -w /workspace oven/bun:1 sh -lc "bun install && cd web && bun run typecheck"`

### DIFF
--- a/web/src/components/AssistantChat/HappyComposer.tsx
+++ b/web/src/components/AssistantChat/HappyComposer.tsx
@@ -107,7 +107,7 @@ export function HappyComposer(props: {
         const path = (attachment as { path?: string }).path
         return typeof path === 'string' && path.length > 0
     })
-    const canSend = (hasText || hasAttachments) && attachmentsReady && !controlsDisabled && !threadIsRunning
+    const canSend = (hasText || hasAttachments) && attachmentsReady && !controlsDisabled
 
     const [inputState, setInputState] = useState<TextInputState>({
         text: '',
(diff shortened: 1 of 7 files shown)


hqhq1025:
Hey @webwww123 👋 Are you still working on this? We're pretty interested in message queuing support (ref #428) — the backend already handles it via `MessageQueue2`, so it's really just the web UI that needs the queue state display.

Your approach looks solid — the optimistic `queued` status + auto-dequeue on idle is exactly what's needed. A couple of thoughts:

1. The dequeue trigger uses `session.thinking` — but since the CLI-side `MessageQueue2` already batches queued messages, we could also just send immediately and let the backend queue handle ordering. The UI would still show the `Queued` badge for user feedback.
2. The PR currently holds messages client-side until the session goes idle. An alternative is to POST them right away (the hub accepts messages at any time) and only use the `queued` status as a visual indicator.

Let us know if you plan to continue this — happy to help review or pick it up if you're busy. Thanks!

PR #328, opened 2026-03-20
Fix Codex default approval policy for remote sessions

### LINKED ISSUE #327: Codex remote default mode cannot request sandbox escalation approvals in app
## Summary
In Codex remote sessions, `default` mode currently maps to `workspace-write + approvalPolicy=untrusted`. In practice, that means commands that need elevation do not flow cleanly into HAPI's existing in-app approval UI, which is especially visible with `gpt-5.4` running inside the sandbox.

## Expected
`default` mode should preserve the workspace-write sandbox but still allow the app to present an approval request when Codex needs elevated execution.

## Actual
The session stays sandboxed, but the user cannot complete the escalation flow from the app.

## Proposed fix
Map Codex remote `default` mode to `workspace-write + approvalPolicy=on-request` so the existing approval UI can authorize the escalation when needed.

## Notes
`read-only`, `safe-yolo`, and `yolo` semantics should remain unchanged.

pppobear:
## Summary
- switch Codex remote `default` mode from `approvalPolicy=untrusted` to `approvalPolicy=on-request`
- keep the existing `workspace-write` sandbox for default mode
- add regression coverage for the permission mode mapping and thread-start config

Closes #327

## Why
In the app, Codex remote sessions need to be able to surface sandbox escalation requests through HAPI's approval UI. With the current default mapping, `gpt-5.4` running in the sandbox can get stuck without a usable in-app escalation path.

## Testing
- `cd cli && bun run test -- src/codex/utils/permissionModeConfig.test.ts src/codex/utils/appServerConfig.test.ts src/codex/utils/permissionHandler.test.ts src/codex/codexLocalLauncher.test.ts`
- `cd cli && bunx vitest run src/claude/utils/sessionScanner.test.ts src/claude/utils/startHookServer.test.ts src/opencode/utils/startOpencodeHookServer.test.ts src/agent/backends/acp/AcpSdkBackend.test.ts`

Note: a full `cd cli && bun run test` still has pre-existing failures in `src/claude/utils/sessionScanner.test.ts` and `src/runner/runner.integration.test.ts` that are unrelated to this change.

### DIFF
--- a/cli/src/codex/utils/appServerConfig.test.ts
+++ b/cli/src/codex/utils/appServerConfig.test.ts
@@ -27,6 +27,17 @@ describe('appServerConfig', () => {
         });
     });
 
+    it('uses on-request approvals for default Codex threads', () => {
+        const params = buildThreadStartParams({
+            cwd: '/workspace/project',
+            mode: { permissionMode: 'default', collaborationMode: 'default' },
+            mcpServers
+        });
+
+        expect(params.sandbox).toBe('workspace-write');
+        expect(params.approvalPolicy).toBe('on-request');
+    });
+
     it('ignores CLI overrides when permission mode is not default', () => {
         const params = buildThreadStartParams({
             cwd: '/workspace/project',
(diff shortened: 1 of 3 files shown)


pppobear:
## before
IMG_6624

## after
IMG_6625

PR #363, opened 2026-03-25
fix(hub): print direct local login link on startup

Sun-sunshine06:
## Summary

Print a direct local login link when HAPI Hub starts, and update CLI guidance to point users to that startup output.

## Why

When HAPI Hub is opened remotely or through a browser on another device, it can be inconvenient to manually find the access token in local config files.
A startup login URL makes first-time access much easier.

## Changes

- print a local login URL on hub startup
- extract direct-access URL helpers into a small utility
- add unit tests for tokenized URL generation
- update CLI auth/token help text to mention the startup login URL

## Notes

- this keeps the change scoped to discoverability and onboarding
- no runtime auth flow is changed

## Testing

- `cd hub && bun test src/utils/directAccess.test.ts`
- `bun typecheck`

### DIFF
--- a/cli/src/commands/auth.ts
+++ b/cli/src/commands/auth.ts
@@ -32,7 +32,7 @@ export async function handleAuthCommand(args: string[]): Promise<void> {
         if (!hasToken) {
             console.log('')
             console.log(chalk.yellow('  Token not configured. To get your token:'))
-            console.log(chalk.gray('    1. Check the server startup logs (first run shows generated token)'))
+            console.log(chalk.gray('    1. Check the server startup logs (hub now prints a local login URL)'))
             console.log(chalk.gray('    2. Read ~/.hapi/settings.json on the server'))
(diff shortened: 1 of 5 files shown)


hqhq1025:
Useful feature — the local login link is a nice DX improvement, and the code is well-structured (extracted utility + tests).

One concern: **the token is printed in plain text in the terminal log.**

\`\`\`
[Web] Login:  http://localhost:3006/?token=<full-token>
\`\`\`

If the user shares terminal output (e.g. in a bug report, screenshot, or log aggregation), the token leaks. For a local-only scenario the risk is low, but worth considering:

- Mask the token in the log output (e.g. show only the first 8 chars: \`?token=abc12345....\`)
- Or print the URL only on first run / behind a flag

This may be an acceptable tradeoff for the maintainer to decide — just flagging the risk.

tiann:

PR #382, opened 2026-04-01
fix(codex): handle remote request_user_input approvals

lwc-alex:
Fix remote Codex sessions hanging/canceling on `request_user_input` tool approvals.

Root cause:
- `codexRemoteLauncher` registers app-server permission handlers without `onUserInputRequest`
- `appServerPermissionAdapter` falls back to `cancel` for `item/tool/requestUserInput`

This patch wires `request_user_input` into the existing Codex remote permission flow and adds focused tests.

Validated with:
- `bun x vitest run src/codex/utils/permissionHandler.test.ts src/codex/utils/appServerPermissionAdapter.test.ts src/codex/codexRemoteLauncher.test.ts`
- `bun run typecheck`

### DIFF
--- a/cli/src/codex/codexRemoteLauncher.ts
+++ b/cli/src/codex/codexRemoteLauncher.ts
@@ -177,6 +177,17 @@ class CodexRemoteLauncher extends RemoteLauncherBase {
                 : undefined;
         }, {
             onRequest: ({ id, toolName, input }) => {
+                if (toolName === 'request_user_input') {
+                    session.sendAgentMessage({
+                        type: 'tool-call',
+                        name: 'request_user_input',
+                        callId: id,
+                        input,
+                        id: randomUUID()
+                    });
+                    return;
+                }
+
                 const inputRecord = input && typeof input === 'object' ? input as Record<string, unknown> : {};
                 const message = typeof inputRecord.message === 'string' ? inputRecord.message : undefined;
                 const rawCommand = inputRecord.command;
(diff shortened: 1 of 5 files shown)


hqhq1025 (approved):
Looks good — the root cause is real and the fix is correct. I traced the full approve flow end-to-end and it works.

Re: the bot finding about "deny downgraded to cancel" — this is consistent with existing behavior. All tool denials from Telegram (which calls `denyPermission` without a `decision` parameter) end up as `cancel` via `mapDecision('abort')`. The web UI's `RequestUserInputFooter` has no deny button, so this path is only reachable from Telegram. Not a regression.

Two minor suggestions (non-blocking):

**1. Status label inversion in `handlePermissionResponse`**

```ts
// Current (semantically inverted):
status: response.approved ? 'denied' : 'canceled',

// Should be:
status: response.approved ? 'canceled' : 'denied',
```

When `approved: true` but answers are empty, it's a system-level cancellation (not a user denial). When `approved: false`, it's a user denial. The current labels are swapped.

**2. No `onComplete` on the deny path**

The approve path calls `onComplete` (which sends `tool-call-result` to the web UI), but the deny path rejects without calling `onComplete`. This could leave the web UI tool card visually stuck in "pending" when denied from Telegram. Consider adding an `onComplete` call in the deny branch to match the regular tool pattern.

Both are minor — happy to see these addressed in a follow-up if you prefer.

PR #393, opened 2026-04-04
feat(codex): Add support for codex /fast mode

paulcalcraft:
Codex has a fast inference mode that uses tokens at a higher rate. I've added support for that when creating a new session. I also added the full set of reasoning levels for codex to the dropdown.

GPT 5.4 xhigh in codex wrote the code, I've tested that it doesn't break the flow and it seems to work, but hard to know for certain that the fast mode is applied correctly (it seems fast at least!). Despite docs saying that serviceTier should be 'priority' for fast mode, if you pass that param, you actually get an error saying that it should be 'fast', so this patch uses 'fast'.


hqhq1025 (changes requested):
The feature implementation is thorough — the web UI → hub → CLI → Codex data flow for the primary (remote) path is complete and correct. Good i18n coverage and test additions.

However, there's a critical bug that breaks session resume:

### Critical: Positional argument misalignment in `syncEngine.resumeSession`

The PR inserts `serviceTier` as a new positional parameter in `rpcGateway.spawnSession()` (between `modelReasoningEffort` and `yolo`), but `syncEngine.resumeSession()` (line ~402) calls `spawnSession()` with positional args and is **not updated**.

After this PR, the resume call passes:
- `resumeToken` → `worktreeName` (wrong!)
- `session.effort` → `resumeSessionId` (wrong!)
- `effort` → undefined (lost!)

TypeScript won't catch this because all misaligned params are `string?`. The result: **all inactive session resumes silently fail** — they create a new session instead of resuming.

Fix: add `undefined` for `serviceTier` at position 6 in the resume call. Or better yet, refactor `spawnSession` to accept an options object instead of 11 positional parameters — this class of bug will recur every time a parameter is added.

### Other issues

**Local mode ignores `--service-tier`:** `commands/codex.ts` parses the flag but doesn't forward it to the Codex child process args in local mode. Running `hapi codex --service-tier fast` locally does nothing. Consider at minimum logging a warning.

**`serviceTier` not persisted in session schema:** `SessionSchema` doesn't include `serviceTier`, so even after fixing the resume args, a resumed session loses the fast mode selection. (`modelReasoningEffort` has the same issue, but fast mode costs 2x usage so silent loss has higher impact.)

hqhq1025:
Nice feature — the implementation is thorough with good i18n and test coverage.

One critical issue: **`syncEngine.resumeSession()` is not updated for the new `serviceTier` parameter.**

`rpcGateway.spawnSession()` now takes `serviceTier` as a new positional parameter (between `modelReasoningEffort` and `yolo`), but `resumeSession()` still calls `spawnSession()` with the old positional args. Since all the misaligned params are `string?`, TypeScript won't catch this — but at runtime, **all inactive session resumes will silently fail** (they'll create a new session instead of resuming because the args are shifted).

Suggested fix: add `undefined` for `serviceTier` at the correct position in `syncEngine.resumeSession()`, or better yet, refactor `spawnSession` to accept an options object instead of 11+ positional parameters to prevent this class of bug from recurring.

Also note: `serviceTier` is not persisted in `SessionSchema`, so a resumed session will lose the fast mode selection even after the positional fix.

tiann:

PR #394, opened 2026-04-04
feat(claude): sync Claude Code session title to HAPI web UI

koenzhao:
## Summary
- Intercept `custom-title` events from Claude Code's JSONL session files and update HAPI session `metadata.name`
- When a user runs `/rename` in Claude Code, the title automatically syncs to the HAPI web UI within ~3 seconds
- On session startup, the latest title from JSONL history is applied immediately

## Motivation
Currently, HAPI relies on its own `mcp__hapi__change_title` MCP tool (via system prompt injection) to set session titles. This means Claude Code's native `/rename` command has no effect on the HAPI web UI — they are two separate naming systems.

This PR makes Claude Code the single source of truth for session titles by reading `custom-title` events directly from the JSONL files.

## Changes
- `cli/src/claude/utils/sessionScanner.ts` — Added `onTitleChange` callback; `readSessionLog` now detects `custom-title` events before Zod schema validation and returns them separately
- `cli/src/claude/claudeLocalLauncher.ts` — Wires `onTitleChange` to `session.client.updateMetadata` to set `metadata.name`

## Test plan
- [ ] Start a Claude session via `hapi claude`, verify session appears in web UI
- [ ] Run `/rename my-session` in Claude Code, verify title updates in web UI within ~3s
- [ ] Resume a renamed session, verify title persists on startup
- [ ] Start a fresh session (no prior rename), verify no errors

### DIFF
--- a/cli/src/claude/claudeLocalLauncher.ts
+++ b/cli/src/claude/claudeLocalLauncher.ts
@@ -26,6 +26,12 @@ export async function claudeLocalLauncher(session: Session): Promise<'switch' | 
                 return
             }
             session.client.sendClaudeSessionMessage(message)
+        },
+        onTitleChange: (title) => {
+            session.client.updateMetadata((metadata) => ({
+                ...metadata,
+                name: title
+            }))
         }
     });
 
--- a/cli/src/claude/utils/sessionScanner.ts
+++ b/cli/src/claude/utils/sessionScanner.ts
@@ -20,11 +20,13 @@ export async function createSessionScanner(opts: {
     sessionId: string | null;
     workingDirectory: string;
     onMessage: (message: RawJSONLines) => void;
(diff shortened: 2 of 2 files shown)


hqhq1025:
Clean and focused change. The title sync from Claude Code's JSONL log is a nice addition.

One issue: **pending session title events can overwrite the active session title.**

`findSessionFiles()` still scans `pendingSessions` during Claude session rollover. `parseSessionFile()` now forwards every `custom-title` it sees via `onTitleChange`, so a late rename written to the *previous* JSONL file can replace `metadata.name` for the *new* active Claude session.

Suggested fix: only fire `onTitleChange` when the title event comes from the current active session file, not from pending/finished session files. Something like guarding the `onTitleChange` call in the `parseSessionFile` loop so it only triggers for `currentSessionId`.

PR #421, opened 2026-04-09
feat: display background task count in status bar

mrrnb:
When Claude Code runs background tasks (run_in_background), the web UI now shows "N background task(s) running" with a pulsing blue indicator in the status bar instead of just "online".

Detection: scans message content for "Command running in background" (task start) and "<task-notification>" (task completion). Count is tracked in-memory per session and resets on session end.


mrrnb:
Re: i18n finding — intentionally kept in English to match Claude Code CLI's native status line, which displays "N background task(s) running" regardless of locale. Localizing this would actually diverge from the CLI behavior and break user expectations for those familiar with the terminal experience.

PR #440, opened 2026-04-11
fix(runner): prevent ghost sessions from orphaned spawn webhooks

BigKunLun:
## Summary

`spawnSession()` in `cli/src/runner/run.ts` currently registers a tracked session entry **before** the child's `Session started` webhook arrives. If the 15-second webhook timeout fires, only `pidToAwaiter` / `pidToErrorAwaiter` are cleaned; the `pidToTrackedSession` entry is **left in place** and the detached child keeps running. When the child eventually starts and reports its webhook, `onHappySessionWebhook()` finds the stale tracking entry and silently promotes the orphan into a normal runner-managed session — a message-less "ghost session" shows up on the web UI, and the user has no idea where it came from.

## Reproduction

Seen repeatedly on `opus[1m] --resume` on a macOS machine running the current Homebrew build (`0.16.6`). Concretely:

1. Tap **Resume** on a session whose claude-code child takes >15s to boot (observed real start delays up to **~60 minutes** under rate-limit + large session restore).
2. The hub returns `POST /api/sessions/:id/resume 500` after ~15s.
3. The user, getting 500s, taps Resume a few more times — each click spawns a fresh detached child. All of them eventually finish booting and report their webhook together, creating one ghost session per attempt.

In one user's dashboard this produced **four** ghost `Obsidian1210` sessions from a single troubled session, all rooted at the same `--resume <sessionId>`.

Relevant log excerpt:

```
18:19:35  Spawning session (...)  PID 63881  → webhook timeout 15s later
18:20:02  Spawning session (...)  PID 63921  → webhook timeout
18:20:14  Spawning session (...)  PID 63997  → webhook timeout
19:13:45  Spawning session (...)  PID 70231  → webhook timeout
19:20:16  PID 63921 webhook arrives → promoted to ghost session ab2799dd
19:20:16  PID 70231 webhook arrives → promoted to ghost session 778d408c
19:20:17  PID 63997 webhook arrives → promoted to ghost session 69cdf975
19:20:17  PID 63881 webhook arrives → promoted to ghost session 5996f036
```

Note the ~60-minute gap between `Waiting for session webhook` and the children actually reporting in — far beyond the current hard-coded 15s ceiling.

## Fix

Three small, composable changes in `cli/src/runner/run.ts`:

1. **Make the webhook timeout configurable** via `HAPI_RUNNER_WEBHOOK_TIMEOUT_MS` (default unchanged at `15_000`). Users on slow models / large resumes can raise the ceiling so the timeout never fires in the first place. Zero behaviour change for existing users.

2. **On timeout, clean `pidToTrackedSession` and `SIGTERM` the child.** Without this, the orphan's late webhook can still be silently promoted into a live session by `onHappySessionWebhook()`.

3. **Defence in depth** in `onHappySessionWebhook()`: if a webhook arrives for a PID that has no tracked session but its payload claims `startedBy: 'runner'`, treat it as an orphan — log it, `SIGTERM` the child, and return without registering the session. Genuine terminal-launched children correctly report `startedBy: 'terminal'` (see `commands/claude.ts`, `codex/loop.ts`, `opencode/runOpencode.ts`), so this branch cannot false-positive on them.

The original hard-coded timeout even carried a `// I have seen timeouts on 10 seconds even though session was still created successfully in ~2 more seconds` comment — the race it describes is the same race exploited in (2) and (3).

## Out of scope / follow-ups

- No new tests added in this PR. `runner.integration.test.ts` covers most of this flow but exercising the 60-min late-webhook window requires wall-clock mocking. Happy to add a regression test if reviewers prefer — just say the word and I'll wire one in with `vi.useFakeTimers()`.
- Docs: I did not touch `cli/src/runner/README.md` or `docs/`; let me know if you'd like `HAPI_RUNNER_WEBHOOK_TIMEOUT_MS` documented there.

## Verification

Manually reviewed against:

- `spawnSession()` (spawn → awaiter Promise)
- `onHappySessionWebhook()` (webhook dispatch)
- `stopSession()` / `onChildExited()` (cleanup paths — no new leaks)
- `controlServer.ts` (webhook payload shape — `startedBy` is always present)

Built locally is not currently possible on the submitter's machine (`bun` not installed), but each change is small enough to be reviewed by eye and the CI will catch any typecheck slip. I can provide a local smoke-test log from a Homebrew-replaced binary after merge if useful.

## Why this matters

Until the fix ships, every slow `--resume` on a slow-starting model produces a compounding pile of ghost sessions that the user can only clean up manually via `DELETE /api/sessions/:id` (and only after first `POST .../archive` to clear the `active` flag). The underlying bug has been latent since `detached: true` was introduced; it only becomes visible when the model's boot time crosses the 15s ceiling, which `opus[1m]` does routinely.

Thanks for maintaining HAPI 🙏

### DIFF
--- a/cli/src/runner/run.ts
+++ b/cli/src/runner/run.ts
@@ -204,14 +204,14 @@ export async function startRunner(): Promise<void> {
           logger.debug(
             `[RUNNER RUN] Ignoring late webhook from orphaned runner-spawned PID ${pid} (session ${sessionId}). Terminating child.`
           );
-          try {
-            process.kill(pid, 'SIGTERM');
-          } catch (err) {
-            logger.debug(
-              `[RUNNER RUN] Failed to SIGTERM orphaned runner child PID ${pid}:`,
-              err
-            );
-          }
+          // Use killProcess (SIGTERM → SIGKILL escalation) rather than a
+          // bare process.kill() so the orphan is reliably reaped even if
+          // it ignores SIGTERM.  We don't have a ChildProcess reference
+          // here (tracking entry was already removed by the timeout
+          // handler), so tree-kill via killProcessByChildProcess is not
+          // available — but the timeout handler should have already
+          // tree-killed the process group; this is defence-in-depth.
+          void killProcess(pid);
           return;
         }
 
@@ -557,19 +557,23 @@ export async function startRunner(): Promise<void> {
             // ghost session by onHappySessionWebhook().
             pidToTrackedSession.delete(pid);
 
-            // Best-effort terminate the orphan child. Without this, children
-            // spawned with `detached: true` keep running past the webhook
-            // deadline, eventually report back, and are turned into ghost
-            // sessions in the web UI. A single SIGTERM gives them a chance
-            // to clean up; even if the signal is ignored we already removed
-            // their tracking entry above, so they cannot be promoted.
-            try {
-              happyProcess?.kill('SIGTERM');
-            } catch (err) {
-              logger.debug(
-                `[RUNNER RUN] Failed to SIGTERM orphaned PID ${pid} after webhook timeout:`,
-                err
-              );
+            // Terminate the entire process tree (wrapper + agent
+            // grandchildren).  Using killProcessByChildProcess instead of
+            // a bare SIGTERM ensures that detached grandchild processes
+            // (the actual claude/codex agent) are also reaped, and that
+            // SIGTERM → SIGKILL escalation kicks in if needed.
+            if (happyProcess) {
+              void killProcessByChildProcess(happyProcess);
+            }
+
+            // If this was a worktree session, the worktree can only be
+            // safely removed after the child has actually exited (the
+            // child may still be writing to it).  Register a one-shot
+            // exit listener so cleanup happens once the tree-kill lands.
+            if (worktreeInfo && happyProcess) {
+              happyProcess.once('exit', () => {
+                void cleanupWorktree();
+              });
             }
 
             logger.debug(`[RUNNER RUN] Session webhook timeout for PID ${pid}`);
--- a/cli/src/runner/run.ts
+++ b/cli/src/runner/run.ts
@@ -127,6 +127,18 @@ export async function startRunner(): Promise<void> {
     // Setup state - key by PID
     const pidToTrackedSession = new Map<number, TrackedSession>();
 
+    // Webhook timeout tolerance. Opus 1M + --resume can legitimately take
+    // longer than the default 15s to reach the "Session started" webhook
+    // (observed real-world durations of 30s – 60min under rate-limit /
+    // heavy session restore). Allow advanced users to raise this ceiling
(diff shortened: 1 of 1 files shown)


hqhq1025 (commented):
Nice analysis on the ghost session root cause — the 15s ceiling + detached children race is well documented.

Two things on the kill path:

1. Both the timeout handler and the `onHappySessionWebhook` orphan branch use bare `SIGTERM`, but runner sessions are spawned with `detached: true` — the actual agent (claude/codex) is a grandchild that survives wrapper termination. The existing `stopSession()` already handles this via `killProcessByChildProcess()`. Suggested fix for the timeout path:

```ts
const orphan = pidToTrackedSession.get(pid);
if (orphan?.childProcess) {
    void killProcessByChildProcess(orphan.childProcess);
}
pidToTrackedSession.delete(pid);
```

For the webhook defense path where you only have a PID number, `killProcess(pid)` from `cli/src/utils/process.ts` handles SIGTERM → SIGKILL escalation.

2. When a worktree session times out, `pidToTrackedSession.delete(pid)` runs before the child exits, so `onChildExited()` can't find the entry to trigger worktree cleanup → permanent worktree leak. One approach: listen for the child's `exit` event in the timeout branch and clean up there.

Both are the same pattern — reusing the existing process lifecycle helpers instead of raw signals. The rest of the logic (configurable timeout, tracked entry cleanup, `startedBy` guard) looks solid.

BigKunLun:
Thanks for the thorough review @hqhq1025! Both points are spot-on — pushed a follow-up commit addressing them:

**1. Kill path — tree-kill instead of bare SIGTERM**

- **Timeout handler**: replaced `happyProcess.kill('SIGTERM')` with `killProcessByChildProcess(happyProcess)` to reap the entire process tree (wrapper + detached agent grandchildren), consistent with how `stopSession()` already works.
- **Orphan webhook handler**: replaced bare `process.kill(pid, 'SIGTERM')` with `killProcess(pid)` for proper SIGTERM → SIGKILL escalation. A `ChildProcess` reference isn't available here (tracking entry was already deleted by the timeout handler), so `killProcessByChildProcess` isn't possible — but the timeout handler should have already tree-killed the group by this point; this branch is defence-in-depth.

**2. Worktree leak — deferred cleanup on child exit**

Registered a one-shot `exit` listener on the child process inside the timeout branch:

```ts
if (worktreeInfo && happyProcess) {
  happyProcess.once('exit', () => {
    void cleanupWorktree();
  });
}
```

Previously `maybeCleanupWorktree('spawn-error')` skipped cleanup because `isProcessAlive(pid)` was still true right after sending SIGTERM, and `onChildExited()` couldn't find the tracking entry (already deleted) — so the worktree leaked permanently. Now cleanup waits for the child to actually exit.

heavygee:
Tracked root-cause follow-up: #728 — `resume_failed` / blank ghost session from `set-session-config` RPC racing `session-alive` immediately after spawn (not the same as persistence-only fixes).

PR #445, opened 2026-04-11
feat(cli): support optional extra headers for hub requests

### LINKED ISSUE #430: 功能建议：支持为连接 Hub 的请求配置自定义 Header
### 背景

目前 HAPI 在自托管场景下，客户端 / runner 连接 Hub 时，主要依赖：

- `CLI_API_TOKEN` / Bearer Token 的 HTTP API 认证
- Socket.IO 连接时的 auth token

这在普通部署下没有问题，但如果 Hub 放在某些认证代理后面，就会比较难用。

比如以下场景：

- Cloudflare Access
- OAuth2 Proxy
- Google IAP
- 企业内网里的统一认证反代
- 其他要求固定 Header / Cookie 的反向代理

这些场景下，虽然 Hub 的 URL 可达，但 HAPI 客户端侧如果不能额外附带代理要求的 Header / Cookie，就无法正常连接。

---

### 希望支持的能力

希望 HAPI 能支持一个**通用的“自定义请求头”配置**，用于客户端 / runner 连接 Hub 时附加额外 Header，而不是做某个厂商特定的适配。

例如支持类似：

- 环境变量：`HAPI_EXTRA_HEADERS_JSON`
- 和/或 `settings.json` 中增加 `extraHeaders`

例如：

```json
{
  "Cookie": "CF_Authorization=...",
  "CF-Access-Client-Id": "...",
  "CF-Access-Client-Secret": "..."
}
```

---

### 希望生效的范围

希望这个能力能统一应用到：

1. 连接 Hub 的 HTTP API 请求
2. Socket.IO / WebSocket / polling 连接

因为如果只支持 HTTP，不支持 socket，很多实时能力还是会受影响。

---

### 为什么建议做成通用 Header，而不是 Cloudflare 专项支持

因为这个需求不只是 Cloudflare Access 才会遇到，很多认证代理 / 反代场景都会有类似要求。

如果做成通用能力，可以同时覆盖：

- Cloudflare Access
- OAuth2 Proxy
- IAP 类网关
- 企业内网认证代理
- 其他自定义网关

这样对自托管部署会更灵活，也避免引入厂商特定逻辑。

---

### 设计上希望保持简单

我理解这个功能最好是：

- 可选能力，不影响现有用户
- 只有配置了才生效
- 作为高级 / 自托管场景能力提供
- 尽量只做“透传 Header”，不改变现有认证模型

---

### 想确认的问题

1. 维护者是否愿意接受这种“通用自定义 Header”能力？
2. 更推荐只做环境变量，还是同时支持 `settings.json` 配置？
3. 是否建议 HTTP 和 Socket.IO 一起支持，而不是只支持其中一部分？

如果这个方向可以，我可以再整理一个尽量聚焦的 PR。

Shujakuinkuraudo:
## Summary
This PR adds an optional env-based configuration for attaching extra outbound headers when the CLI/runner connects to a self-hosted hub behind an auth proxy.

Scope is intentionally minimal:
- env var only: `HAPI_EXTRA_HEADERS_JSON`
- applies to outbound CLI -> hub traffic only
- supports both REST requests and Socket.IO connections
- no hub-side behavior changes
- no settings.json persistence in this first iteration

## Use case
This helps self-hosted deployments that sit behind an auth proxy or access gateway, such as Cloudflare Access, OAuth2 Proxy, IAP, or other reverse proxies that require extra request headers/cookies.

Example:

```bash
export HAPI_EXTRA_HEADERS_JSON='{"Cookie":"CF_Authorization=..."}'
```

or:

```bash
export HAPI_EXTRA_HEADERS_JSON='{"CF-Access-Client-Id":"...","CF-Access-Client-Secret":"..."}'
```

## What changed
- parse `HAPI_EXTRA_HEADERS_JSON` in CLI configuration
- merge those headers into CLI -> hub REST requests
- pass the same headers into Socket.IO transport options
- add focused tests for parsing and header propagation
- add minimal docs for the new env var

## Notes
- built-in request headers still win over custom ones (for example `Authorization`)
- invalid JSON is ignored with a warning
- only string header values are kept
- response parsing was also relaxed slightly for optional session fields to stay compatible with hubs that omit them

## Validation
Ran focused CLI tests:

```bash
vitest run src/api/hubExtraHeaders.test.ts src/api/api.extraHeaders.test.ts --config vitest.config.ts
```

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -80,6 +80,7 @@ See `src/configuration.ts` for all options.
 
 - `HAPI_HOME` - Config/data directory (default: ~/.hapi).
 - `HAPI_EXPERIMENTAL` - Enable experimental features (true/1/yes).
+- `HAPI_EXTRA_HEADERS_JSON` - JSON object of extra headers to send on CLI → hub requests, e.g. `{"Cookie":"CF_Authorization=..."}`.
 - `HAPI_CLAUDE_PATH` - Path to a specific `claude` executable.
 - `HAPI_HTTP_MCP_URL` - Default MCP target for `hapi mcp`.
 
--- a/cli/src/api/api.extraHeaders.test.ts
+++ b/cli/src/api/api.extraHeaders.test.ts
@@ -1,0 +1,160 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { configuration } from '@/configuration'
+
+const axiosPostMock = vi.hoisted(() => vi.fn())
+const ioMock = vi.hoisted(() => vi.fn())
+
+vi.mock('axios', () => ({
+    default: {
+        post: axiosPostMock
+    }
+}))
+
+vi.mock('@/api/auth', () => ({
+    getAuthToken: () => 'cli-token'
+}))
+
+vi.mock('socket.io-client', () => ({
+    io: ioMock
+}))
+
+vi.mock('@/api/rpc/RpcHandlerManager', () => ({
+    RpcHandlerManager: class {
+        onSocketConnect(): void { }
+        onSocketDisconnect(): void { }
+        registerHandler(): void { }
+        handleRequest(): Promise<string> {
+            return Promise.resolve('{}')
+        }
+    }
+}))
+
+vi.mock('../modules/common/registerCommonHandlers', () => ({
+    registerCommonHandlers: () => { }
+}))
+
+vi.mock('@/terminal/TerminalManager', () => ({
+    TerminalManager: class {
+        closeAll(): void { }
+    }
+}))
+
+import { ApiClient } from './api'
+import { ApiSessionClient } from './apiSession'
+
+describe('API extra headers integration', () => {
+    const now = 1_710_000_000_000
+
+    beforeEach(() => {
+        configuration._setApiUrl('https://hapi.example.com')
+        configuration._setExtraHeaders({})
+        axiosPostMock.mockReset()
+        ioMock.mockReset()
+    })
+
+    it('adds extra headers to REST requests', async () => {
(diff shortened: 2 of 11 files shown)


hqhq1025 (commented):
Clean implementation — the spread order in `buildHubRequestHeaders` correctly prevents user-supplied headers from overriding `Authorization`, and all three CLI→hub paths (REST, session socket, machine socket) are covered.

A couple of notes:

1. Socket.IO's Node.js client supports a top-level `extraHeaders` option directly on `io()`, which would be simpler than the per-transport `transportOptions` approach. Both work, but the simpler version is less code:
```ts
return Object.keys(h).length > 0 ? { extraHeaders: { ...h } } : {}
```

2. The schema change (`.nullable()` → `.nullable().optional().default(null)`) in `shared/src/schemas.ts` is a good defensive fix but unrelated to extra headers — might be worth a separate commit for clarity in git history.

3. Minor: `parseExtraHeaders` silently drops non-string values without warning. A `console.warn` like the invalid-JSON path would help users debug misconfigured headers.

Looks good overall.

Shujakuinkuraudo:
Addressed the follow-up suggestions:

1. Switched the Socket.IO client config to use the top-level `extraHeaders` option instead of per-transport `transportOptions`.
2. Added a warning when `HAPI_EXTRA_HEADERS_JSON` contains non-string values, while still ignoring those invalid entries.

Also updated the related tests accordingly.

I did not add any new schema-history cleanup commit in this follow-up.

hqhq1025 (approved):
Looks good — author addressed the follow-up suggestions (top-level extraHeaders, non-string value warning). All three CLI→hub paths covered, tests updated.

PR #459, opened 2026-04-14
fix(voice): create ElevenLabs tools before agent setup

Shujakuinkuraudo:
## Summary
- replace deprecated ElevenLabs `prompt.tools` usage with `prompt.tool_ids`
- auto-create or reuse the required client tools before creating/updating the Hapi voice agent
- keep the shared web helper aligned with the new ElevenLabs tool-based agent setup

## Testing
- verified `createOrUpdateHapiAgent()` succeeds with an ElevenLabs key that has ConvAI write access
- verified the resulting ElevenLabs agent includes `messageCodingAgent` and `processPermissionRequest`
- built the Linux x64 single-executable binary successfully via `bun run build:single-exe`

### DIFF
--- a/hub/src/web/routes/voice.ts
+++ b/hub/src/web/routes/voice.ts
@@ -4,7 +4,9 @@ import type { WebAppEnv } from '../middleware/auth'
 import {
     ELEVENLABS_API_BASE,
     VOICE_AGENT_NAME,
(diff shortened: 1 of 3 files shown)


Shujakuinkuraudo:
Closing in favor of #464, which contains only the fix portion from this branch.

PR #470, opened 2026-04-16
Fix session list freshness and unread indicators

Liu-KM:
## Summary

Fix stale session-list state and add lightweight unread indicators for sessions that update while the user is looking elsewhere.

- update `sessions.updated_at` whenever messages are stored or moved between sessions
- keep a global SSE subscription active so the sidebar receives updates even when a specific session is selected
- patch session-list timestamps immediately from `message-received` events
- persist per-session last-seen timestamps in localStorage and show unread dots on sessions, project groups, and machines

## Why

The sidebar could show stale thinking/status and stale relative times until a session was opened. This made completed sessions still look active and made it hard to tell which background sessions had new output.

## Validation

- `cd hub && bun test src/sync/sessionModel.test.ts`
- `cd web && bun test src/components/SessionList.test.ts`
- `bun run typecheck:hub`
- `bun run typecheck:web`

## Notes

AI-assisted with GPT-5.4.

### DIFF
--- a/hub/src/store/messages.ts
+++ b/hub/src/store/messages.ts
@@ -64,6 +64,16 @@ export function addMessage(
         local_id: localId ?? null
     })
 
+    db.prepare(`
+        UPDATE sessions
+        SET updated_at = CASE WHEN updated_at > @updated_at THEN updated_at ELSE @updated_at END,
+            seq = seq + 1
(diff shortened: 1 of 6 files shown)


Liu-KM:
Closing this PR and keeping focus on the clean one (#471).

PR #472, opened 2026-04-16
Support steering active Codex turns from queued messages

Liu-KM:
## Summary

Add support for steering an active Codex app-server turn with queued messages.

- add typed `turn/steer` support to the Codex app-server client
- when a same-mode message arrives while a turn is active, send it to `turn/steer`
- requeue messages if steering fails
- keep turn lifecycle state active until a terminal turn event arrives, preventing a second turn from starting too early

## Why

Codex can accept additional instructions while a task is running. HAPI previously queued those messages for a later turn, and could also start another turn before the active turn had emitted a terminal event, which could leave sessions stuck in a thinking state.

## Validation

- `cd cli && bun x vitest run src/codex/codexRemoteLauncher.test.ts`

## Notes

AI-assisted with GPT-5.4.

### DIFF
--- a/cli/src/agent/backends/acp/AcpSdkBackend.test.ts
+++ b/cli/src/agent/backends/acp/AcpSdkBackend.test.ts
@@ -98,7 +98,7 @@ describe('AcpSdkBackend', () => {
                     });
                 }, 0);
 
-                await sleep(5);
+                await sleep(1);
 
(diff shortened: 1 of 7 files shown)


Liu-KM:
Closing this PR and keeping focus on the clean one (#471).

PR #489, opened 2026-04-18
feat(web): add collapsible archived section in session list

junmo-kim:
## Summary

When working across multiple machines, archived project groups accumulate
in the sidebar and push active sessions far apart, requiring excessive
scrolling.

- Within each machine group, archived project groups are now separated
  into a collapsible "Archived (N)" section (default: collapsed)
- The section only appears when both active and archived groups coexist
  in the same machine — if all groups are archived, no separator is shown
- Selecting an archived session auto-expands its section

## Test plan

- [x] Open session list with multiple machines, each having active and archived sessions
- [x] Verify archived groups appear under "Archived (N)" toggle, collapsed by default
- [x] Click toggle to expand/collapse archived section
- [x] Select an archived session — verify section auto-expands
- [x] Verify no section separator when all groups in a machine are archived
- [x] Keyboard navigation: Tab to archived toggle, Enter/Space to toggle

### DIFF
--- a/web/src/components/SessionList.tsx
+++ b/web/src/components/SessionList.tsx
@@ -483,22 +483,24 @@ function ProjectGroupItem(props: {
     const { group, isCollapsed, onToggle, onSelect, api, selectedSessionId } = props
     return (
         <div>
-            <button
-                type="button"
-                className="group/project sticky top-0 z-10 flex items-center gap-2 px-1 py-1.5 text-left rounded-lg transition-colors hover:bg-[var(--app-subtle-bg)] cursor-pointer min-w-0 w-full select-none"
-                onClick={onToggle}
-                title={group.directory}
-                aria-expanded={!isCollapsed}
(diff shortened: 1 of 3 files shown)


junmo-kim:
Closing this PR.

After rebasing on latest `main`, I noticed PR #547 (sidebar search + per-group preview limit) was merged just before, which addresses overlapping concerns from a different angle (generic per-group preview cap with show-more). The two approaches now conflict semantically: this PR splits at the *group* level by archived state, while #547 caps at the *session* level regardless of state.

I think a unified design is needed — preferably one that distinguishes active vs archived sessions explicitly, since the preview cap can still surface archived sessions in front of active ones in mixed groups. But that's a bigger conversation and likely worth opening as an issue first rather than re-pushing here.

Closing for now to avoid maintaining a half-redundant feature. Thanks for the reviews!

Issue #428, opened 2026-04-10
feat: Add message queuing support to match Claude.ai's streaming prompt behavior


ashavolian:
### What feature would you like to see?

When using CC directly, I often stream thoughts as separate prompts — sending them as they come to mind, regardless of whether the LLM is currently processing. I'd like HAPI to support message queuing the same way CC does, so messages sent while a response is in-flight are held and processed in order. HAPI already supports this behavior.

hqhq1025:
Same need here — I often want to send follow-up instructions while the agent is still running, without waiting for the current turn to finish.

From what I can see, the backend already supports this: `MessageQueue2` in the CLI accepts messages at any time and batches them. The gap is purely on the web UI side — the send button is disabled while `threadIsRunning`.

There's an existing draft PR #274 that tackles this with a client-side queue + `Queued` status indicator. I've left some feedback there.

cc @tiann — for the UI design (queued message display, status badges, etc.), might be worth discussing together to make sure it fits the overall UX direction before we finalize the implementation.

PR #492, opened 2026-04-19
feat(web): show queued status for messages pending inference

junmo-kim:
## Motivation

Since v0.16.7 (#422), messages sent mid-turn are persisted on the hub immediately but sit in the CLI's queue until the agent finishes. The UI currently treats "saved on hub" and "consumed by CLI" identically — both render as `sent`.

The two states are behaviorally different. "Saved on hub" means durable; "consumed by CLI" means the agent has started processing it. Future work — floating queued messages above the composer, cancelling them, re-ordering — needs this distinction. This PR adds a minimal visual cue for it.

## UX

While the session is thinking, new messages show a clock icon and dimmed text (`queued`). When the CLI drains its queue, it emits `messages-consumed` with the consumed `localId`s, and those messages transition to `sent`. Backward-compatible.

## Implementation

- `MessageQueue2` carries `localId` per item; `collectBatch()` reports the consumed ones via `onBatchConsumed(localIds)`.
- `messages-consumed` socket/SSE event: `{ sid, localIds: string[] }`, forwarded unchanged by the hub.
- Web `markMessagesConsumed` flips only the matching queued messages; `mergeMessages` preserves `queued` across the optimistic → server-echo replacement (otherwise the affordance would vanish within tens of ms).

## Follow-ups (separate PRs)

- Floating UI for queued messages (relies on `localId`-based ack).
- Cancel queued messages (adds a cancel event on `MessageQueue2`).
- SSE replay/backfill — if `messages-consumed` is dropped mid-disconnect a delivered message can stay dimmed until refresh. Visual glitch, not correctness. Broader SSE concern.

## Related

- Related: #428
- Builds on: #422
- Alternative to: #274 (server-side ack vs client-only queue)

### DIFF
--- a/cli/src/agent/runners/runAgentSession.ts
+++ b/cli/src/agent/runners/runAgentSession.ts
@@ -50,9 +50,9 @@ export async function runAgentSession(opts: {
 
     const messageQueue = new MessageQueue2<Record<string, never>>(() => hashObject({}));
 
-    session.onUserMessage((message) => {
+    session.onUserMessage((message, localId) => {
         const formattedText = formatMessageWithAttachments(message.content.text, message.content.attachments);
-        messageQueue.push(formattedText, {});
+        messageQueue.push(formattedText, {}, localId);
     });
 
     let currentPermissionMode: SessionPermissionMode = opts.permissionMode ?? sessionInfo.permissionMode ?? 'default';
--- a/cli/src/agent/sessionBase.ts
+++ b/cli/src/agent/sessionBase.ts
@@ -70,7 +70,7 @@ export class AgentSessionBase<Mode> {
         this.effort = opts.effort;
         this.collaborationMode = opts.collaborationMode;
 
-        this.queue.onBatchConsumed = (consumedAt) => this.client.emitMessagesConsumed(consumedAt);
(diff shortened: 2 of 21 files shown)


junmo-kim:
Thanks for catching this — the edge case is real, but the suggested fix would undo the PR's intent.

The `mergeMessages` carry-over of `queued` onto the stored server message was added deliberately. Without it the affordance disappears within tens of ms (as soon as the hub echoes the message back), long before the CLI actually picks it up. Distinguishing "saved on hub" from "consumed by CLI" is the whole point of the PR — see the updated description.

The remaining concern is narrower: an in-flight `messages-consumed` SSE being dropped across a disconnect. Options I considered:

1. **Broad `thinking:false` fallback** — removed in 65dbf69 because it over-acks later batches (your earlier review).
2. **SSE replay/backfill for `messages-consumed`** — larger scope, server-side work.
3. **Accept the trade-off**: on a missed ack the delivered message stays dimmed, and a refresh keeps it that way (same merge logic). Visual glitch, not correctness — the message was processed.

I'd prefer to leave this PR as-is and tackle the replay path separately if it bites users in practice. Listed in Follow-ups.

PR #514, opened 2026-04-22
fix(hub): mark queued sessions thinking before Codex task_start

xiaobaifly7:
## Summary
- mark a session as `thinking=true` immediately after the hub accepts a user message
- keep that queued-thinking state through short `thinking=false` keepalive heartbeats until a real task start arrives or a short grace window expires
- add sync-layer regression coverage for immediate thinking, grace-window preservation, and expiry

## Problem
When a remote Codex message was sent from the web app, the UI could stay on `online` for a while even though the message had already been accepted by the hub. The status only switched later when Codex emitted `task_started` or a later alive update.

In practice this produced a visible delay where the message was already sent but the session state did not reflect that work had begun yet.

## Fix
The hub now sets queued sessions to `thinking=true` as soon as `sendMessage()` succeeds, and preserves that state across short false heartbeats from the CLI until either:
- a real `thinking=true` update arrives
- the grace window expires

This keeps the UI responsive without leaving sessions stuck in thinking forever.

## Tests
- `bun test src/sync/aliveEvents.test.ts`
- `bun test src/sync/sessionModel.test.ts`

### DIFF
--- a/hub/src/sync/aliveEvents.test.ts
+++ b/hub/src/sync/aliveEvents.test.ts
@@ -169,7 +169,13 @@ describe('alive incremental events', () => {
         cache.markMessageQueued(session.id, now + 10)
         events.length = 0
 
-        cache.handleSessionAlive({ sid: session.id, time: now + 2_000, thinking: false })
+        const originalNow = Date.now
+        Date.now = () => now + 2_000
+        try {
+            cache.handleSessionAlive({ sid: session.id, time: now + 2_000, thinking: false })
+        } finally {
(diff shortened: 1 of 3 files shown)


xiaobaifly7:
@tiann 可以看看我这个pr吗

tiann:
看上面 bot 的 review

xiaobaifly7 ***@***.***> 于2026年4月23日周四 23:22写道：

xiaobaifly7:
@tiann 搞定了，麻烦合并

PR #516, opened 2026-04-22
fix(hub): raise Socket.IO maxHttpBufferSize for file downloads

GeT-LeFt:
## Summary

Socket.IO Engine defaults `maxHttpBufferSize` to **1MB**. File downloads in HAPI go through WebSocket RPC (the CLI reads a file → base64-encodes it → returns it via Socket.IO ack), so any file larger than ~750KB (after base64 inflation) silently drops the response. The frontend download spinner then hangs indefinitely with no error surfaced.

## Changes

- Raise the default `maxHttpBufferSize` to **100MB** in `hub/src/socket/server.ts`.
- Expose it via a new `HAPI_SOCKET_MAX_BUFFER_SIZE` env var (using the existing `resolveEnvNumber` helper) so operators can tune it up or down based on their deployment.

## Test plan

- [x] `bun run typecheck` passes
- [x] Manually reproduced the hang on `main` with a ~2MB file download
- [x] Verified downloads of files up to 50MB succeed after the change

### DIFF
--- a/hub/src/socket/server.ts
+++ b/hub/src/socket/server.ts
@@ -20,15 +20,15 @@ const jwtPayloadSchema = z.object({
 
 const DEFAULT_IDLE_TIMEOUT_MS = 15 * 60_000
 const DEFAULT_MAX_TERMINALS = 4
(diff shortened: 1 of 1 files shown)


tiann:

PR #522, opened 2026-04-23
[codex] support GPT-5.5 model

mrrnb:
## Summary

- Add shared Codex model presets with GPT-5.5 as the first selectable option.
- Reuse the shared Codex model list in new-session and composer model controls.
- Allow remote Codex sessions to apply model changes through hub and CLI config RPC.

## Impact

Codex sessions can be started or remotely switched to `gpt-5.5`. Local Codex sessions still cannot be reconfigured from the web while controlled by the local user.

## Validation

- `bun test shared/src/models.test.ts shared/src/flavors.test.ts`
- `bun run test -- modelOptions.test.ts types.test.ts` in `web`
- `bun test src/web/routes/sessions.test.ts` in `hub`
- `bun run test -- runCodex.test.ts` in `cli`
- `bun run typecheck` in `cli`, `web`, and `hub`
- `git diff --check`

### DIFF
--- a/cli/src/codex/runCodex.test.ts
+++ b/cli/src/codex/runCodex.test.ts
@@ -1,0 +1,95 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+const harness = vi.hoisted(() => ({
+    rpcHandlers: new Map<string, (payload: unknown) => Promise<unknown> | unknown>(),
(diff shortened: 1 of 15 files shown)


tiann:
https://github.com/tiann/hapi/commit/97be34e21c7276b603dba82aa35f00d7240e0988

PR #557, opened 2026-05-01
fix(web): show metadata slash commands

xiaobaifly7:
## Summary

- merge slash commands captured in session metadata into the `/api/sessions/:id/slash-commands` response
- fall back to metadata-provided slash commands when live RPC listing fails
- add regression coverage for metadata fallback and API-provided built-in command preservation

## Why

Claude sessions can persist a `metadata.slashCommands` snapshot at startup. The web UI previously depended on live RPC listing for slash-command suggestions, so commands captured in metadata could disappear when the active RPC path was unavailable or incomplete. This made the remote `/` menu show fewer commands than the session had already captured.

## Validation

- `cd hub && bun test src/web/routes/sessions.test.ts`
- `cd web && bun test src/lib/codexSlashCommands.test.ts`


xiaobaifly7:
Superseded by #578. Consolidating approved fixes into one canonical PR; the rejected runner process reuse change from #560 is excluded.

PR #559, opened 2026-05-03
fix(codex): preserve thread during retryable failures

xiaobaifly7:
## Summary

- keep Codex remote sessions on the same app-server thread after retryable thread-level failures
- compact the same thread before retrying context-window overflow failures
- surface resume/compact failures without silently starting a fresh conversation

## Root cause

HAPI treated Codex thread-level `systemError` notifications as invalid thread failures and cleared the current thread state. The next queued message could then create a new app-server thread, which breaks conversation continuity. Context-window overflow also needed an explicit same-thread compact/retry path instead of asking the user to start over.

## Validation

- `rtk bun run test:win src/codex/utils/appServerEventConverter.test.ts`
- `rtk bun run test:win src/codex/codexRemoteLauncher.test.ts`
- `rtk bun run typecheck`

Note: the final validation commands were run from `tmp/hapi-src-0.17.2/cli`, which has the workspace dependencies installed, after syncing the same six Codex files from this PR branch. Running Vitest directly inside the clean PR worktree was blocked by missing `node_modules` / `vitest` resolution on Windows.

### DIFF
--- a/cli/src/codex/codexRemoteLauncher.test.ts
+++ b/cli/src/codex/codexRemoteLauncher.test.ts
@@ -12,7 +12,12 @@ const harness = vi.hoisted(() => ({
     interruptedTurns: [] as Array<{ threadId: string; turnId: string }>,
     compactThreadIds: [] as string[],
     suppressTurnCompletion: false,
-    remainingThreadSystemErrors: 0
+    remainingThreadSystemErrors: 0,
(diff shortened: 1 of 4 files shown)


xiaobaifly7:
Superseded by #578. Consolidating approved fixes into one canonical PR; the rejected runner process reuse change from #560 is excluded.

PR #560, opened 2026-05-04
[codex] Reuse duplicate runner spawn targets

xiaobaifly7:
## Summary

This PR prevents the runner from spawning duplicate simple sessions for the same target.

- add a runner spawn key for simple sessions based on agent, directory, resume target, model/effort, permission mode, and token hash
- reuse an existing live runner-spawned session when a duplicate target is requested
- make concurrent duplicate spawn requests wait for the first in-flight spawn result instead of launching another child process
- keep worktree sessions out of the dedupe path

## Root cause

Repeated remote spawn requests for the same Codex resume target could pass through the runner as independent child processes. On Windows this can leave multiple long-running `hapi codex resume ...` processes for the same conversation, increasing process/pipe pressure and making later agent turn startup less reliable.

## Validation

- `bunx --bun vitest run src/runner/spawnDedupe.test.ts`
- `bun run typecheck`

### DIFF
--- a/cli/src/runner/run.ts
+++ b/cli/src/runner/run.ts
@@ -24,6 +24,7 @@ import { join } from 'path';
 import { buildMachineMetadata } from '@/agent/sessionFactory';
 import { resolveWorkspaceRoot } from '@/utils/workspaceRoot';
 import { hashRunnerCliApiToken } from './runnerIdentity';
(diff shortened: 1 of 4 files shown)


tiann:
I don’t think this is a good idea; reusing processes may introduce wired bugs that are difficult to troubleshoot.

PR #561, opened 2026-05-04
fix(hub): harden resume recovery and Codex model discovery

xiaobaifly7:
## Summary
- Recover Claude resume IDs from prior assistant/user message content so resumed sessions can reconnect to the original provider thread.
- Extend Codex model-list RPC acknowledgements to 120s while preserving the 30s default timeout for normal RPC calls.
- Keep compiled HAPI child processes on the same executable via `HAPI_CLI_EXECUTABLE`, avoiding fallback to a globally installed `hapi.exe`.
- Add regression coverage for resume recovery, RPC timeout selection, and compiled executable inheritance.

## Root cause
- Some resumed Claude sessions only had the provider resume identifier in message content, not in the newer metadata location, so resume could lose the original provider thread id.
- Codex model discovery can take longer than the generic 30s Socket.IO ack timeout during app-server startup or plugin sync, causing HTTP 500 `operation has timed out` even when the model list eventually becomes available.
- In compiled Windows runs, child CLI processes could resolve through `process.execPath` / PATH in a way that used a different global npm HAPI binary instead of the executable that launched the runner.

## Validation
- `bun run --cwd cli test:win src/utils/spawnHappyCLI.test.ts` -> 11 passed
- `bun run --cwd hub test src/sync/rpcGateway.test.ts` -> 2 passed
- `bun run typecheck:cli` -> exit 0
- `bun run typecheck:hub` -> exit 0
- `bun run build:single-exe` -> exit 0, Windows exe SHA256 `E2000B0576761F710F9BD28EB937185D9134DB6FA52AD2E23476F1B04DFD4BB9`
- Local live smoke: `/api/machines/:id/codex-models` returned success in ~3.3s with 4 models: `gpt-5.5`, `gpt-5.4-mini`, `gpt-5.3-codex`, `gpt-5.2`
- Local live process check: hub, runner, Codex, Claude, and MCP HAPI child processes are now running from `C:\Users\Administrator\.hapi\patched\resume-recovery-0.17.2\hapi.exe`

### DIFF
--- a/cli/src/utils/spawnHappyCLI.test.ts
+++ b/cli/src/utils/spawnHappyCLI.test.ts
@@ -1,7 +1,17 @@
 import { beforeAll, afterAll, beforeEach, describe, expect, it, vi } from 'vitest';
 import type { SpawnOptions } from 'child_process';
 
-const spawnMock = vi.fn((..._args: any[]) => ({ pid: 12345 } as any));
+const {
+  spawnMock,
+  existsSyncMock,
+  isBunCompiledMock,
+  projectPathMock
+} = vi.hoisted(() => ({
+  spawnMock: vi.fn((..._args: any[]) => ({ pid: 12345 }) as any),
+  existsSyncMock: vi.fn((path: string) => !path.includes('missing-hapi.exe')),
+  isBunCompiledMock: vi.fn(() => false),
+  projectPathMock: vi.fn(() => process.cwd())
+}));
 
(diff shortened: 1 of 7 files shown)


xiaobaifly7:
@tiann

xiaobaifly7:
Superseded by #578. Consolidating the approved fixes into one canonical branch/PR to avoid overlapping or stacked PRs.

PR #578, opened 2026-05-06
fix(hapi): consolidate approved web and Codex recovery fixes

xiaobaifly7:
## Summary

This is the canonical consolidated PR for the approved HAPI fixes from my currently open branches. It combines the non-overlapping fixes from #557, #559, and #561 into one branch so the upstream review does not need to deal with stacked or duplicate PRs.

Included fixes:

- Show metadata-provided slash commands when the RPC slash-command listing fails.
- Preserve the existing Codex conversation/thread across retryable app-server failures instead of starting a new conversation.
- Recover Claude resume IDs from stored messages when session metadata is missing them.
- Recover the newest Claude resume ID when stored messages contain multiple Claude session IDs.
- Extend Codex model-list RPC timeout and fix Windows child-process path handling for packaged `hapi.exe`.

Explicitly excluded:

- #560 / `515b50e` (`fix(runner): 复用重复启动目标`) is not included, because the runner process reuse approach was rejected in review.

## Validation

Local:

- `bun run --cwd cli test:win src/codex/codexRemoteLauncher.test.ts src/codex/utils/appServerEventConverter.test.ts src/utils/spawnHappyCLI.test.ts` — 56 tests passed
- `bun run --cwd hub test src/sync/sessionModel.test.ts src/sync/rpcGateway.test.ts src/web/routes/sessions.test.ts` — 54 tests passed
- `bun run typecheck:cli` — passed
- `bun run typecheck:hub` — passed
- `bun run build:single-exe` — passed before the review follow-up; the follow-up only changes hub resume-ID selection and tests

GitHub Actions:

- `test` — passed
- `Codex PR Review` — passed after addressing the review finding

### DIFF
--- a/cli/src/codex/codexRemoteLauncher.test.ts
+++ b/cli/src/codex/codexRemoteLauncher.test.ts
@@ -12,7 +12,12 @@ const harness = vi.hoisted(() => ({
     interruptedTurns: [] as Array<{ threadId: string; turnId: string }>,
     compactThreadIds: [] as string[],
     suppressTurnCompletion: false,
-    remainingThreadSystemErrors: 0
+    remainingThreadSystemErrors: 0,
+    startTurnMessages: [] as string[],
+    failResumeThreadIds: [] as string[],
+    nextThreadSystemErrorMessage: null as string | null,
+    failNextCompact: false,
+    deferThreadStatusNotifications: false
 }));
 
 vi.mock('./codexAppServerClient', () => {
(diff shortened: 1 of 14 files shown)


xiaobaifly7:
Temporarily closing and reopening to retrigger the Codex PR Review check after it failed due to model capacity, without changing the PR commits.

xiaobaifly7:
@tiann

PR #580, opened 2026-05-06
fix(web): show subagent trace for Agent tool name

junmo-kim:
## Problem

PR #539 added a Trace section to the Task tool dialog by reading `block.children`, gated on `block.tool.name === 'Task'`. The current Claude Code SDK reports the same subagent invocation under the name `Agent` (with identical `input.prompt` and `subagent_type` shape). As a result the Trace section, the main-card child labels, and the reducer sidechain grouping all skip these calls — the modal shows only Input and Result again, even though `block.children` would otherwise be populated.

## Solution

Extends the Task-tool guards across the reducer pipeline and UI surfaces to recognise `Agent` as an additional subagent tool name. The shared contract — sidechain user message text matching `tool_use.input.prompt` — applies unchanged. No data plumbing or schema changes.

- `tracer.ts` — index Agent prompts alongside Task prompts.
- `reducerTimeline.ts` — populate `block.children` and skip the duplicated prompt text for Agent calls.
- `ToolCard.tsx` / `trace.tsx` — render the Trace section, the input prompt block, and the inline-suppressed layout for Agent.
- `ToolMessage.tsx` — show the same child summary in the main card.
- `knownTools.tsx` / `getToolName.ts` — i18n label `Agent` → "Launch Agent" for parity with Task.
- `hub/src/telegram/sessionView.ts` — Telegram session view falls through Agent to the existing Task case.

A small `web/src/chat/subagentTool.ts` helper (`isSubagentToolName`) centralises the `Task | Agent` predicate so future SDK additions only touch one place.

## Tests

New unit tests cover the reducer pipeline and the dialog-level rendering:

- `web/src/chat/tracer.test.ts` — indexes prompts from both names and matches sidechain roots.
- `web/src/chat/reducerTimeline.test.ts` — populates `block.children` and suppresses duplicate prompt text for both names.
- `web/src/components/ToolCard/trace.test.tsx` — `getTaskTraceChildren` and `TraceSection` accept both names.

All existing tests continue to pass (`bun run test`). Typecheck is clean across `cli`, `web`, and `hub`.

End-to-end verification on a live isolated stack was not run for this PR; the change is intentionally narrow (extending a name guard while keeping the data contract intact) and is fully exercised by the unit tests above. Reviewers running a recent SDK should observe the Trace section render against any general-purpose Agent invocation.

### DIFF
--- a/cli/src/claude/utils/getToolName.ts
+++ b/cli/src/claude/utils/getToolName.ts
@@ -28,6 +28,7 @@ const STANDARD_TOOLS: Record<string, string> = {
     'TodoWrite': 'Update Tasks',
     'TodoRead': 'Read Tasks',
     'Task': 'Launch Agent',
+    'Agent': 'Launch Agent',
 
     // Team management
     'TeamCreate': 'Create Team',
--- a/hub/src/telegram/sessionView.ts
+++ b/hub/src/telegram/sessionView.ts
@@ -101,6 +101,7 @@ function formatToolArgumentsDetailed(tool: string, args: any): string {
                 return `Command: ${truncate(cmd, MAX_TOOL_ARGS_LENGTH)}`
             }
 
+            case 'Agent':
             case 'Task': {
                 const desc = args.description || args.prompt || ''
                 return `Task: ${truncate(desc, MAX_TOOL_ARGS_LENGTH)}`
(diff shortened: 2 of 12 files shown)


junmo-kim:
Closing in favor of #585 after live TUI parity verification re-framed the change. The new PR carries the same `Task → Agent` guard fix plus AgentFullView/AgentResultView re-alignment commits per HAPI Bot's [Major] feedback — the Agent dialog now matches the Claude Code TUI `ctrl+o` expand transcript view (Prompt / Trace / Response) end-to-end.

PR #598, opened 2026-05-08
Fix Codex subagent final result preservation

SmallSpider0:
## Summary

Fix Codex subagent result handling so completed subagent cards preserve the actual final assistant message instead of sometimes showing only a terminal status such as `done`.

## Problem

In some Codex multi-agent flows, a child agent could complete successfully but the parent-side result would only contain a completion marker (`done` / `completed`) rather than the child agent's final message.

This could happen when:

- `wait_agent` returned a completed/done status without a message
- child terminal events and child message events arrived in a non-ideal order
- later child trace/tool events arrived after the card had already reached a terminal state

## Changes

- Track each child Codex runtime's latest final `agent_message`
- Use that cached final message as fallback when `wait_agent` reports completed/done without a message
- Normalize `done`, `complete`, and `completed` agent states consistently
- Prevent terminal child agent cards from regressing back to running due to late child events
- Allow explicit parent operations like `send_input` / `resume_agent` to remain visible without letting late child tool events overwrite terminal state
- Reset cached final message and terminal state on child `task_started`
- Preserve late child messages as completed result updates instead of writing/running updates

## Tests

Added coverage for:

- `wait_agent` completed/done without message
- preserving child final message as result
- multiple child messages using the latest message
- terminal child event before late `agent_message`
- preventing terminal cards from regressing to running
- late child command after `resume_agent`
- converter normalization for done-without-message states

Verified locally:

```bash
bun run test -- src/codex/utils/appServerEventConverter.test.ts src/codex/codexRemoteLauncher.test.ts
bun run typecheck
```

Also smoke-tested against a deployed local HAPI build with real Codex subagents:

- spawned two subagents
- waited for both
- verified `wait_agent` result contained full final messages, not only `done`

### DIFF
--- a/cli/src/agent/backends/acp/AcpSdkBackend.test.ts
+++ b/cli/src/agent/backends/acp/AcpSdkBackend.test.ts
@@ -214,7 +214,7 @@ describe('AcpSdkBackend', () => {
     });
 
     it('emits turn_complete after trailing tool updates from the same turn', async () => {
-        backendStatics.UPDATE_QUIET_PERIOD_MS = 8;
+        backendStatics.UPDATE_QUIET_PERIOD_MS = 25;
         backendStatics.UPDATE_DRAIN_TIMEOUT_MS = 200;
         backendStatics.PRE_PROMPT_UPDATE_QUIET_PERIOD_MS = 1;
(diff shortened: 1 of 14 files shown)


SmallSpider0:
Closing per request.

PR #603, opened 2026-05-08
feat(cli): add /model slash command to show or set Claude model

metowolf:
## Summary

- Add a new `/model` slash command that allows users to view or switch the Claude model during a session.
- `/model` (no argument) — prints the currently active model (e.g. `Claude model: claude-sonnet-4-6`).
- `/model <name>` — switches the session to the given model (e.g. `/model claude-opus-4-7`).
- `/model auto` — resets the session to the default/auto model selection.

## Changes

- **`cli/src/parsers/specialCommands.ts`** — Added `parseModel()` function and `ModelCommandResult` interface; wired into `parseSpecialCommand()` which now returns `type: 'model'` and the optional model string.
- **`cli/src/parsers/specialCommands.test.ts`** — Full unit-test coverage for `parseModel` and the `/model` cases in `parseSpecialCommand`.
- **`cli/src/claude/runClaude.ts`** — Handle `specialCommand.type === 'model'`: show current model when no argument is supplied, or call `setModel` / `pushKeepAlive` when a model name is given.
- **`cli/src/modules/common/slashCommands.ts`** — Register `/model` in the built-in command list with a description so it appears in autocomplete / help.

## Test plan

- [x] Run `vitest` — all existing and new tests pass.
- [x] Send `/model` in a session — response shows `Claude model: <current>`.
- [x] Send `/model claude-opus-4-7` — response shows `Claude model set to claude-opus-4-7` and subsequent messages use that model.
- [x] Send `/model auto` — resets to default model selection.
- [x] Confirm that `/modeler` and other partial matches do **not** trigger the command.

### DIFF
--- a/cli/src/claude/runClaude.ts
+++ b/cli/src/claude/runClaude.ts
@@ -317,6 +317,28 @@ export async function runClaude(options: StartOptions = {}): Promise<void> {
             return;
         }
 
+        if (specialCommand.type === 'model') {
+            logger.debug('[start] Detected /model command');
+            if (specialCommand.model == null) {
+                // /model with no arg: show current model
+                const modelName = currentModel ?? 'auto';
+                session.sendSessionEvent({ type: 'message', message: `Claude model: ${modelName}` });
+            } else {
+                // /model <name> or /model auto: set model
+                const newModel = specialCommand.model === 'auto' ? null : specialCommand.model;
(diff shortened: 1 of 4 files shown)


tiann:
I don’t think this is a good idea. Codex / Claude Code have slash-command functionality because they are TUIs, whereas we are on the web, with buttons and dialogs, so there’s no need to use this kind of interaction.

PR #589, opened 2026-05-06
[OpenCode] fix: Add SQLite support to OpenCode storage scanner

DullJZ:
- Add SQLite database connection and query capabilities
- Implement session discovery from database (5-10x faster)
- Add fallback to file system for compatibility
- Improve performance from 100-500ms to 10-50ms
- Add comprehensive error handling and logging
- Support both SQLite and JSON file storage formats
- Resolve: No OpenCode session found within 120000ms timeout error

Files changed:
  - cli/src/opencode/utils/opencodeStorageScanner.ts

PR #617, opened 2026-05-12
fix(opencode): Windows path case mismatch in SQLite scanner

spralle:
Builds on #589 — the `WHERE directory = ?` query never matched on Windows because `normalizePath()` lowercases paths but the DB stores original casing (`C:\foo` vs `c:\foo`).

Uses `LOWER(directory)` on win32 so the comparison works. Tested locally, session discovery now succeeds.


tiann:
Already fixed in #589

PR #618, opened 2026-05-13
Fix Codex title MCP startup and remote terminal spawning

SmallSpider0:
# Fix Codex title MCP startup and remote terminal spawning

## Summary

- Wrap the HAPI MCP stdio bridge with an inline Node proxy when launching it from Codex app-server.
- Keep the proxy inline so compiled/release HAPI does not require an extra sidecar file.
- Replace Bun terminal spawning in remote terminals with Node `child_process.spawn`, using `/usr/bin/script` on non-Darwin Unix when available.
- Safely quote `$SHELL` when constructing the `script -c` command.

## Motivation

Codex app-server can fail to initialize the Bun-backed HAPI MCP stdio bridge, leaving `mcp__hapi__change_title` unavailable in web sessions. Remote terminal sessions can also close immediately with `Terminal stream closed unexpectedly` in daemon/web environments.

## Notes

- `HAPI_NODE_EXECUTABLE` can override the Node executable used for the MCP stdio proxy.
- Windows remote terminals remain unsupported, matching the existing behavior.
- Terminal resize is best-effort in the new implementation; fidelity can vary depending on `/usr/bin/script` availability.

## Validation

- `cd cli && bun x vitest run src/codex/utils/codexMcpConfig.test.ts src/codex/utils/appServerConfig.test.ts src/codex/codexLocal.test.ts src/codex/utils/appServerEventConverter.test.ts src/codex/utils/terminalEventGuard.test.ts` — 79 tests passed
- Codex app-server smoke: HAPI MCP startup reports `HAPI_MCP ready` using the generated inline proxy args
- Terminal smoke: observed `echo HAPI_TERMINAL_OK` output and exit 0

## Known local test limitation

`bun run test` was attempted, but this machine is Ubuntu 20.04 and the bundled `cli/tools/unpacked/difft` binary requires newer GLIBC versions (`GLIBC_2.32`, `GLIBC_2.33`, `GLIBC_2.34`). As a result, only `src/modules/difftastic/index.test.ts` fails locally; the remaining CLI tests run past that point and the targeted tests for this change pass.

### DIFF
--- a/cli/src/codex/utils/buildHapiMcpBridge.ts
+++ b/cli/src/codex/utils/buildHapiMcpBridge.ts
@@ -7,6 +7,7 @@
 
 import { startHappyServer } from '@/claude/utils/startHappyServer';
 import { getHappyCliCommand } from '@/utils/spawnHappyCLI';
+import { spawnSync } from 'node:child_process';
 import type { ApiSessionClient } from '@/api/apiSession';
 
 /**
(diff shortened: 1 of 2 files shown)


SmallSpider0:
Updated in 925001c to address the bot review:

- The inline MCP proxy now falls back to the original HAPI MCP bridge if no Node executable is available; `HAPI_NODE_EXECUTABLE` remains supported as an explicit override.
- macOS remote terminals now keep the Bun terminal path so they retain PTY behavior and resize support.
- `$SHELL` is POSIX-quoted before use with `script -c`.

Revalidated locally:
- targeted Codex/terminal vitest suite: 79 tests passed
- generated bridge config smoke: Codex app-server reports `HAPI_MCP ready`
- terminal smoke: `echo HAPI_TERMINAL_OK`, exit 0

SmallSpider0:
Updated again in 0d36b8d to address the follow-up resize finding:

- The default non-Darwin terminal path now uses a small Python stdlib PTY bridge when `python3`/`python` is available.
- Resize requests are forwarded to the active PTY with `TIOCSWINSZ` and `SIGWINCH`.
- `/usr/bin/script` remains only a last-resort fallback when Python is unavailable.

Revalidated locally:
- targeted Codex/terminal vitest suite: 79 tests passed
- generated bridge config smoke: Codex app-server reports `HAPI_MCP ready`
- terminal smoke: `echo HAPI_TERMINAL_OK`, exit 0
- terminal resize smoke: after resizing to 100x30, `stty size` reports `30 100`, exit 0

tiann:
> Codex app-server may fail to initialize the Bun-backed HAPI MCP stdio bridge

Why might this failure occur? We should investigate the root cause if possible.

SmallSpider0:
Closing this combined PR so the two independent fixes can be split into smaller, user-focused PRs.

PR #622, opened 2026-05-14
fix(codex): support app-server plan mode

NightWatcher314:
## Summary

Implements Codex plan mode using the latest Codex app-server capability surface.

## What changed

- Adds `collaborationMode/list` probing after app-server initialization.
- Sends `turn/start.collaborationMode` for Codex plan turns when the runtime advertises `plan`.
- Falls back to a normal `turn/start` only when the runtime clearly rejects the `collaborationMode` field as unsupported/unknown.
- Keeps unrelated `collaborationMode` failures as real turn failures instead of silently downgrading them.
- Handles latest app-server `item/permissions/requestApproval` requests and maps approved permission profiles back to app-server responses.
- Keeps generic `item/tool/requestApproval` support for tools such as `exit_plan_mode` / `ExitPlanMode`.
- Resets HAPI's stored Codex collaboration mode to `default` after an approved exit-plan request and emits keepalive immediately.

## Difference from the previous implementation

Before this change, HAPI treated Codex collaboration mode as a static turn parameter: if the UI/session mode was `plan`, it always attached `collaborationMode` to `turn/start` and assumed the runtime would accept it. There was no app-server capability probe, no narrow fallback path, and no latest permission-request handler for app-server permission grants.

With this change, HAPI first checks the app-server's advertised collaboration modes, sends native plan mode only when appropriate, and has a controlled fallback for older or incompatible runtimes. The fallback preserves the selected model as a top-level `model` parameter, so a downgraded turn does not lose model selection. Plan exit also updates HAPI's session state back to `default`, keeping Web status and CLI runtime state aligned.

## Validation

- `cd cli && bun run test codexRemoteLauncher.test.ts appServerConfig.test.ts appServerPermissionAdapter.test.ts`
- `cd cli && bun run typecheck`
- `git diff --check`

### DIFF
--- a/cli/src/codex/appServerTypes.ts
+++ b/cli/src/codex/appServerTypes.ts
@@ -44,6 +44,19 @@ export interface ModelListResponse {
     [key: string]: unknown;
 }
 
+export interface CollaborationModeListItem {
+    name?: string;
+    mode?: 'plan' | 'default' | string | null;
+    model?: string | null;
+    reasoning_effort?: ReasoningEffort | null;
+    [key: string]: unknown;
+}
+
+export interface CollaborationModeListResponse {
+    data?: CollaborationModeListItem[];
+    [key: string]: unknown;
+}
+
 export interface ThreadStartParams {
     model?: string;
     modelProvider?: string;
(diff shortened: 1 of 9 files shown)


NightWatcher314:
Follow-up for the review findings is pushed in `c5825d7`.

Changes:
- `shouldRetryWithoutCollaborationMode` now recognizes spaced `collaboration mode` wording in addition to `collaborationMode` / `collaboration_mode`.
- `collaborationMode/list` support detection now accepts name-only entries such as `{ name: 'plan' }`.
- Added launcher coverage for both review cases.

Validation rerun:
- `cd cli && bun run test codexRemoteLauncher.test.ts appServerConfig.test.ts appServerPermissionAdapter.test.ts`
- `cd cli && bun run typecheck`
- `git diff --check`

PR #638, opened 2026-05-18
Fix Codex MCP startup under Bun stdin

SmallSpider0:
Split out from the closed combined PR #618: https://github.com/tiann/hapi/pull/618

This PR fixes the Codex/HAPI MCP startup issue without introducing a Node stdio proxy. HAPI still runs the MCP bridge with Bun.

## Problem

This affects runner-managed Codex remote sessions, especially when the hub and the runner are running on different machines. In that setup, users start or control a Codex session from the web/mobile UI, while the actual Codex process runs on the runner machine.

HAPI expects Codex to load the built-in title update tool so new sessions can get useful names in the session list. On current `main`, Codex app-server can fail to start the HAPI MCP bridge, so the title update tool is unavailable. The user-visible result is that remote Codex sessions may keep generic, empty, or stale titles, making them hard to identify later from the browser or phone.

## Reproduction on current `main`

Tested on:

- HAPI `main`: `5512890`
- Bun: `1.3.13`
- Codex CLI: `0.130.0`
- OS: Ubuntu 20.04 x86_64

Steps:

1. Start a local HAPI HTTP MCP server and build the normal HAPI MCP stdio config:

   ```ts
   const happyServer = await startHappyServer(client);
   const bridgeCommand = getHappyCliCommand(['mcp', '--url', happyServer.url]);
   ```

   On `main`, this produces a Bun-backed stdio MCP server similar to:

   ```bash
   /root/.bun/bin/bun --cwd /path/to/hapi/cli /path/to/hapi/cli/src/index.ts mcp --url http://127.0.0.1:<port>/
   ```

2. Start Codex app-server and create a thread with that MCP server config:

   ```ts
   const app = new CodexAppServerClient();

   await app.connect();
   await app.initialize({
     clientInfo: { name: 'hapi-mcp-smoke', version: '0' },
     capabilities: { experimentalApi: true }
   });

   await app.startThread({
     cwd: '/tmp',
     approvalPolicy: 'on-request',
     sandbox: 'workspace-write',
     baseInstructions: 'smoke',
     developerInstructions: 'smoke',
     config: {
       'mcp_servers.hapi': {
         command: bridgeCommand.command,
         args: bridgeCommand.args
       },
       developer_instructions: 'smoke'
     }
   });
   ```

3. Observe Codex app-server MCP startup events:

   ```text
   mcpServer/startupStatus/updated {"name":"hapi","status":"starting","error":null}
   mcpServer/startupStatus/updated {"name":"hapi","status":"failed","error":"MCP client for `hapi` failed to start: MCP startup failed: handshaking with MCP server failed: connection closed: initialize response"}
   ```

Expected:

```text
hapi: ready
```

Actual on current `main`:

```text
hapi: failed
MCP startup failed: handshaking with MCP server failed: connection closed: initialize response
```

Control checks on the same `main` checkout:

- directly starting `hapi mcp` stays alive after 2 seconds
- stdout/stderr are empty
- MCP SDK stdio client can connect to the same Bun-backed bridge
- `listTools` returns `change_title`
- `callTool(change_title)` succeeds

So the bridge is valid for a normal MCP stdio client; the failure is specific to Codex app-server talking directly to the Bun-backed stdio MCP child.

## Root cause

The failing boundary is Codex app-server 0.130 talking directly to a Bun-backed stdio MCP child.

A minimal Bun MCP stdio server using the MCP SDK's default `StdioServerTransport()` reproduces the same failure under Codex app-server:

```text
MCP startup failed: handshaking with MCP server failed: connection closed: initialize response
```

The Bun child logs show:

```text
stdin error Error: EPERM: operation not permitted, read
stdin close
```

This happens because the SDK default transport reads from Bun's global `process.stdin` stream. In this Codex app-server child-process setup, that Bun stdin stream can fail with `EPERM` even though fd 0 itself is readable.

Two controls confirm that fd 0 and the MCP messages are fine:

- a Bun minimal server that reads fd 0 with `fs.readSync(0, ...)` works with Codex app-server
- a Bun minimal server using `fs.createReadStream(..., { fd: 0 })` also works with Codex app-server

## Fix

Use an explicit fd 0 read stream for the HAPI MCP stdio bridge when running under Bun:

```ts
export function createHapiMcpStdin(): Readable {
  return process.versions.bun
    ? createReadStream('/dev/stdin', { fd: 0, autoClose: false }) as Readable
    : process.stdin;
}
```

Then pass that stream to the MCP SDK transport:

```ts
const stdio = new StdioServerTransport(createHapiMcpStdin(), process.stdout);
```

This avoids the Bun `process.stdin` edge case while keeping the bridge running on Bun. No Node proxy or extra runtime process is introduced.

## Validation

- `cd cli && bun x vitest run src/codex/happyMcpStdioBridge.test.ts src/codex/utils/codexMcpConfig.test.ts src/codex/utils/appServerConfig.test.ts`

Smoke checks after this fix:

- direct `hapi mcp`: alive after 2s, stdout/stderr empty
- MCP SDK stdio client -> Bun-backed `hapi mcp`: `listTools` and `callTool(change_title)` succeed
- Codex app-server -> Bun-backed `hapi mcp`: `hapi` startup reaches `ready`

### DIFF
--- a/cli/src/codex/happyMcpStdioBridge.test.ts
+++ b/cli/src/codex/happyMcpStdioBridge.test.ts
@@ -1,0 +1,50 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { PassThrough } from 'node:stream';
+
+const mocks = vi.hoisted(() => ({
+    createReadStream: vi.fn(() => new PassThrough())
+}));
+
+vi.mock('node:fs', async (importOriginal) => ({
+    ...(await importOriginal<typeof import('node:fs')>()),
+    createReadStream: mocks.createReadStream
+}));
+
+import { createHapiMcpStdin } from './happyMcpStdioBridge';
+
+describe('createHapiMcpStdin', () => {
+    const originalBunVersion = process.versions.bun;
+
+    afterEach(() => {
+        if (originalBunVersion === undefined) {
+            Reflect.deleteProperty(process.versions, 'bun');
+        } else {
+            Object.defineProperty(process.versions, 'bun', {
+                value: originalBunVersion,
+                configurable: true
+            });
+        }
+        vi.clearAllMocks();
+    });
+
+    it('uses an explicit fd 0 fs stream under Bun to avoid process.stdin EPERM', () => {
+        Object.defineProperty(process.versions, 'bun', {
+            value: '1.3.13',
+            configurable: true
+        });
+
+        const stdin = createHapiMcpStdin();
+
+        expect(mocks.createReadStream).toHaveBeenCalledWith('/dev/stdin', { fd: 0, autoClose: false });
+        expect(stdin).not.toBe(process.stdin);
+    });
+
+    it('keeps process.stdin outside Bun', () => {
+        Reflect.deleteProperty(process.versions, 'bun');
+
+        const stdin = createHapiMcpStdin();
+
+        expect(stdin).toBe(process.stdin);
+        expect(mocks.createReadStream).not.toHaveBeenCalled();
+    });
+});
--- a/cli/src/codex/happyMcpStdioBridge.ts
+++ b/cli/src/codex/happyMcpStdioBridge.ts
@@ -13,9 +13,17 @@
 
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { createReadStream } from 'node:fs';
+import type { Readable } from 'node:stream';
 import { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js';
 import { z } from 'zod';
+
+export function createHapiMcpStdin(): Readable {
+  return process.versions.bun
+    ? createReadStream('/dev/stdin', { fd: 0, autoClose: false }) as Readable
+    : process.stdin;
+}
 
 function parseArgs(argv: string[]): { url: string | null } {
   let url: string | null = null;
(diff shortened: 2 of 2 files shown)


tiann:
> Codex app-server can fail to start the HAPI MCP bridge.

Why does the HAPI MCP bridge fail? HAPI itself is written by Bun. If you fix the issue this way, does it mean we should use Node instead?

Actually, I have never encountered this issue. I think you should find the root cause.

SmallSpider0:
Thanks for pushing on the root cause. I investigated this further and replaced the Node proxy workaround with a narrower Bun-side fix.

The failure is not that HAPI needs Node, and it is not that the HAPI MCP bridge is invalid.

What I found:

1. On current `main`, directly starting the Bun-backed `hapi mcp` process works:
   - the process stays alive
   - stdout/stderr are clean

2. A normal MCP SDK stdio client can connect to the same Bun-backed `hapi mcp` bridge:
   - `listTools` returns `change_title`
   - `callTool(change_title)` succeeds

3. The failure only appears when Codex app-server 0.130 starts that Bun-backed stdio MCP bridge:
   - Codex reports:
     `handshaking with MCP server failed: connection closed: initialize response`

4. A minimal Bun MCP stdio server reproduces the same failure under Codex app-server when it uses the MCP SDK default `StdioServerTransport()`, which reads from Bun's global `process.stdin`.

5. In that failing minimal server, the Bun child logs:
   `stdin error Error: EPERM: operation not permitted, read`

6. The same minimal server works if it reads fd 0 through either:
   - `fs.readSync(0, ...)`, or
   - `fs.createReadStream(..., { fd: 0 })`

So the root cause appears to be a Bun `process.stdin` stream edge case in the Codex app-server child-process stdio setup. fd 0 itself is readable, but Bun's global `process.stdin` stream can fail with `EPERM`.

The updated fix keeps HAPI running on Bun. Instead of wrapping the bridge with Node, the HAPI MCP bridge now passes an explicit fd 0 stream to the MCP SDK transport under Bun:

```ts
const stdin = process.versions.bun
  ? createReadStream('/dev/stdin', { fd: 0, autoClose: false })
  : process.stdin;

const stdio = new StdioServerTransport(stdin, process.stdout);
```

After this change, Codex app-server can start the Bun-backed HAPI MCP bridge directly and the `hapi` MCP server reaches `ready`.

So this PR is no longer a Node compatibility proxy. It is a targeted fix for the Bun `process.stdin` read failure in this specific Codex app-server stdio child scenario.

tiann:
I’m also using Ubuntu, and my bun version and codex version are the same as yours, but I don’t have this issue on my system. I think there might be something wrong with your environment.

SmallSpider0:
I tested this on another server and was not able to reproduce the issue.

This suggests the behavior is likely caused by a server-specific environment rather than a general HAPI bug. Because of that, I don’t think this change is suitable for merging into the main branch.

I’ll close this PR for now. For future customization needs like this, I’ll consider implementing them through the plugin system instead.

PR #639, opened 2026-05-18
Fix runner remote terminal disconnecting from web/mobile UI

SmallSpider0:
# Fix runner remote terminal disconnecting from web/mobile UI

Split out from the closed combined PR #618: https://github.com/tiann/hapi/pull/618

This PR contains only the runner remote-terminal PTY fix. The Codex/HAPI MCP startup fix is intentionally kept in a separate PR so each user-visible issue has its own root-cause discussion and review scope.

## Problem

This affects runner-managed remote terminals, especially when the hub and the runner are running on different machines. In that setup, users open a terminal from the web/mobile UI expecting to run commands on the runner machine.

On current `main`, the terminal can report ready and then immediately disconnect in daemon/headless runner environments. The user-visible result is that the web/mobile terminal cannot be used to run commands on the runner machine, which breaks the remote terminal workflow.

## Root cause

The current remote terminal implementation relies on Bun's `Bun.spawn(..., { terminal })` PTY path. In the runner daemon/headless path this can be unstable: the shell may receive `SIGHUP` shortly after the terminal is created, before user input is processed.

A pipe-only fallback is not enough because interactive shells and full-screen terminal apps need a real PTY, and resize events must reach the active PTY instead of only updating parent-side bookkeeping.

## Fix

Use platform-specific terminal creation:

- macOS keeps the existing Bun PTY path to preserve native PTY behavior.
- Linux/non-Darwin defaults to a small Python stdlib PTY bridge when `python3` or `python` is available.
- The Python bridge owns the PTY, forwards terminal input/output, applies resize with `TIOCSWINSZ`, and sends `SIGWINCH` to the child process group.
- `/usr/bin/script` / pipe spawning remains a last-resort fallback when Python is unavailable.

This gives runner-managed web/mobile terminals a real PTY that stays alive and handles resize correctly.

## Validation

- `cd cli && bun x vitest run src/terminal/TerminalManager.test.ts`

### DIFF
--- a/cli/src/terminal/TerminalManager.test.ts
+++ b/cli/src/terminal/TerminalManager.test.ts
@@ -1,0 +1,136 @@
+import { PassThrough } from 'node:stream';
+import { EventEmitter } from 'node:events';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const mocks = vi.hoisted(() => ({
+    spawn: vi.fn(),
+    spawnSync: vi.fn()
+}));
+
+vi.mock('node:child_process', () => ({
+    spawn: mocks.spawn,
+    spawnSync: mocks.spawnSync
+}));
+
+vi.mock('@/ui/logger', () => ({
+    logger: {
+        debug: vi.fn(),
+        warn: vi.fn(),
+        error: vi.fn(),
+        info: vi.fn()
+    }
+}));
+
+import { TerminalManager } from './TerminalManager';
+
+type MockChild = EventEmitter & {
+    stdin: PassThrough;
+    stdout: EventEmitter;
+    stderr: EventEmitter;
+    killed: boolean;
+    exitCode: number | null;
+    kill: ReturnType<typeof vi.fn>;
+};
+
+function createMockChild(): { child: MockChild; stdinLines: string[] } {
+    const stdin = new PassThrough();
+    const stdinLines: string[] = [];
+    stdin.on('data', (chunk) => {
+        stdinLines.push(chunk.toString());
+    });
+
+    const child = Object.assign(new EventEmitter(), {
+        stdin,
+        stdout: new EventEmitter(),
+        stderr: new EventEmitter(),
+        killed: false,
+        exitCode: null,
+        kill: vi.fn(function (this: MockChild) {
+            this.killed = true;
+            this.exitCode = 0;
+        })
+    }) as MockChild;
+
+    return { child, stdinLines };
+}
+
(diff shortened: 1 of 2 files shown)


SmallSpider0:
Updated in 785df70 to address the bot review:

- The Python PTY bridge child now calls `setsid()` and `TIOCSCTTY` in `preexec_fn`, so the shell gets the PTY slave as its controlling terminal.
- Bridge shutdown now handles `SIGTERM`/`SIGINT`/`SIGHUP`, terminates the child process group, and closes the PTY master so foreground terminal jobs receive hangup instead of being orphaned.
- The normal close path also terminates the child group and closes the PTY master before waiting, with a SIGKILL fallback if the child does not exit.
- Added test assertions covering the controlling-terminal and child-group cleanup code paths.

Validation:
- `cd cli && bun x vitest run src/terminal/TerminalManager.test.ts`
- local smoke: normal terminal command exits 0
- local smoke: closing a terminal while a foreground `sleep` is running leaves no matching orphan process

SmallSpider0:
Updated in d223a38 to address the latest bot review and similar lifecycle issues:

- Stale Bun PTY fallback exits are now ignored by checking that each output/error/exit callback still belongs to the currently registered runtime before it can emit events or call `cleanup()`.
- Python PTY cleanup no longer immediately kills the bridge process from Node. `cleanup()` now sends the bridge `close` command and lets the bridge run its own child-process-group termination and SIGKILL escalation path.
- The Python bridge signal path now uses the same `shutdown_child_group()` helper, so external `SIGTERM`/`SIGINT`/`SIGHUP` also wait briefly and escalate to `SIGKILL` if needed.
- Added regression tests for both cases: Python close cleanup must not call `child.kill()`, and a stale failed Bun subprocess must not tear down the fallback terminal.
- Also guarded stale stdout/stderr/error callbacks for Bun, Python, and pipe-backed terminals so replacement terminals do not receive events from an older process using the same terminal id.

Validation:
- `cd cli && bun x vitest run src/terminal/TerminalManager.test.ts`
- local smoke: normal terminal command exits 0
- local smoke: closing a terminal while a foreground `sleep` is running leaves no matching orphan process

SmallSpider0:
I tested this on another server and was not able to reproduce the issue.

This suggests the behavior is likely caused by a server-specific environment rather than a general HAPI bug. Because of that, I don’t think this change is suitable for merging into the main branch.

I’ll close this PR for now. For future customization needs like this, I’ll consider implementing them through the plugin system instead.

PR #652, opened 2026-05-20
fix(codex): stabilize goal status UI events

SmallSpider0:
## Summary

- Suppress redundant Codex goal status text events (`Goal active`, `Goal active · N tokens`, `Goal complete`, `Goal cleared`) before persistence/broadcast, and filter already-persisted redundant rows from message history.
- Keep structured goal update/clear events silent in the chat timeline while still using them for the composer/status-bar goal indicator.
- Clear the bottom-right completed-goal indicator as soon as later non-`/goal` user activity is known, including queued optimistic messages and message-window pending rows.

## Background / previous PRs

This is a follow-up to the earlier goal status fixes in #649 and #651.

- #649 de-duplicated repeated structured `thread_goal_updated` events. That fixed one source of repeated goal updates, but it did not cover Codex's human-readable status text events. Those text events are separate `message` events, so payloads such as `Goal active` and `Goal active · 8016 tokens` could still both be persisted and rendered.
- #651 fixed the reducer semantics for completed goals: if the reducer sees a later normal user message, `goal complete` should no longer be treated as the current status. That logic was correct, but the status-bar calculation was still fed from `SessionChat`'s timeline-visible message list.

## Root cause

There were two remaining gaps:

1. **Duplicate timeline status events used a different event shape.**
   The previous de-dupe logic targeted structured goal updates, while Codex can also emit plain status text events. Since those text events were not recognized as redundant goal status notifications, they could be stored, broadcast, and displayed as extra timeline blocks.

2. **The bottom-right goal status used a timeline-only view of messages.**
   `SessionChat` intentionally hides queued optimistic user messages from the rendered timeline because they belong in the queued-message UI until the CLI ack arrives. Also, when the message window is not at the bottom, newer user messages can live in the `pending` bucket. As a result, #651's reducer rule could miss the later non-goal user message, so an old `goal complete` indicator stayed visible after the user sent a new message.

## Solution

- Add a shared helper to recognize redundant goal status text notifications.
- Drop redundant goal status text events at hub ingress, before they are persisted or broadcast.
- Filter historical message reads so old persisted redundant goal status rows do not resurface.
- Add web-side filtering for any remaining redundant goal status blocks in nested timeline output.
- Expose pending messages from `useMessages` and let `SessionChat` build a separate goal-state message list from visible + queued/pending known messages.
- Use that goal-state list only for `latestGoal`; timeline rendering continues to use the existing visible-only list.
- Exclude scheduled, uninvoked user messages from goal-state clearing so scheduled prompts do not clear a completed goal before the CLI actually invokes them.

## Tests

- `cd hub && bun run test -- src/socket/handlers/cli/sessionHandlers.test.ts src/sync/messageService.test.ts`
- `cd web && bun run test -- src/chat/reducer.test.ts src/components/SessionChat.test.ts src/lib/message-window-store.test.ts`
- `bun typecheck`

### DIFF
--- a/hub/src/socket/handlers/cli/sessionHandlers.test.ts
+++ b/hub/src/socket/handlers/cli/sessionHandlers.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from 'bun:test'
+import { Store, type StoredSession } from '../../../store'
+import type { SyncEvent } from '../../../sync/syncEngine'
+import type { CliSocketWithData } from '../../socketTypes'
+import { registerSessionHandlers } from './sessionHandlers'
+
+class FakeSocket {
+    readonly roomEvents: Array<{ room: string; event: string; data: unknown }> = []
+    private readonly handlers = new Map<string, (data: unknown) => void>()
+
+    on(event: string, handler: (data: unknown) => void): this {
+        this.handlers.set(event, handler)
+        return this
+    }
+
+    to(room: string): { emit: (event: string, data: unknown) => void } {
+        return {
+            emit: (event: string, data: unknown) => {
+                this.roomEvents.push({ room, event, data })
+            }
+        }
+    }
+
+    trigger(event: string, data: unknown): void {
+        this.handlers.get(event)?.(data)
+    }
+}
+
+function redundantGoalStatusContent(message: string): unknown {
+    return {
+        role: 'agent',
+        content: {
+            id: `event-${message}`,
+            type: 'event',
+            data: { type: 'message', message }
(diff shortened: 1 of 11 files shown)


SmallSpider0:
Addressed the review finding in 6fde35b: `buildGoalStateMessages` now excludes any scheduled user message while it is still uninvoked (`scheduledAt != null && invokedAt == null`), including mature scheduled prompts. Added coverage for a mature scheduled prompt with `scheduledAt < now` and `invokedAt: null`.

PR #657, opened 2026-05-21
fix(web): preserve failed inactive send prompts

oodavy41:
## Summary

Preserve the user’s prompt when sending to an inactive/invalid session fails during session resolution/resume.

Previously, if `resolveSessionId` failed before the send mutation started, the composer had already cleared the input, so a 500/502 resume/send failure could make the user’s prompt disappear with no visible recovery path.

## Changes

- On `resolveSessionId` failure before send, append a failed optimistic user bubble containing the original prompt.
- Keep the prompt visible/copyable/retryable instead of losing it.
- Preserve `scheduledAt` and attachments in the failed bubble metadata/content.
- Update retry behavior so failed preserved prompts re-run `resolveSessionId` before sending.
- Restore failed status if retry-time resolution fails.
- Add regression coverage for:
  - failed resolve preserving a prompt bubble;
  - retry resolving an inactive session before sending;
  - scheduled retry still preserving `scheduledAt`.

## Notes

Known pre-existing limitation: retrying a failed bubble does not currently re-send attachments. This PR preserves attachments visually in the failed bubble but does not change the broader retry attachment behavior.

## Validation

- `git diff --check origin/main..HEAD`
- `bun typecheck`
- `bun --cwd web test src/hooks/mutations/useSendMessage.test.tsx`
- `env -u HAPI_CLI_EXECUTABLE bun run test`

Full suite passed.

### DIFF
--- a/web/src/hooks/mutations/useSendMessage.test.tsx
+++ b/web/src/hooks/mutations/useSendMessage.test.tsx
@@ -158,21 +158,48 @@ describe('useSendMessage', () => {
         await expect(acceptedPromise!).resolves.toBe(false)
     })
 
-    it('resolves false when resolveSessionId throws (inactive-session resume failure)', async () => {
-        const api = createMockApi()
+    it('resolves false and keeps a failed prompt bubble when resolveSessionId throws', async () => {
+        const sendMock = vi.fn()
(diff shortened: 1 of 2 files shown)


tiann:

Issue #686, opened 2026-05-25
feat(voice): voice picker in settings — select from account voices including clones


sterlingchad:
## Problem

The voice assistant currently uses a hardcoded list of ElevenLabs pre-made voices with no way to select between them in the UI. Users who:

- Have an ElevenLabs paid plan and have created **instant voice clones** cannot use those clones with Hapi
- Want a different pre-made voice cannot change it without editing config manually

## Proposed solution

Add a voice picker to Settings → Voice Assistant that:

1. Fetches the user's available voices dynamically from ElevenLabs (`GET /v1/voices`) via a hub proxy route, so cloned voices appear alongside pre-made voices
2. Shows a **▶ preview button** per voice using the ElevenLabs `preview_url` (CDN-hosted, no auth needed to play)
3. Labels cloned voices distinctly so users can identify them
4. Persists the selection to `localStorage` and passes it as an override when starting a voice session (ElevenLabs already supports `platform_settings.overrides.conversation_config_override.tts.voice_id`)
5. Degrades gracefully — if no API key is configured, shows a static fallback list

## Notes

- ElevenLabs voice override is already supported via `overrides.tts.voiceId` in `startSession`
- The hub proxy approach keeps the API key server-side (consistent with the existing token endpoint pattern)
- This lays the groundwork for a future phase where the voice provider can be abstracted (local TTS, Gemini voice, etc.)

heavygee:
Related Discussion (roadmap/context): https://github.com/tiann/hapi/discussions/691

heavygee:
Closing as addressed by #690 (merged 2026-05-27), which added the dynamic ElevenLabs voice picker in Settings — fetching account voices including instant clones, with a preview button on each. Per-backend voice selection for Gemini Live and Qwen Realtime is tracked in #742 / PR #743.

PR #690, opened 2026-05-25
feat(voice): dynamic settings voice picker with safe fallback + preview

heavygee:
## Summary

- add a Settings voice picker (Default + selectable ElevenLabs voices)
- load voices dynamically from a new authenticated hub route: `GET /api/voice/voices`
- include preview controls in the picker:
  - enabled when `preview_url` is available
  - visible but disabled with tooltip when unavailable (including no-key fallback)
- label cloned voices with a `clone` badge
- keep no-key behavior safe:
  - static fallback voice list remains available
  - starting voice session without key still returns clear 400 error
- route voice selection via per-voice agent/token selection (no fragile runtime voice override)
- localize fallback voice display names for zh-CN
- add/extend tests for settings picker and hub voice routes

Refs #686

## Test plan

- [x] `cd web && bun run test src/routes/settings/index.test.tsx src/lib/voices.test.ts`
- [x] `cd hub && bun test src/web/routes/voice.test.ts`
- [x] Manual:
  - [x] default voice path works
  - [x] non-default premade voice (Alice) works
  - [x] cloned voice works
  - [x] no API key path shows static list, disabled preview controls, and clear 400 toast on voice start

### DIFF
--- a/cli/src/utils/spawnHappyCLI.test.ts
+++ b/cli/src/utils/spawnHappyCLI.test.ts
@@ -157,6 +157,8 @@ describe('spawnHappyCLI windowsHide behavior', () => {
 
   it('falls back to a real argv0 executable before process.execPath in compiled mode', async () => {
     isBunCompiledMock.mockReturnValue(true);
+    const previousCliExecutable = process.env.HAPI_CLI_EXECUTABLE;
+    delete process.env.HAPI_CLI_EXECUTABLE;
     const previousArgv0 = process.argv[0];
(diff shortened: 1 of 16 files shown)


heavygee:
Linking related upstream Discussion for roadmap context: https://github.com/tiann/hapi/discussions/691

heavygee:
Happy to rebase this onto post-#692 main once that lands — the two PRs are complementary (voice picker + multi-backend are separate concerns) but share enough files that a clean rebase makes more sense than trying to merge both simultaneously. No action needed here until #692 merges.

Issue #697, opened 2026-05-26
Support displaying images returned by agents


NightWatcher314:
It would be great if Hapi could support displaying images returned by coding agents directly in the chat UI, similar to the image/artifact display experience in apps like Codex or Claude.

Use cases:
- An agent generates a screenshot, diagram, chart, mockup, or other image and sends it back to the user.
- The user can view the image inline in the session instead of only seeing a local file path or textual description.
- This would be especially useful when controlling agents remotely from the web/mobile UI.

No specific implementation requested here — just hoping Hapi can support image outputs as first-class visible chat content.

heavygee:
Follow-up: scoped actionable work in #956 (cross-flavor gaps after #700 — MCP prompts for Cursor/Gemini/Kimi/Pi, ACP image block handling, #508 tool results). #700 was not Claude-only but coverage is incomplete.

tiann:
Closed as outdated after auditing current `main` (223be6d) — the problem described here is already resolved in code. Evidence below.

Implemented by PR #700 (`de5dc97`): a `display_image` MCP tool is wired into all agent paths — Claude (`cli/src/claude/utils/startHappyServer.ts:97`), Codex (`cli/src/codex/happyMcpStdioBridge.ts:113`), and generic ACP agents via the MCP bridge (`cli/src/agent/runners/runAgentSession.ts:74-99`). Web renders returned images inline via `GeneratedImageCard` (`web/src/components/AssistantChat/messages/ToolMessage.tsx:52`), with follow-up reliability fixes merged (#927/#934, #944, lightbox #715). Remaining cross-flavor gaps (Cursor/Kimi prompts, ACP image blocks) are tracked in #956, which stays open.

PR #700, opened 2026-05-26
Add display_image MCP tool for inline images

NightWatcher314:
## Summary
- add a HAPI MCP `display_image` tool that registers a local image as an existing `generated-image` chat payload
- forward `display_image` through the stdio MCP bridge
- update Codex, Claude, and OpenCode prompt hints so agents know to call the tool for local images

Refs #697

## Test
- `bun run typecheck:cli`


NightWatcher314:
图片

Issue #708, opened 2026-05-26
fix(web): TanStack scroll restoration fills sessionStorage and can crash the SPA


heavygee:
## Summary

Under load (long sessions, many navigations, heavy scroll targets), `sessionStorage.setItem` for TanStack Router key `tsr-scroll-restoration-v1_3` can throw **QuotaExceededError** / DOMException-shaped quota errors. That bubbles into React and can take down the full UI with:

`Failed to execute 'setItem' on 'Storage': Setting the value of 'tsr-scroll-restoration-v1_3' exceeded the quota.`

HAPI already wraps `sessionStorage.setItem` in `web/src/lib/scrollStorageGuard.ts` (see #611), but a **prune-only** recovery is insufficient: TanStack keeps the full scroll map **in memory**; trimming the JSON written to disk does not shrink the in-RAM cache, so the next persist can throw again immediately.

## Proposed direction (implemented in PR)

- **PR:** https://github.com/tiann/hapi/pull/707  
- On persist failure for the scroll restoration key (after prune+retry fails, or invalid JSON), clear storage and reset TanStack `scrollRestorationCache` when guarding real `window.sessionStorage` (unit tests use a mock `Storage` and skip the library reset).
- Treat **any** write failure on that key as recoverable for the scroll cache (not only `instanceof Error` quota shapes).

Follow-up optional hardening: bump `@tanstack/react-router` / `@tanstack/router-core` when upstream wraps `setItem` in try/catch (belt and suspenders).

## Suggested verification

1. From a checkout that includes the merged fix: `cd web && bun run test src/lib/scrollStorageGuard.test.ts && bun run typecheck`
2. Manual: long session chat, scroll, navigate until storage is stressed; confirm no uncaught `setItem` / no white-screen error boundary.

PR #544, opened 2026-04-28
fix(codex): improve web rendering for Codex events

oodavy41:
## Problem

Codex App Server events are not fully normalized for the HAPI web UI. Several Codex-specific updates either lose useful context or render as generic/raw output, including plan updates, MCP tool calls, terminal/context details, token usage, and status context.

These rendering paths are also difficult to validate manually without realistic fixture sessions and a repeatable local web smoke test.

## Solution

Improves Codex event conversion, web presentation, and local validation support:

- Surfaces Codex plan updates in the web timeline.
- Maps Codex MCP tool calls into known web tool rendering paths.
- Improves terminal and context display for Codex sessions.
- Formats Codex token usage events for readable web output.
- Improves Codex status context display.
- Adds normalization/presentation coverage for the new event shapes.
- Adds a dev fixture and headless web reader for repeatable Codex web validation.

## Tests

- `bun install`
- `bun typecheck`
- `bun run test`
- `bun run build:web`
- `git diff --check origin/main..HEAD`
- `scripts/dev/self-test-codex-web-env.sh`

PR #717, opened 2026-05-27
fix(web): scroll guard setItem unwrap race (#716)

### LINKED ISSUE #716: fix(web): #707 scroll guard still throws uncaught QuotaExceededError (setItem unwrap race)
## Summary

**PR #707 (`Fixes #708`) is merged but insufficient.** Uncaught `QuotaExceededError` on `tsr-scroll-restoration-v1_3` still occurs in production on the current web bundle, tripping React's error boundary during normal session chat use.

This is a follow-up to the scroll-restoration quota chain:

| Item | Link | What it did |
|------|------|-------------|
| Original report | #611 | `sessionStorage` quota exhaustion on TanStack scroll key crashes SPA |
| Pathname key collapse | #632 | `getScrollRestorationKey` → pathname to slow cache growth |
| In-memory cache gap | #708 | Prune-only guard left TanStack RAM cache oversized |
| Merged fix | #707 | Hard-reset + sync in-memory cache via `scrollRestorationCache.set` |

The #707 fix introduced a **temporary unwrap** of `sessionStorage.setItem` that creates a race: TanStack can call `scrollRestorationCache.set` (which calls native `setItem`) while the guard is bypassed.

---

## Evidence (production, not stale cache)

Observed on `https://hapi.tail9944ee.ts.net` after driver rebuild serving bundle **`index-BoTBCCy9.js`** (bundle contains `__hapiScrollRestorationGuard` — guard is installed).

**Console log excerpt** (session chat, 2026-05-27):

```
Uncaught QuotaExceededError: Failed to execute 'setItem' on 'Storage':
Setting the value of 'tsr-scroll-restoration-v1_3' exceeded the quota.
    at Object.set (index-BoTBCCy9.js:10:94081)
    at s (index-BoTBCCy9.js:10:95765)
    ...
setTimeout
scroll
```

Second occurrence (caught by error boundary / React DevTools hook):

```
QuotaExceededError: ... 'tsr-scroll-restoration-v1_3' ...
    at Object.set (index-BoTBCCy9.js:10:94081)
    at Object.fn (index-BoTBCCy9.js:10:96489)
    at Set.forEach
    at lR.emit
    at ref
```

**Interpretation of stacks:**

1. **`Object.set`** = TanStack `scrollRestorationCache.set` in `@tanstack/router-core/src/scroll-restoration.ts` (line 71: `safeSessionStorage.setItem(storageKey, JSON.stringify(state))`).
2. **`setTimeout` → `scroll`** = throttled `document` scroll listener in `setupScrollRestoration` (100ms throttle).
3. **`emit` → `ref`** = router `onRendered` subscriber also calling `scrollRestorationCache.set` to mark location seen.

These paths call `scrollRestorationCache.set` **directly**, not only through the wrapped `sessionStorage.setItem` entry point.

HAR capture confirms the same JS bundle (`index-BoTBCCy9.js`) was loaded for the failing session URL.

---

## Root cause

`web/src/lib/scrollStorageGuard.ts` — `writeScrollRestorationCache()`:

```typescript
function writeScrollRestorationCache(storage, originalSetItem, updater) {
    const guardedSetItem = storage.setItem
    storage.setItem = originalSetItem   // <-- temporarily removes guard
    try {
        scrollRestorationCache?.set(updater)
    } finally {
        storage.setItem = guardedSetItem
    }
}
```

Called from:

- Successful prune path (`wrappedSetItem`, after trimmed write succeeds)
- Hard-reset path (`hardResetScrollRestorationPersistedState`)

**Problem:** While `storage.setItem === originalSetItem`, any concurrent TanStack scroll write (`scrollRestorationCache.set` from throttled scroll handler or `onRendered`) hits **unwrapped native `setItem`**. On quota exhaustion that throw is **uncaught** — it never enters `wrappedSetItem`'s recovery logic.

The existing unit test `hard reset calls scrollRestorationCache through unwrapped setItem` documents that unwrap is intentional for in-memory sync, but there is **no protection against concurrent TanStack writes during the unwrap window**.

Additional unit test added locally (not yet upstream):

`exposes unwrap window where concurrent TanStack scroll writes can throw uncaught` — simulates TanStack calling `setItem` while `writeScrollRestorationCache` has unwrapped; demonstrates uncaught throw path.

---

## Why #707 manual test plan item was never satisfied

PR #707 test plan included:

> Manual: long session chat, scroll, navigate until storage is stressed; confirm no uncaught setItem / no white-screen error boundary

That step was **unchecked** at merge. Playwright stress tests against local + production hub with artificial `sessionStorage` fill often pass (guard handles wrapped-path recovery), but the race is **intermittent** and matches real operator profiles with accumulated storage + active scrolling during recovery.

---

## Suggested fix directions (for follow-up PR)

Do **not** temporarily assign `storage.setItem = originalSetItem`. Options (pick one, verify with tests):

1. **Patch `scrollRestorationCache.set` directly** — update in-memory state without persisting, then persist once through the wrapped `setItem` (or call `originalSetItem` only from inside `wrappedSetItem` where recovery already handles failures).

2. **Re-entrant lock / depth counter** — if guard recovery is in progress, route all `scrollRestorationCache.set` persistence through `wrappedSetItem` even when syncing RAM (never expose native `setItem` on the shared `sessionStorage` object).

3. **Upstream belt-and-suspenders** — #683 tracks upgrading `@tanstack/react-router`; upstream may wrap `setItem` in try/catch, but HAPI should not rely on that alone given the unwrap race.

4. **Reduce concurrent writers** — optional: disable or debounce scroll restoration persist while hard-reset runs (harder to reason about; prefer fixing unwrap).

---

## Reproduction / verification checklist for fix PR

```bash
# Unit tests (extend scrollStorageGuard.test.ts)
cd web && bun run test src/lib/scrollStorageGuard.test.ts

# Playwright stress (repo root; uses system Chrome on Linux)
HAPI_URL=https://your-hub HAPI_ACCESS_TOKEN=... \
  node scripts/dev/scroll-quota-repro-playwright.mjs \
  --session <session-id> --fill-mb 4.5

# Must add: concurrent-write race test (see test name above)
# Must pass: no pageerror/console QuotaExceededError on tsr-scroll-restoration-v1_3
```

**Manual (required):** Long-lived browser profile with stressed `sessionStorage`, session chat open, scroll + navigate while storage near quota — confirm **zero** uncaught `QuotaExceededError` and no React error boundary.

---

## Related

- #611 — original quota crash
- #632 — pathname scroll key (reduces growth, does not fix guard race)
- #708 — in-memory cache not reset by prune-only guard
- #707 — merged fix (insufficient — this issue)
- #683 — router upgrade (optional additional hardening)

## Files

- `web/src/lib/scrollStorageGuard.ts` — bug location (`writeScrollRestorationCache`)
- `web/src/main.tsx` — `installScrollRestorationGuard()` at bootstrap (guard IS installed; not a boot-order bug)
- `node_modules/@tanstack/router-core/src/scroll-restoration.ts` — TanStack persist + scroll listeners
- `web/src/lib/scrollStorageGuard.test.ts` — existing + proposed race test

heavygee:
## Summary

- **#707 regression:** `writeScrollRestorationCache` temporarily set `sessionStorage.setItem = originalSetItem` while syncing TanStack's in-memory scroll map. Concurrent `scrollRestorationCache.set` calls (throttled scroll handler, router `onRendered`) could hit **unwrapped native `setItem`** during that window and throw uncaught `QuotaExceededError` (#716).
- **Fix:** Remove the unwrap entirely. At bootstrap, patch `scrollRestorationCache.set` to update local cache state and persist **only** through the guarded `storage.setItem` wrapper. Hard-reset and post-prune sync both use the patched setter; a `recoveryDepth` guard prevents nested hard-reset loops (nested fallback uses `originalSetItem`, not the wrapper).
- **Tests:** 17 unit tests (15 guard + 2 legacy positive repro).

## Verification

Full write-up: https://github.com/tiann/hapi/issues/716#issuecomment-4560444570

- **Positive unit repro (deterministic):** `scrollStorageGuard.legacy-regression.test.ts` re-implements the #707 unwrap window and proves it throws `QuotaExceededError` through native `setItem`; the #717 patched path survives the same quota pressure.
- **Live upstream/main stress (intermittent):** Playwright against parallel local hubs filled `sessionStorage` to quota and stress-scrolled. No uncaught `pageerror` every run — the production bug is a timing race. #716 production stacks and stack traces remain the real-world proof upstream was broken.
- **Dev harnesses:** `scripts/dev/scroll-quota-repro-playwright.mjs`, `scroll-quota-prove-positive.mjs`, `scroll-quota-ab-compare.mjs`. Same `scripts/dev/` pattern as `read-hapi-web.mjs` (#544); defaults to `http://127.0.0.1:3006`. Not wired to CI.
- **Operator deploy:** fix built and tested on a private stack before merge; no white-screen recurrence under normal session chat use.

## Test plan

- [x] `cd web && bun run test src/lib/scrollStorageGuard.test.ts src/lib/scrollStorageGuard.legacy-regression.test.ts` (17/17 pass)
- [x] `bun typecheck` (repo root)
- [x] Playwright dev scripts committed (`scripts/dev/scroll-quota-*.mjs`)
- [ ] Manual: long-lived browser profile, scroll + navigate near quota

## Issues

Fixes #716

### DIFF
--- a/scripts/dev/scroll-quota-ab-compare.mjs
+++ b/scripts/dev/scroll-quota-ab-compare.mjs
@@ -1,0 +1,118 @@
+#!/usr/bin/env node
+/**
+ * A/B compare scroll quota stress: fixed vs upstream-main (broken #707 unwrap).
+ *
+ * Usage:
+ *   node scripts/dev/scroll-quota-ab-compare.mjs --session <id>
+ *
+ * Env:
+ *   HAPI_ACCESS_TOKEN — CLI token for auth
+ *   HAPI_URL_FIXED — default http://127.0.0.1:3006 (PR #717 worktree)
+ *   HAPI_URL_BROKEN — default http://127.0.0.1:3007 (upstream/main hub)
+ */
+import { spawnSync } from 'node:child_process'
+import { resolve, dirname } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { writeFileSync, mkdirSync } from 'node:fs'
+
+const __dir = dirname(fileURLToPath(import.meta.url))
+const repro = resolve(__dir, 'scroll-quota-repro-playwright.mjs')
+
+function parseArgs(argv) {
+    const args = { sessionId: '', fillMb: 4.5, scrollRoutes: 250, navRounds: 12 }
+    for (let i = 0; i < argv.length; i += 1) {
+        const arg = argv[i]
+        if (arg === '--session') args.sessionId = argv[++i]
+        else if (arg === '--fill-mb') args.fillMb = Number(argv[++i])
+        else if (arg === '--scroll-routes') args.scrollRoutes = Number(argv[++i])
+        else if (arg === '--nav-rounds') args.navRounds = Number(argv[++i])
+    }
+    return args
+}
+
+const args = parseArgs(process.argv.slice(2))
+const token = process.env.HAPI_ACCESS_TOKEN ?? ''
+const fixedUrl = process.env.HAPI_URL_FIXED ?? 'http://127.0.0.1:3006'
+const brokenUrl = process.env.HAPI_URL_BROKEN ?? 'http://127.0.0.1:3007'
+
+function runVariant(label, baseUrl) {
+    const env = {
+        ...process.env,
+        HAPI_URL: baseUrl,
+        HAPI_ACCESS_TOKEN: token,
+    }
+    const proc = spawnSync(
+        process.execPath,
+        [
+            repro,
+            '--session', args.sessionId,
+            '--fill-mb', String(args.fillMb),
+            '--scroll-routes', String(args.scrollRoutes),
+            '--nav-rounds', String(args.navRounds),
+        ],
+        { env, encoding: 'utf8', maxBuffer: 20 * 1024 * 1024 },
+    )
+    let parsed = null
+    try {
+        const jsonStart = proc.stdout.lastIndexOf('{')
+        parsed = JSON.parse(proc.stdout.slice(jsonStart))
+    } catch {
+        parsed = { ok: false, parseError: true, stdout: proc.stdout.slice(-2000), stderr: proc.stderr }
+    }
+    return { label, baseUrl, exitCode: proc.status ?? 1, result: parsed }
+}
+
+if (!args.sessionId) {
+    console.error('Usage: scroll-quota-ab-compare.mjs --session <id>')
+    process.exit(2)
+}
+
+console.log('Running A/B scroll quota stress...')
+console.log(`  FIXED  → ${fixedUrl}`)
+console.log(`  BROKEN → ${brokenUrl}`)
+console.log('')
+
+const broken = runVariant('upstream-main (#707 unwrap)', brokenUrl)
+const fixed = runVariant('PR #717 (patched cache.set)', fixedUrl)
+
+const summary = {
+    sessionId: args.sessionId,
+    params: args,
+    broken,
+    fixed,
+    regressionDetected: Boolean(broken.result?.ok === false && (broken.result?.quotaErrors?.length || broken.result?.pageErrors?.length)),
+    fixVerified: Boolean(fixed.result?.ok === true && !(fixed.result?.quotaErrors?.length)),
+}
+
+mkdirSync(resolve('localdocs/playwright-runs'), { recursive: true })
+const out = resolve('localdocs/playwright-runs', `scroll-quota-ab-${Date.now()}.json`)
+writeFileSync(out, JSON.stringify(summary, null, 2))
+
+console.log(JSON.stringify({
+    broken: {
+        ok: broken.result?.ok,
+        quotaErrors: broken.result?.quotaErrors,
+        pageErrors: broken.result?.pageErrors,
+        scrollKeyAfter: broken.result?.scrollKeyAfter,
+        exitCode: broken.exitCode,
+    },
+    fixed: {
+        ok: fixed.result?.ok,
+        quotaErrors: fixed.result?.quotaErrors,
+        pageErrors: fixed.result?.pageErrors,
+        scrollKeyAfter: fixed.result?.scrollKeyAfter,
+        exitCode: fixed.exitCode,
+    },
+    regressionDetected: summary.regressionDetected,
+    fixVerified: summary.fixVerified,
+    report: out,
+}, null, 2))
+
+if (!summary.regressionDetected) {
+    console.error('\nNOTE: upstream-main did NOT fail this harness — may need harsher race trigger or pre-#707 tree.')
+    process.exitCode = 3
+} else if (!summary.fixVerified) {
+    process.exitCode = 2
+} else {
+    process.exitCode = 0
+}
--- a/scripts/dev/scroll-quota-ab-compare.mjs
+++ b/scripts/dev/scroll-quota-ab-compare.mjs
@@ -7,8 +7,8 @@
  *
  * Env:
  *   HAPI_ACCESS_TOKEN — CLI token for auth
- *   HAPI_URL_FIXED — default http://127.0.0.1:3006 (PR #717 worktree)
- *   HAPI_URL_BROKEN — default http://127.0.0.1:3007 (upstream/main hub)
+ *   HAPI_URL_FIXED — default http://127.0.0.1:3006 (fixed build)
+ *   HAPI_URL_BROKEN — default http://127.0.0.1:3007 (upstream/main build)
  */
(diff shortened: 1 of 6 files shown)


swear01:
Hi @heavygee, thanks for the thorough fix and the regression repros — the `recoveryDepth` guard is a nice touch.

Just a heads-up: there's an overlapping PR #722 that takes a different approach — upgrading `@tanstack/react-router` to `^1.170.8`. In `@tanstack/router-core` >=1.145.6, `scrollRestorationCache` is no longer part of the public exports, so the in-memory cache sync path (and the unwrap race it introduced) goes away entirely. Upstream also ships its own `try-catch` around `setItem` from that version, so crashes are covered at the library level too.

The tradeoff with #722 is that after a prune the in-memory cache may be momentarily stale (scroll position could be wrong on the next nav), whereas #717 keeps it in sync. If the maintainer prefers the sync guarantee, #717 is the right call — but it would need to account for `scrollRestorationCache` being missing in newer versions.

Happy to coordinate whichever direction gets picked.

heavygee:
Hi @swear01, thanks for the heads-up and for laying out the tradeoffs so clearly. Really helpful.

I wasn't aware of #722 when I opened this; good to know there's another path in flight.

My read (could be wrong): #717 was mostly trying to close the **#707 regression**, the temporary `setItem` unwrap that let concurrent TanStack writes hit native storage and throw uncaught (#716). The `recoveryDepth` / cache.set patch was my attempt to keep **RAM and disk in sync after a prune**, which I *think* was the gap behind #708, but I'm not deeply familiar with TanStack's internals and may be overfitting to what I saw in prod.

If #722's upgrade removes the public `scrollRestorationCache` export and wraps persist in try/catch, that sounds like it **eliminates the failure mode I patched** at the source, which I'd honestly prefer to fighting library internals I don't fully understand. I suspect my cache.set patch **wouldn't survive** that bump anyway, so I'm not attached to that specific mechanism.

The bit I'm less sure about: whether prune-only (without clearing whatever the library keeps in memory on newer versions) is enough to stop **repeat quota pressure**, or if we'd mostly be trading crashes for silent persist failures / occasional wrong scroll. I haven't tested against 1.170+ myself. That's speculation on my part.

Happy to go whichever way the maintainer prefers:
- **#717 as a short-term fix** on the current pin if they want something targeted now, or
- **rebase/coordinate toward #722** and trim this down to proactive pruning + tests on the new API (`storageKey` etc.), dropping the cache patch.

I can help with either direction or close #717 if #722 covers it. Just didn't want #716 to sit open while I figured out overlap.

Thanks again for offering to coordinate.

swear01:
Thanks for the detailed breakdown — really appreciate the openness.

To answer your question about repeat quota pressure: in >=1.145.6, TanStack only calls `setItem` once per page lifetime — in the [`pagehide` event listener (L237–L244)](https://github.com/TanStack/router/blob/5268ba4566233ea58880df85f167ad0401a93a46/packages/router-core/src/scroll-restoration.ts#L237-L244), via [`persistScrollRestorationCache` (L42–L52)](https://github.com/TanStack/router/blob/5268ba4566233ea58880df85f167ad0401a93a46/packages/router-core/src/scroll-restoration.ts#L42-L52). Navigations only update the in-memory `scrollRestorationCache` object — no repeated write pressure during a session.

At pagehide, the in-memory cache may have grown large, so our guard intercepts that single write, prunes to 50 entries, and retries — sessionStorage ends up with the 50 most recent positions instead of silently losing everything (which is what upstream's own try-catch does).

The scroll tradeoff is narrow: during a session `onRendered` [reads directly from the in-memory cache (L320)](https://github.com/TanStack/router/blob/5268ba4566233ea58880df85f167ad0401a93a46/packages/router-core/src/scroll-restoration.ts#L320), not from sessionStorage, so scroll restoration within a session is unaffected. The pruned entries only matter after a full page reload, where a pruned route scrolls to 0 — same as pre-#707 behaviour.

I think #722 covers the failure mode you were targeting. Feel free to close #717 — and thanks again for the regression tests and repro scripts, they're useful reference regardless of which approach lands.

heavygee:
Closing in favour of #722 — @swear01's breakdown convinced me the router upgrade is the better long-term fix, and my cache.set patch would not survive that bump anyway.

Thanks again for the thoughtful review and for answering the repeat-quota question; really appreciated.

**What I am doing locally:** keeping `fix/web-scroll-guard-unwrap-race` as a **driver-manifest branch layer** until #722 lands on `upstream/main`, then dropping it and rebuilding soup from main. Daily driver hit the full white-screen `QuotaExceededError` again without that layer (expected on current upstream #707 code).

Happy to help if anything from the regression tests or repro scripts is useful for #722.

PR #733, opened 2026-05-30
feat(antigravity): add Antigravity CLI (agy) integration

swear01:
## Summary

Adds first-class support for Google's **Antigravity CLI** (`agy`) as a new agent flavor in HAPI, alongside Claude / Codex / Cursor / Gemini / Kimi / OpenCode.

Two commits:

1. **`feat(antigravity): add Antigravity CLI (agy) integration`** — full integration
2. **`fix(antigravity): audit fixes`** — flavor labels, sessionId change detection, permission mode tests

## What's added

**`shared/`**
- `antigravity` added to `AGENT_FLAVORS`
- `AntigravityPermissionMode` + `ANTIGRAVITY_PERMISSION_MODES` (`'default' | 'yolo' | 'sandbox'`)
- New `'sandbox'` member in `PERMISSION_MODES`
- `DEPRECATED_FLAVORS = ['gemini']` (Gemini CLI EOL June 18; Antigravity is the suggested successor)
- `antigravitySessionId` on `MetadataSchema`
- `isCodexFamilyFlavor('antigravity')` returns true

**`cli/`**
- `cli/src/antigravity/` — `runAntigravity`, `antigravityLocal`, `antigravityLocalLauncher`, `loop`, `session`, `utils/sessionScanner`, `utils/config`, `types`
- Session detection via watcher on `~/.gemini/antigravity-cli/implicit/*.pb`
- Resume via `--conversation <UUID>`
- Yolo via `--dangerously-skip-permissions`; sandbox via `--sandbox`
- New `hapi antigravity` command
- Resume dispatcher branch + `buildCliArgs` branch
- Slash command builtins: `/model`, `/clear`, `/compress`, `/about`, `/stats`, `/help`, `/history`, `/memory`, `/quit`, `/reset`, `/tools`, `/feedback`

**`hub/`**
- `resolveAgentResumeId` handles the antigravity flavor
- `hasSameAgentSessionIds` now compares `antigravitySessionId` (and previously-missed `kimiSessionId`) so dedup fires correctly when these IDs land in metadata

**`web/`**
- `Ag` badge (indigo `#4f46e5`) in `AgentFlavorIcon`
- `antigravity` entry in `MODEL_OPTIONS`
- `AgentSelector` uses `getFlavorLabel()` so Gemini shows as "Gemini (deprecated)" in the option list
- Amber banner under `AgentSelector` when a deprecated flavor is selected

## Test plan

- [x] `npm run typecheck` — clean across cli / web / hub
- [x] `shared && bun test` — 55/55 pass
- [ ] `cd cli && npm test` — (792+ tests; 5 integration tests are pre-existing failures per local notes)
- [ ] Manual smoke: install `agy`, run `hapi antigravity`, verify session shows up in web UI with `Ag` badge
- [ ] Manual smoke: resume an antigravity session via `hapi resume`
- [ ] Manual smoke: select Gemini in new-session UI, confirm "Gemini (deprecated)" label + amber banner

via [HAPI](https://hapi.run)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ execplan/
 # Generated npm bundle output (local)
 cli/npm/main/
 .ace-tool/
+
+# Antigravity CLI project directory (created by agy on first run)
+.antigravitycli/
--- a/cli/src/antigravity/antigravityLocal.ts
+++ b/cli/src/antigravity/antigravityLocal.ts
@@ -1,0 +1,43 @@
+import { logger } from '@/ui/logger'
+import { spawnWithTerminalGuard } from '@/utils/spawnWithTerminalGuard'
+import { buildAntigravityEnv } from './utils/config'
+import type { PermissionMode } from './types'
+
+export async function antigravityLocal(opts: {
+    path: string;
+    sessionId: string | null;
+    abort: AbortSignal;
+    permissionMode?: PermissionMode;
+}): Promise<void> {
+    const args: string[] = []
+
+    if (opts.sessionId) {
+        args.push('--conversation', opts.sessionId)
+    }
+
+    if (opts.permissionMode === 'yolo') {
+        args.push('--dangerously-skip-permissions')
+    } else if (opts.permissionMode === 'sandbox') {
+        args.push('--sandbox')
+    }
+
+    args.push('--add-dir', opts.path)
+
+    const env = buildAntigravityEnv()
+
+    logger.debug(`[AntigravityLocal] Spawning agy with args: ${JSON.stringify(args)}`)
+
+    await spawnWithTerminalGuard({
+        command: 'agy',
+        args,
+        cwd: opts.path,
+        env,
+        signal: opts.abort,
+        shell: process.platform === 'win32',
+        logLabel: 'AntigravityLocal',
+        spawnName: 'agy',
+        installHint: 'Antigravity CLI',
+        includeCause: true,
+        logExit: true
+    })
+}
--- a/cli/src/antigravity/antigravityLocalLauncher.ts
+++ b/cli/src/antigravity/antigravityLocalLauncher.ts
@@ -1,0 +1,42 @@
+import { antigravityLocal } from './antigravityLocal'
+import { AntigravitySession } from './session'
+import { createAntigravitySessionScanner } from './utils/sessionScanner'
+import { BaseLocalLauncher } from '@/modules/common/launcher/BaseLocalLauncher'
+
+export async function antigravityLocalLauncher(
+    session: AntigravitySession
+): Promise<'switch' | 'exit'> {
+    const scanner = createAntigravitySessionScanner({
+        existingSessionId: session.sessionId,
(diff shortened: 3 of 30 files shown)


swear01:
Closing this PR for now.

The current Antigravity CLI integration is blocked by a design mismatch rather than a small implementation bug. `agy` is a TUI-first CLI that requires an interactive `/dev/tty`, while the HAPI web runner may run daemonized without a controlling terminal. In that mode, `agy` exits before writing an `antigravitySessionId`, which causes `/resume` to fail with a generic error.

Adding `node-pty` alone would not make this a correct web backend, because HAPI would still need to bridge web messages into the TUI, stream ANSI terminal output back to the browser, handle resize/reconnect/approval prompts, and define session resume semantics. That is a larger terminal-hosting project, not a small backend integration.

### Root cause summary (so we don't rediscover it)

- `agy` (Antigravity CLI, v1.x) is Bubble Tea–based and requires an interactive `/dev/tty`.
- HAPI web runners commonly launch detached/daemonized with no controlling terminal.
- Under no-TTY conditions `agy` exits immediately with a Bubble Tea TTY error.
- Because `agy` exits before producing a `.pb` session file, no `antigravitySessionId` is recorded.
- `/resume` then has nothing to resume and surfaces a generic 500 / "Resume session ID unavailable".
- `agy` v1.x also has **no `--model` CLI flag** — model selection is `/model` inside the TUI only — so the web ModelSelector has nothing meaningful to render either.
- `agy` has no ACP / headless / programmatic mode, so HAPI's existing ACP transport cannot be reused; the launcher fell back to a local spawn that doesn't fit the web runner model.

### Why short-term we shouldn't ship a web Antigravity backend

For HAPI's web flow to drive `agy` correctly we would need, at minimum:

- PTY allocation (e.g. `node-pty`) for the child process
- web message queue → TUI keypress synthesis
- TUI ANSI output → web event stream (xterm.js or equivalent)
- terminal resize, scrollback, and reconnect semantics
- approval-prompt forwarding
- a real session lifecycle definition (start / detach / resume / archive) that matches a TUI, not a JSON-RPC agent

That is a 1–2 week terminal-hosting project, not an Antigravity-specific change.

### Recommended path forward (separate work)

A proper future integration should wait for one of:

1. An official `agy` ACP / headless / programmatic mode.
2. An Antigravity **SDK** runner (with the corresponding quota / auth implications spelled out).
3. An intentional full PTY + xterm.js terminal bridge inside HAPI, treated as its own feature.

Until one of those exists, Antigravity CLI should be considered **terminal-only / experimental** and should not be exposed as a normal web-launchable agent backend.

### What this PR leaves behind

- Branch `feat/antigravity-integration` is preserved on the fork for reference (commits `a1bf89c` integration + `ba385ad` audit fixes).
- No code from this PR is merged to `main`, so there is nothing to revert on `main` and the web UI is not in a confusing half-working state.
- Locally-explored UX fixes (e.g. ModelSelector showing a `/model` hint instead of an empty selector) were intentionally **not committed**, since they only make sense on top of a working Antigravity backend.

Closing without merge.

PR #734, opened 2026-05-30
fix(test): isolate integration tests from production hub via temp hub globalSetup

swear01:
## Summary

- Integration tests in `runner.integration.test.ts` were previously connecting to the **production HAPI hub** when the shell env contained real `CLI_API_TOKEN` / `HAPI_API_URL`, because `vitest.config.ts` injected the entire `process.env` into test workers via `env: { ...process.env }`.
- This PR fixes that by spinning up a throwaway hub on a random free port before tests run, and tearing it down afterwards.

## Changes

- **`cli/vitest.config.ts`** — Remove the `dotenv` / `env: { ...process.env }` block that leaked shell credentials into test workers. Add `globalSetup` and `setupFiles`.
- **`cli/src/test/globalSetup.ts`** (new) — Spawns a temporary hub (`hub/src/index.ts`) via `bun run` on a random free port with a generated token and an isolated `HAPI_HOME`. Writes connection config to `/tmp/hapi-test-config.json`; tears down hub and temp dir after all tests.
- **`cli/src/test/setup.ts`** (new) — Runs at the top level of each vitest worker before any test file is imported. Reads the config file and sets `HAPI_API_URL`, `CLI_API_TOKEN`, `HAPI_HOME`, `HAPI_BUN_EXEC`, and `HAPI_RUNNER_HEARTBEAT_INTERVAL` so that module singletons (like `configuration`) pick up the isolated values.
- **`cli/src/utils/spawnHappyCLI.ts`** — Add `HAPI_BUN_EXEC` override so that when vitest runs under Node.js, spawned CLI child processes still use Bun (required for TypeScript entrypoints).
- **`cli/src/utils/spawnHappyCLI.test.ts`** — Save/restore `HAPI_BUN_EXEC` in `beforeEach`; clear it in the test that asserts Node.js fallback behaviour.
- **`cli/src/runner/runner.integration.test.ts`** — Remove two `execSync('yarn build')` calls (dead code: `yarn` is not installed; bun runs TypeScript directly, no compilation step needed).

## Test plan

- [ ] `cd cli && pnpm test` — all unit + integration tests pass against the isolated hub, not production
- [ ] Shell env with real `CLI_API_TOKEN` / `HAPI_API_URL` no longer affects test results
- [ ] `globalSetup` / `teardown` correctly starts and stops the temp hub

via [HAPI](https://hapi.run)

### DIFF
--- a/cli/src/runner/runner.integration.test.ts
+++ b/cli/src/runner/runner.integration.test.ts
@@ -16,7 +16,7 @@
  */
 
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
-import { execSync, spawn } from 'child_process';
+import { spawn } from 'child_process';
 import { existsSync, unlinkSync, readFileSync, writeFileSync, readdirSync } from 'fs';
 import path, { join } from 'path';
 import { configuration } from '@/configuration';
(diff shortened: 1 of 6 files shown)


swear01:
Filed #757 with a concrete repro and the real-world impact: anyone who has their normal `CLI_API_TOKEN` / `HAPI_API_URL` exported (which is how everyone runs the CLI against their own hub) silently runs `pnpm test` against that hub — for many of us that's prod. It writes session/runner rows there before they realise what happened.

That issue is the user-facing motivation; this PR is the fix that closes it. Happy to address any remaining concerns — the last HAPI Bot review round came back with no findings, CI is green, and the branch is mergeable.

Fixes #757

PR #773, opened 2026-06-02
fix(web): retain composer text on send failure (closes #776)

### LINKED ISSUE #776: Bug: composer eats typed text on 4xx/5xx send failure
## Problem

When the message composer submits and the hub responds with a 4xx or 5xx (or the fetch fails outright), the composer **clears the input** -- the operator's typed text is destroyed. This is destructive at exactly the moment the operator most needs their text retained so they can retry without losing what they wrote.

Example: composer fires `POST /api/sessions/:id/messages`, hub returns 503 during a restart -> text discarded.

This is reproducible during any hub restart blip. `@assistant-ui/react`'s composer clears `composer.text` synchronously the moment send is invoked, so by the time `useSendMessage`'s mutation rejects the input is already gone.

## Proposal

**Don't clear the composer on error.** Only clear on `2xx`.

1. On `2xx` - clear input, finalise the optimistic render (today's behaviour).
2. On `4xx` / `5xx` / network error - **restore** the input (don't destroy it), surface the error inline, allow retry.
3. The operator can edit the retained text and retry without retyping.

## Acceptance

- [ ] Submit -> 500/502/503/network error -> composer text **not** cleared
- [ ] Submit -> 400/401/403 -> composer text **not** cleared, error surfaces inline
- [ ] Submit -> 2xx -> composer clears as today
- [ ] Operator can edit retained text and retry without re-typing
- [ ] Test covering all three branches

## Notes

- Composer wiring lives in `web/src/components/AssistantChat/HappyComposer.tsx` and `web/src/hooks/mutations/useSendMessage.ts`.
- A small visual on the composer (red ring + inline error message) is welcome alongside text retention so the failure is obvious.

heavygee:
Closes #776.

## Problem

Submitting a message while the hub is unreachable (a daily-rebuild restart blip, a 5xx, or any 4xx) clears the composer immediately, destroying the typed text. Operators have to retype before retrying.

`@assistant-ui/react`'s composer clears `composer.text` synchronously when send is invoked, so by the time `useSendMessage`'s mutation rejects, the text is already gone. SessionChat additionally clears the pending schedule the moment the mutation is accepted, so a failed scheduled send was also silently downgrading to immediate on the next attempt.

## Fix

- `useSendMessage` exposes a new `onError` option that hands the original input back to the caller alongside the thrown error AND the resolved `scheduledAt` (absolute epoch-ms or null).
- `router.tsx` owns a `sendError` state, set from `onError` and cleared on success or session change.
- `HappyComposer` accepts `sendError` and:
  - Calls `api.composer().setText(sendError.text)` once per failure id, restoring the typed text.
  - When `scheduledAt` is non-null, also calls `onSchedule({ type: 'absolute', ms: scheduledAt })` so the clock button re-activates and the schedule fires at the originally intended instant.
  - Renders a red ring on the composer wrapper and a `role="alert"` inline message; both clear the moment the operator starts editing.

## Acceptance (issue #776)

- [x] Submit -> 500/502/503/network error -> composer text not cleared
- [x] Submit -> 400/401/403 -> composer text not cleared, error surfaces inline
- [x] Submit -> 2xx -> composer clears as today
- [x] Operator can edit retained text and retry without re-typing
- [x] Test covering all three branches
- [x] Failed scheduled sends restore as scheduled (not silently downgraded to immediate)

## Tests

`web/src/hooks/mutations/useSendMessage.test.tsx` - seven new cases under "issue #776":

- 5xx (HTTP 503) - onError fires with the original text.
- Network (TypeError: Failed to fetch) - onError fires with the original text.
- 4xx (HTTP 400) - onError fires with the original text.
- 2xx success - onError does not fire; onSuccess does.
- Non-Error throw (bare string) - onError still fires with the text + null scheduledAt.
- Scheduled send + 5xx - scheduledAt is carried through unchanged so the composer can restore as scheduled.
- Immediate send - scheduledAt is null in onError.

`bun typecheck` and the full `bun --filter web run test` suite (701 tests) both pass on this branch.

## Notes

- No SCHEMA_VERSION bump (frontend-only).
- No `docs/operator/` or `docs/plans/` changes.
- Two commits: initial fix, then a follow-up addressing the bot review on scheduled-send preservation (Major finding).

### DIFF
--- a/web/src/components/AssistantChat/HappyComposer.tsx
+++ b/web/src/components/AssistantChat/HappyComposer.tsx
@@ -40,6 +40,31 @@ export interface TextInputState {
     selection: { start: number; end: number }
 }
 
+/**
+ * One rejected send.  `id` is bumped per failure so two failures with the
+ * same `text` still trigger a fresh restore (the dedupe key is the id, not
+ * the text).
+ *
+ * - `text` is the original input that should be put back into the composer.
+ * - `message` is the user-facing error string we render inline.
+ * - `scheduledAt` is the absolute epoch-ms the rejected send was bound for,
+ *   or null for an immediate send.  When non-null, the composer also
+ *   restores the schedule via `onSchedule` so the operator can edit and
+ *   retry without silently downgrading a scheduled send to immediate.
+ *
+ * Owned by the route component (`router.tsx`); the composer is a pure
+ * consumer that:
+ *  1. restores the text once per `id` via `api.composer().setText`,
+ *  2. restores the schedule (if any) via `onSchedule`, and
+ *  3. shows a red ring + inline message until the user types or sends.
+ */
+export type ComposerSendError = {
+    id: number
+    text: string
+    message: string
+    scheduledAt: number | null
+}
+
 const defaultSuggestionHandler = async (): Promise<Suggestion[]> => []
 
 export function HappyComposer(props: {
@@ -81,6 +106,11 @@ export function HappyComposer(props: {
     pendingSchedule?: PendingSchedule | null
     onSchedule?: (pending: PendingSchedule) => void
     onClearSchedule?: () => void
+    // Set when the most recent send failed (4xx/5xx/network).  The composer
+    // restores the original text once per `sendError.id` and renders an
+    // inline error affordance until the user dismisses or starts editing.
+    sendError?: ComposerSendError | null
+    onClearSendError?: () => void
 }) {
     const { t } = useTranslation()
     const {
@@ -119,7 +149,9 @@ export function HappyComposer(props: {
         onVoiceMicToggle,
         pendingSchedule: pendingScheduleProp,
         onSchedule: onScheduleProp,
-        onClearSchedule: onClearScheduleProp
+        onClearSchedule: onClearScheduleProp,
+        sendError = null,
+        onClearSendError
     } = props
 
     // Use ?? so missing values fall back to default (destructuring defaults only handle undefined)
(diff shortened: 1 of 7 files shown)


heavygee:
Addressed the Major finding in 2b48cf4.

`SendErrorInfo` and `ComposerSendError` now carry `scheduledAt` (resolved to absolute epoch-ms before the failed send). `HappyComposer`'s restore effect feeds it back through `onSchedule` as `{ type: 'absolute', ms: scheduledAt }`, so a failed scheduled send is restored as scheduled - the operator can edit and retry without silently downgrading to immediate. Two new test cases pin the carry-through for both scheduled and immediate sends.

Also corrected the issue ref - this PR closes upstream #776 (the original "closes #15" referenced an unrelated, already-closed upstream issue; my mistake).

heavygee:
Addressed Major #2 in 10dcc32.

`mutation.onError` now calls `removeOptimisticMessage(sessionId, localId)` instead of `updateMessageStatus(_, _, 'failed')`. The optimistic row is dropped from the thread on failure, so the composer-restore + inline alert is the single retry surface - no duplicate failed bubble next to the restored text, no way for the operator to stack a stale failed turn. `retryMessage` stays exposed for any future code path that still produces failed rows.

Test in `useSendMessage.test.tsx` ("removes the optimistic row on failure...") pins the new behaviour: `removeOptimisticMessage` is called, and nothing transitions the row to `'failed'` on this path.

Full web suite green: 702/702. `bun typecheck` clean.

heavygee:
Addressed Major #3 + Major #4 in 29cc498.

**Major #3 (sessionId carry-through):** `SendErrorInfo` now carries the resolved `sessionId`, and `router.tsx` stores `sendErrors` as a `Record<sessionId, ComposerSendError>`. The clear-on-session-change effect is gone - per-session lookup replaces it, so the inactive-session resume race (resolve target id, navigate async, POST fails before nav completes) restores into the resumed-session composer instead of the old one.

**Major #4 (attachment sends):** Forked `mutation.onError` on `input.attachments`:
- Has attachments: keep `updateMessageStatus(_, _, 'failed')` and skip the `options.onError` callback. The composer-restore path can't reinstate uploaded attachment metadata; the legacy failed-bubble UX with the in-thread retry button is the right surface.
- Text-only: `removeOptimisticMessage` + `options.onError({ sessionId, text, scheduledAt, error })` as before.

Also fixed a latent bug the bot's snippet flagged: `retryMessage` was not passing attachments through. Added a `getMessageAttachments` helper that defensively narrows the (`unknown`-typed) message content and pulls the attachments off the stored optimistic; failed-bubble retry of an attachment send now re-fires with its files instead of silently downgrading to text-only.

Three new tests cover the cases:
- `carries sessionId through onError so a resumed-session POST that fails restores into the right composer`
- `attachment send: keeps the failed row in the thread and skips composer-restore`
- `retryMessage: passes attachments through so failed-bubble retry of an attachment send keeps its files`

Full web suite: 705/705. `bun typecheck` clean.

heavygee:
Closing in favour of #798.

While rebuilding scratchlist v1.1 (#798) against fresh upstream/main, the
composer-retain-text fix from this PR was folded into #798 as a single
commit (preserved authorship). The two PRs touched the same composer
surface (HappyComposer.tsx + SessionChat.tsx) and would have stack-conflicted
when shipped separately - cleaner to land them together since the retain-text
bug was effectively discovered while building scratchlist's composer
integration.

Codex review history from this PR (5 iterations, all Major findings addressed)
remains visible here for context; the resolved code is now in #798 at
commit 5aa23ab.

If we want the retain-text fix to land independently of scratchlist v1.1
(faster review path), this PR can be re-opened and #798 can drop the
commit.

PR #774, opened 2026-06-02
feat(codex): support file mentions

NightWatcher314:
## Summary
- add Codex app-server `mention` user input support
- parse `@path` tokens into mention inputs when starting Codex turns
- surface `@` autocomplete suggestions from session file search in the web composer

## Tests
- `bun typecheck`
- `cd cli && bun test src/codex/utils/appServerConfig.test.ts`

### DIFF
--- a/cli/src/codex/appServerTypes.ts
+++ b/cli/src/codex/appServerTypes.ts
@@ -128,6 +128,11 @@ export type UserInput =
         type: 'skill';
         name: string;
         path: string;
+    }
+    | {
+        type: 'mention';
+        name: string;
+        path: string;
     };
 
 export type SandboxPolicy =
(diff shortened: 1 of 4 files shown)


NightWatcher314:
Thanks, fixed in `6c130f4`: autocomplete now quotes paths containing whitespace and the Codex input parser supports quoted/escaped file mentions.

Verified:
- `bun typecheck`
- `cd cli && bun test src/codex/utils/appServerConfig.test.ts`

NightWatcher314:
Thanks, fixed the remaining parser edge case in `728fbe7`: unquoted mentions now keep trailing sentence punctuation outside the file path, while quoted mentions remain unchanged.

Verified:
- `bun typecheck`
- `cd cli && bun test src/codex/utils/appServerConfig.test.ts`

PR #811, opened 2026-06-05
fix(web): drop stale queued-message ghosts on at-bottom refresh

ejjcc:
## Problem

Two old queued-message cards stayed pinned above the composer forever and could never be dismissed. Investigation showed the hub DB had **zero** rows with \`invoked_at IS NULL\` — the cards were pure client-side residue, not real queued work.

### Root cause

A queued user message (\`invokedAt === null\`) is only ever cleared by the live \`messages-consumed\` SSE (which flips \`invokedAt\` via \`markMessagesConsumed\`). That event is **one-shot**. If the client is offline/closed when the CLI consumes the message, the signal is lost permanently. On reload:

1. The row is restored from \`sessionStorage\` still flagged queued (\`invokedAt: null\`).
2. The server's now-invoked copy is too old to appear in the latest fetch window, so \`mergeMessages\` never corrects it.
3. \`trimPreservingQueued\` deliberately never evicts queued rows — so it's pinned forever as a phantom card.

## Fix

Reconcile the rebuilt window against the authoritative latest page in \`fetchLatestMessages\` (at-bottom path). An **immediate, server-echoed, locally-queued** message whose id is absent from the server response is a ghost and is dropped.

False-positive guards (kept even when absent from the response):
- **optimistic rows** (\`id === localId\`) — the server echo may still be in flight; \`mergeMessages\` owns their reconciliation.
- **scheduled rows** (\`scheduledAt != null\`) — the hub omits not-yet-mature scheduled messages from \`getMessages\`, so absence is expected.

A genuinely-still-queued immediate message sorts by \`createdAt\` to the top of the latest page, so it is always present in the fetched window and never dropped.

## Tests

- New integration test reproducing the ghost + its removal on at-bottom refresh.
- New pure-function test for \`reconcileQueuedAgainstLatest\` covering all three keep-guards + the drop case.
- Full web suite: **696 passed**; \`tsc --noEmit\` clean.

### DIFF
--- a/web/src/lib/message-window-store.test.ts
+++ b/web/src/lib/message-window-store.test.ts
@@ -9,6 +9,7 @@ import {
     getMessageWindowState,
     ingestIncomingMessages,
     markMessagesConsumed,
+    reconcileQueuedAgainstLatest,
     removeOptimisticMessage,
     setAtBottom,
     VISIBLE_WINDOW_SIZE,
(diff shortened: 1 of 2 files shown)


ejjcc:
Thanks — confirmed and fixed in 8ca2c11.

The race was real: `reconcileQueuedAgainstLatest` executed inside the updater against the live window, but `response.messages` is the snapshot from the request's start. A `message-received` SSE landing mid-fetch could insert a real server-echoed queued row (`id !== localId`) that the snapshot never saw, and it got filtered as a ghost.

Fix follows the suggested approach: capture the immediate, server-echoed queued ids at request start (`queuedReconcileCandidateIds(initial.messages, initial.pending)`) and gate the helper on them — only rows already queued when the fetch was issued are reconcile candidates; anything that arrives mid-fetch is exempt.

Added a regression test (`keeps a queued row that arrives via SSE while the latest fetch is in flight`) that resolves the HTTP fetch via a deferred and ingests a fresh queued row in the gap, asserting the ghost is dropped while the fresh row survives. Full web suite green (665).

PR #799, opened 2026-06-04
feat(cursor): migrate remote sessions to ACP with model/variant pickers

swear01:
## Summary

Migrates HAPI Cursor **remote** integration from `agent -p --output-format stream-json` to `agent acp`, with legacy stream-json resume only for pre-migration sessions. Adds Zed-style permission/plan/question flows and a two-tier **model catalog** (ACP wire ids + CLI `agent --list-models` SKUs) for New Session and in-session pickers.

Closes #781
Fixes #705
Fixes #687

### Protocol & routing

- **New Cursor sessions → ACP only.** Metadata `cursorSessionProtocol: 'acp'`. No silent fallback to stream-json if ACP init/load fails (fail loud; operator starts fresh).
- **Legacy resume only when metadata says so:** existing `cursorSessionId` without `acp` keeps `cursorLegacyRemoteLauncher` until removed in a follow-up.
- **Live session truth = ACP `configOptions` + hub `session.model`.** Match Zed: prefer `session/set_config_option` (model/mode categories), not ad-hoc CLI flags mid-session.

### Model catalog (two sources, one UI)

| Source | Shape | Role |
|--------|--------|------|
| ACP `session/new` / snapshot | ~28 **wire ids** (`base[param=…]`) | What the agent can actually `set_config_option` |
| `agent --list-models` | ~100+ **CLI slugs** (`gpt-5.5-high-fast`, …) | Human-facing **variants** under each wire base |

- **Merge rule:** group CLI SKUs under `cursorCliSkuBaseId(slug)`; only attach SKUs whose base matches an ACP wire base.
- **Picker display:** when a base has CLI SKUs, **hide raw ACP wire rows** (e.g. `context=272k,reasoning=medium`) and show probe/display names instead.
- **Apply rule:** Web sends wire id or CLI slug → CLI resolves via `matchCliSkuToAcpWireId` before `set_config_option`.
- **No semantic guessing on ACP suffix strings** for labels; scoring uses explicit SKU suffix hints (`-high-fast`, `-low`, …) only for mapping slug → wire.

### Picker UX (Web)

- **Dual columns (Model + Variant)** only when a base has **multiple** variants; otherwise flat list of wires.
- **New Session:** `defaultValue: 'auto'`; **in-session:** `null` (Default = agent default).
- **Multi-variant base click:** update highlighted base **without** applying until user picks a variant (`shouldApply: false`).
[...]

Issue #784, opened 2026-06-03
Safety bug: cursor-agent CLI fabricates `Questions skipped by the user` responses for `AskQuestion` in headless mode - HAPI agents take action on consent the operator never gave


heavygee:
# Safety bug: `cursor-agent` CLI fabricates `Questions skipped by the user` responses for `AskQuestion` in headless mode - HAPI agents take action on consent the operator never gave

## TL;DR

When a HAPI-spawned agent (`cursor-agent --print --output-format stream-json`) calls the `AskQuestion` tool with no Cursor IDE attached, **`cursor-agent` returns the synthetic string `Questions skipped by the user, continue with the information you already have` as the tool result with `is_error=false`** - same wall-clock second as the call, no human in the loop possible. The agent receives this as a legitimate user response, treats it as "operator dismissed the question, proceed", and takes action.

The operator never saw a question form. HAPI's web UI does not render `AskQuestion` (it has a renderer for `AskUserQuestion` only - the claude-code variant). There is no input surface. There is no button to click. **The SDK is fabricating user consent.**

This is a **safety bug**, not a UX paper-cut. In the documented incident below the agent's next three calls happened to be read-only (`git status`, `gh issue list`). The SDK gave the agent permission to do anything - it could have been `gh pr merge`, `git push --force`, file deletes, peer agent spawns. Worse: the bug **reproduces 100% of the time** in headless mode. Every cursor-flavor HAPI agent is one model decision away from acting on fabricated consent.

This issue requests **immediate HAPI-side remediation** (proposals at the bottom) so HAPI is safe regardless of when - or whether - Cursor's CLI team fixes the upstream SDK.

## Environment

- `cursor-agent --version` = `2026.06.02-8c11d9f`
- Invocation by HAPI runner: `cursor-agent -p "<prompt>" --output-format stream-json --resume <sid> --model auto`
- No Cursor IDE attached. Headless. Stream-json over stdout/stdin, no GUI surface.

## Reproduction (100% reliable)

1. Spawn an agent under HAPI, cursor flavor.
2. Get the agent's underlying model to call the `AskQuestion` tool with structured options. (Trivially induced: prompt the agent to ask a multi-choice question; if it doesn't auto-pick the tool, instruct "use AskQuestion tool".)
3. Watch the stream-json:
   - `tool-call` event with the `AskQuestion` payload.
[...]

Issue #781, opened 2026-06-03
Migrate Cursor integration from `agent -p` stream-json mode to Cursor ACP mode


swear01:
### What feature would you like to see?

HAPI currently integrates Cursor Agent CLI remote mode through `agent -p --output-format stream-json --trust`. This works for basic streaming/tool activity, but it misses richer client-side interaction support.

Cursor CLI now supports ACP via `agent acp`, and HAPI should migrate the Cursor integration to ACP mode for a more complete remote-control experience.

Expected Cursor ACP support in HAPI:

- Use `agent acp` instead of `agent -p --output-format stream-json --trust` for remote Cursor sessions.
- Handle ACP `session/request_permission` so HAPI web/mobile can show approve/deny UI.
  - approve once
  - approve for session / allow always
  - reject once
  - cancel/abort
- Preserve Cursor session resume/load behavior through ACP if supported.
- Render ACP `session/update` events:
  - assistant/user message chunks
  - tool calls / tool call updates
  - plan updates
  - mode changes
  - available commands
- Support Cursor ACP extension methods:
  - `cursor/ask_question`
  - `cursor/create_plan`
  - `cursor/update_todos`
  - `cursor/task`
  - `cursor/generate_image`
- Support Cursor modes via ACP:
  - agent/default
  - plan
  - ask
- Support Cursor model selection through ACP where possible:
  - `session/set_model`
  - session `configOptions`
- Keep MCP server wiring compatible with HAPI’s existing happy MCP bridge.
- Prefer reusing HAPI’s existing ACP infrastructure:
  - `AcpSdkBackend`
  - `PermissionAdapter`
  - existing web permission cards
  - existing ACP message conversion path used by Gemini/OpenCode/Kimi

This would make Cursor behave more like the other ACP-backed agents in HAPI and avoid the limitations of current headless `agent -p` mode.

### Additional information

Related Cursor issues already filed in HAPI:

- #687 — Cursor plan mode: plan items not rendered in web UI
- #705 — Cursor model catalog / model picker issue

Those issues are narrower bugs. This issue proposes the broader integration migration: move Cursor remote mode to ACP so permission approval, plan approval, questions, todos, model/mode changes, and richer tool updates can be handled through one protocol.

Current limitation in `docs/guide/cursor.md` says remote Cursor mode uses `agent -p` and `--trust`, so tools run without per-request approval. ACP should let HAPI remove that limitation.

heavygee:
[...]

PR #821, opened 2026-06-06
fix(cursor): trim #784 safety patch to marker-only on legacy stream-json path (closes #822)

### LINKED ISSUE #822: fix(cursor): #801 AskQuestion safety patch misfires in real legacy traffic (post-#799 cleanup)
## Summary

The transitional AskQuestion safety patch shipped in #801 (closing #784) misfires in real legacy traffic. Its marker-string detector never fires (the marker is discarded by the converter before the check runs) and its timing-signature defense-in-depth rewrites legitimate Claude tool calls as `no_input_surface` failures. With the ACP migration (#799) now merged, this code path is **pure legacy** - hit only by resumed pre-ACP sessions - and the safety net should be trimmed to a single, accurate detection rule.

## Scope after #799

After #799 landed, new cursor remote sessions go through `cursorAcpBackend` and the `cursor/ask_question` ACP extension method. They never touch `cursorLegacyEventConverter` and are inherently immune to this fabrication.

The legacy stream-json path (`cursorLegacyRemoteLauncher` → `cursorLegacyEventConverter`) is selected only when:

- A pre-ACP cursor session is resumed (metadata indicates the legacy protocol), or
- ACP is unavailable for some reason and a fallback occurs.

That population drains over time. Everything in this issue is scoped to that draining legacy code path. No new code paths are involved.

## Data: what the #801 intercept actually does in legacy traffic

Hub SQLite snapshot from a single legacy-resumed session (`7b769423`) and the whole driver DB, scanned after #801 merged:

| Metric | Session `7b769423` | All sessions in driver DB |
|---|---:|---:|
| Total messages | 3,229 | 95,524 |
| `tool-call` events | 1,345 | - |
| `name=unknown` tool-call rows | 1,136 | - |
| Rewrites as `no_input_surface` | 16 | 52 |
| Stored tool-result rows containing the literal `Questions skipped by the user...` marker | **0** | **0** |
| Stored rows containing the marker text anywhere | - | 15 (all agent prose: post-mortems, the #784 issue body, #801 source code being read by `read_file`) |

All 16 rewrites in the focal session carried Anthropic Vertex `toolu_vrtx_*` call IDs with `name=unknown` and `result={}`. They were legitimate fast tool calls (TodoWrite, Bash, etc.) that the legacy converter doesn't specifically destructure, mischaracterized as fabricated skip responses.

## Root cause

#801 implemented two detection strategies that both have a defect for the legacy-only code path:

### Strategy 1 (marker-string match) - blind to the shape that actually carries it

The marker scan runs on the output of `extractToolResult`. For shapes the converter doesn't recognize (`readToolCall` / `writeToolCall` / `function`-shaped checks all fall through), `extractToolResult` returns `{}` and the marker is discarded before the check sees it.

The Anthropic Vertex Claude tool calls cursor-agent surfaces in legacy stream-json mode all land in that "unrecognized" bucket. So the only path where a real fabrication could plausibly happen for an unknown-shape AskQuestion has the marker stripped before detection.

### Strategy 2 (timing-signature defense-in-depth) - fires on every fast Claude tool call

The fallback rewrites any AskQuestion-shaped or `name=unknown` tool completion arriving within 500 ms of its start event with a trivial extracted result. Anthropic Vertex Claude tool calls are exactly that population: surfaced as `name=unknown`, extracted result `{}`, and many complete inside the threshold. They are not fabricated skips - they are real, fast tool calls. They should pass through with `status: completed`.

In other words: the false positives are not edge cases. They are the dominant signal the heuristic actually fires on.

## Proposed fix (pure legacy, narrow)

Single detection rule, applied only on the legacy stream-json path:

1. Gate on the AskQuestion-shaped tool names plus the converter-fallback `name=unknown` (#784 issue body documents that cursor's stream-json drops the AskQuestion name in some configurations).
2. Scan the **raw `tool_call`** payload (not `extractToolResult`'s output) for the literal `SYNTHETIC_SKIP_MARKER` string.
3. For function-shaped tools, exclude `function.arguments` from the scan (agent-controlled input - a legitimate AskQuestion debugging this exact bug must not false-positive).
4. For non-function-shaped tools (including the `toolu_vrtx_*` case), scan the whole payload. No agent-input field exists at the top level there.

Remove the timing heuristic entirely. Remove the `toolCallStartedAt` map and its helpers.

## Why marker-only is sufficient now

- New sessions are immune (#799). The legacy converter is no longer the load-bearing surface for AskQuestion - it's the residual surface for already-existing sessions.
- The literal marker text is the only signal we can detect with high confidence in legacy traffic. The timing signature was speculative and turned into a false-positive cannon.
- If cursor-agent ever changes the marker string, the intercept just stops firing - same blast radius as having no intercept, which is exactly the world #799 has put us in for everything except resumed legacy sessions.
- The intercept removes itself naturally as the legacy session population drains. No further upstream work is required to retire it.

## Out of scope

- ACP code, the remote launcher, anything in `cursorAcpBackend`.
- The CursorStreamEvent / parseCursorEvent / convertCursorEventToAgentMessage public surface.
- Retiring `cursorLegacyEventConverter` itself - that's a separate cleanup, blocked on the legacy session population reaching zero.

## Related

- Original bug: #784
- Original transitional patch: #801
- ACP migration that supersedes the legacy path for new sessions: #799
- Proposed fix PR: #821

heavygee:
Closes #822. Follow-up to #801. Originating bug: #784. Code path superseded for new sessions by #799.

## Pure-legacy scope

After #799 (ACP migration) merged, this code path - `cursorLegacyEventConverter` reached only through `cursorLegacyRemoteLauncher` - is **pure legacy**. It is selected only when a pre-ACP cursor session is resumed (legacy stream-json protocol recorded in session metadata). New cursor remote sessions go through `cursorAcpBackend` and the `cursor/ask_question` ACP extension method and are immune to the original #784 fabrication.

That legacy population drains naturally as old sessions age out. This PR is a narrow correctness fix on a draining surface, not new surface area.

## Summary

The #801 intercept misfires in real legacy traffic. Its marker-string detector never fires (the marker is discarded by `extractToolResult` before the check runs for tool shapes the converter doesn't specifically destructure - notably the Anthropic Vertex `toolu_vrtx_*` Claude tool calls cursor-agent surfaces as `name=unknown`). Its timing-signature defense-in-depth (sub-500 ms AskQuestion-shaped completion with trivial result) fires only on false positives - those same Anthropic Vertex Claude tool calls, which are legitimate fast tool calls, not fabricated skips.

This PR replaces both strategies with a single, narrower rule: scan the **raw** `tool_call` payload for the literal `SYNTHETIC_SKIP_MARKER` string, gated on AskQuestion-shaped (and converter-fallback `name=unknown`) names, excluding `function.arguments` for function-shaped tools so a legitimate AskQuestion debugging this exact bug doesn't false-positive. No timing heuristic.

Full diagnosis with data lives in #822.

## Data from one resumed legacy session (`7b769423`)

Hub SQLite snapshot, scanned post-#801:

| Metric | Count |
|---|---:|
| Total messages in session | 3,229 |
| `tool-call` events | 1,345 |
| `name=unknown` tool-call rows | 1,136 |
| Rewrites as `no_input_surface` | 16 |
| Stored rows containing the literal `Questions skipped by the user...` marker | **0** |

All 16 rewrites carried Anthropic Vertex `toolu_vrtx_*` call IDs and `name=unknown` with `result={}` - the converter-fallback shape for Anthropic-format Claude tool calls (TodoWrite, Bash, web_search, etc.) that the legacy converter doesn't specifically destructure. They were legitimate fast tool calls mischaracterized as fabricated skip responses.

Across the whole driver DB (95,524 messages, all sessions): 52 `no_input_surface` rewrites, **zero** rows containing the marker text in tool-result content.

## Changes

- Drop `SYNTHETIC_LATENCY_THRESHOLD_MS`, `toolCallStartedAt` map, `rememberToolCallStart`, `takeToolCallElapsedMs`, `isTrivialResult`, and the test-only `__resetCursorEventConverterStateForTests` export.
- Replace `shouldRewriteAsNoInputSurface(name, result, elapsedMs)` with `shouldRewriteAsNoInputSurface(name, toolCall)` that gates on the AskQuestion name set and delegates to `findMarkerInToolCallResponse(toolCall)`.
- New helper `findMarkerInToolCallResponse`:
  - For function-shaped: scans `function.*` excluding `function.arguments`, plus sibling fields outside `function`.
  - For other shapes: scans the whole `toolCall` payload (no agent-input field exists at the top level for read/write/unknown shapes).
- Refresh tests:
  - Drop timing-based positive cases.
  - Add: marker reachable in a `name=unknown` raw payload where `extractToolResult` would return `{}` (the real-world case the old code missed).
  - Add: legitimate fast `name=unknown` tool call without the marker passes through with `status: 'completed'` (regression against the false positives this PR addresses).
  - Keep the existing false-positive regressions (read_file with marker in content, write_file with marker in args, function tool with marker in result, AskQuestion with marker in arguments).
- Update `docs/guide/cursor.md`:
  - Rename the section to "Legacy stream-json safety: AskQuestion behavior".
  - Lead with "New sessions go through ACP and are immune; this intercept exists only for legacy resumed sessions."
  - Drop the timing-heuristic mention.
  - Drop the "intercept is removed when #781 lands" line (#799 has landed; the intercept now drains naturally with the legacy session population).

## Why marker-only is sufficient now (post-#799)

- The legacy converter is no longer the load-bearing surface for AskQuestion. It's the residual surface for already-existing sessions.
- The literal marker text is the only signal we can detect with high confidence in legacy traffic. The timing signature was speculative and was the dominant false-positive source in real data.
- If cursor-agent ever changes the marker string, the intercept just stops firing - same blast radius as having no intercept, which is exactly the world #799 has put us in for everything except resumed legacy sessions.
- The intercept retires itself naturally as the legacy session population drains. No further upstream work required.

## Test plan

- [x] `bun typecheck` clean (cli + web + hub)
- [x] `bun test src/cursor/utils/cursorLegacyEventConverter.test.ts` → 18/18 pass
- [x] No references to removed exports anywhere in the tree (`__resetCursorEventConverterStateForTests` is no longer imported)
- [x] Verified the file's consumer (`cursorLegacyRemoteLauncher.ts`) still type-checks against the trimmed public surface

## Scope

- Touches only `cli/src/cursor/utils/cursorLegacyEventConverter.{ts,test.ts}` and `docs/guide/cursor.md`.
- Does not touch ACP code, the remote launcher, or any other consumer.
- Does not change the public shape of `CursorStreamEvent`, `parseCursorEvent`, or `convertCursorEventToAgentMessage`.

## Related

- Closes #822
- Original bug: #784
- Original transitional patch: #801
- ACP migration that supersedes the legacy path for new sessions: #799

### DIFF
--- a/cli/src/cursor/utils/cursorLegacyEventConverter.test.ts
+++ b/cli/src/cursor/utils/cursorLegacyEventConverter.test.ts
@@ -1,16 +1,11 @@
-import { describe, it, expect, beforeEach } from 'vitest';
+import { describe, it, expect } from 'vitest';
 import {
     parseCursorEvent,
     convertCursorEventToAgentMessage,
-    __resetCursorEventConverterStateForTests,
     type CursorStreamEvent
 } from './cursorLegacyEventConverter';
 
 describe('cursorLegacyEventConverter', () => {
-    beforeEach(() => {
-        __resetCursorEventConverterStateForTests();
-    });
-
     describe('parseCursorEvent', () => {
         it('parses system init event', () => {
             const line =
@@ -64,7 +59,7 @@ describe('cursorLegacyEventConverter', () => {
             expect(msg).toEqual({ type: 'turn_complete', stopReason: 'success' });
         });
 
-        it('passes through a normal tool result unchanged (read_file)', () => {
+        it('passes through a normal read_file result unchanged', () => {
             const completedEvent = {
                 type: 'tool_call',
                 subtype: 'completed',
@@ -88,16 +83,7 @@ describe('cursorLegacyEventConverter', () => {
     });
 
     describe('#784 transitional safety: AskQuestion synthetic-skip intercept', () => {
-        it('rewrites a tool_call result containing the synthetic skip string to a no_input_surface failure', () => {
-            const startedEvent = {
-                type: 'tool_call',
-                subtype: 'started',
-                call_id: 'q1',
-                session_id: 's1',
-                tool_call: { function: { name: 'AskQuestion', arguments: '{"q":"..."}' } }
-            } as CursorStreamEvent;
-            convertCursorEventToAgentMessage(startedEvent);
-
+        it('rewrites a function-shaped AskQuestion whose result contains the synthetic marker', () => {
             const completedEvent = {
                 type: 'tool_call',
                 subtype: 'completed',
@@ -124,7 +110,7 @@ describe('cursorLegacyEventConverter', () => {
             expect(output.message).toMatch(/Re-prompt in plain text/);
         });
 
-        it('matches the synthetic-skip string even when nested deep inside the tool_call payload', () => {
+        it('matches the synthetic-skip string even when nested deep inside the function payload', () => {
             const completedEvent = {
                 type: 'tool_call',
                 subtype: 'completed',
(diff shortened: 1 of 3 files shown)


heavygee:
Closing to re-route through the fork-side cold-review stage that this repos PR flow requires (see docs/operator/repo-layout-and-dev-flow.md §3 / hapi-pr-create). Will reopen from the same branch and same body once the fork bot has weighed in. Closing as a process correction, not a withdrawal — substance unchanged.

PR #836, opened 2026-06-07
fix(web): hide sidebar fake sessions for Cursor resume/archive

swear01:
## Summary

- Wire `prepareSidebarSessions` into `SessionList`: dedupe by `metadata.agentSessionId`, then hide inactive empty stubs (no agent id, no name/summary)
- Always keep active and currently selected sessions visible
- Expose `lifecycleState` on `SessionSummary` for future sidebar rules
- Use filtered session count in sidebar header

Fixes #833

## Test plan

- [x] `SessionList.test.ts` — dedup, stub filter, selected-stub visibility
- [x] `sessionSummary.test.ts` — lifecycleState mapping
- [x] `bun typecheck`
- [x] `bun run test:web` / `bun run test:shared`
- [x] Live (port 43006): spawn → archive → no ghost sidebar row; resume → one row per thread

PR #844, opened 2026-06-08
feat(cursor): invisible sync-on-open migrator from legacy stream-json to ACP

### LINKED ISSUE #824: feat(cursor): operator-driven legacy stream-json -> ACP session migrator (transplant strategy)
## Summary

#799 migrated NEW Cursor sessions to ACP and left a `cursorLegacyRemoteLauncher` so pre-existing stream-json sessions keep working. It does not provide any migration path - the follow-up was hand-waved as "remove legacy launcher when sessions are migrated or expired". This proposal adds an operator-driven migrator (per-session UI button + bulk CLI command + hub endpoint) that uses a **transplant** strategy: moves the legacy on-disk store into the location `agent acp` already reads from, then flips `cursorSessionProtocol = 'acp'` so the existing ACP launcher picks the session up untouched on next resume.

The strategy avoids replaying the transcript through a fresh ACP session (lossy, token-expensive) because cursor-agent already supports loading the legacy store as-is once the on-disk path is corrected.

## Spike evidence

Two-line summary of what an `agent acp` JSON-RPC probe + hand-driven transplant on two real legacy sessions established (cursor-agent v2026.06.04-8f81907):

1. **`agent acp` `session/load` rejects legacy chat-uuids with a directory-layout error, not a protocol-level error.** The ACP server resolves `session/load` against `~/.cursor/acp-sessions/<uuid>/{store.db, meta.json}`. The legacy stream-json flow stores chats at `~/.cursor/chats/<workspace-hash>/<uuid>/store.db`. The dispatcher in the bundled JS is at `~/.local/share/cursor-agent/versions/2026.06.04-8f81907/8096.index.js` offset ~16685 (`case "session/load": ... e.loadSession(wn.parse(t))`). The rejection an unmigrated uuid returns is:

   ```json
   { "jsonrpc": "2.0", "id": 2,
     "error": { "code": -32602, "message": "Invalid params",
                "data": { "message": "Session \"<uuid>\" not found" } } }
   ```

2. **The on-disk SQLite schema is byte-identical between the two stores.** Both `store.db` files contain:

   ```sql
   CREATE TABLE blobs (id TEXT PRIMARY KEY, data BLOB);
   CREATE TABLE meta  (key TEXT PRIMARY KEY, value TEXT);
   ```

   The `meta` record shape is also functionally identical (`agentId`, `latestRootBlobId`, `name`, `mode`, `createdAt`, optional `lastUsedModel`, optional `isRunEverything`). The only ACP-side addition is a sibling `meta.json` file at the session directory level with `{"schemaVersion":1,"cwd":"<absolute path>","title":"<optional>"}`.

3. **`session/new` does NOT accept any seed-context payload.** Extra fields (`meta`, `instructions`, `transcript`, `history`) on `session/new` are silently dropped by zod's default behavior. There is no path to seed a fresh session with prior content, which rules out the alternative "replay-migration" approach.

4. **Transplant works end-to-end.** Two real legacy sessions (34 messages and 109 messages) transplanted by `cp`ing the legacy `store.db` to a fresh `~/.cursor/acp-sessions/<uuid>/store.db` + synthesizing `meta.json`. In each case:
   - `session/load` completed in ~2.7s.
   - The agent emitted 35 / 127 `session/update` notifications replaying the prior transcript (`user_message_chunk`, `agent_message_chunk`, `agent_thought_chunk`, `tool_call`, `tool_call_update`, `available_commands_update`).
   - A follow-up `session/prompt` asking for the prior conversation's main topic returned a specific, accurate answer drawn from the loaded transcript - i.e. the agent operates on the actual prior state, not a re-derived rendering. Continuity is **semantic**, not cosmetic, and the migration burns **zero preamble tokens**.

## Recipe (per session)

Hub-side migration operates entirely on disk + the local hapi.db. No network. No round-trip through the Cursor IDE.

1. Read the HAPI session row. Guard: `flavor === 'cursor'`, `cursorSessionId` present, `cursorSessionProtocol !== 'acp'`, `lifecycleState !== 'running'`.
2. Locate the on-disk legacy store by scanning `~/.cursor/chats/<wsh>/<cursorSessionId>/store.db`. Refuse if missing (structured telemetry `no_legacy_store_on_disk`).
3. Refuse if `~/.cursor/acp-sessions/<cursorSessionId>/` already exists (collision = invariant violation).
4. `cp` legacy `store.db` → `~/.cursor/acp-sessions/<cursorSessionId>/store.db` (cp, not mv - reversibility).
5. Write `~/.cursor/acp-sessions/<cursorSessionId>/meta.json` with `{"schemaVersion":1,"cwd":"<HAPI metadata.path>","title":"<HAPI metadata.title || legacy meta.name>"}`.
6. **Verify**: spawn `agent acp`, send `initialize` + `session/load {sessionId, cwd, mcpServers:[]}`, then send one minimal `session/prompt`. If either fails, **rollback** by removing the new `~/.cursor/acp-sessions/<cursorSessionId>/` directory and leave the legacy store untouched.
7. If the legacy meta record carries `lastUsedModel`, call `session/set_model` post-load so the operator's model preference survives migration.
8. UPDATE the hapi.db row: `metadata.cursorSessionProtocol = 'acp'`. Existing launcher in #799 (`cursorAcpRemoteLauncher`) picks the session up correctly on next resume; no launcher change needed.
9. Remove the source `store.db` (and the now-empty `<wsh>/<uuid>/` directory). Gated behind a `--keep-source` flag for the paranoid operator. Every removal logs a structured telemetry line for audit.

**Running sessions** are refused by default. A `--force-archive-then-migrate` flag archives the session first, waits for the launcher to release the `store.db` file lock, then proceeds.

**Bulk runs** must first back up the hapi.db (`cp ~/.hapi/hapi.db ~/.hapi/hapi.db.bak.before-migrate-<ts>`).

## Surfaces

- `POST /api/sessions/:id/migrate-to-acp` - per-session HTTP endpoint; body accepts `{keepSource?: boolean, forceArchiveRunning?: boolean, skipVerify?: boolean}`. Returns `{ok, migrated: {acpSessionId, replayNotifications, durationMs}}` or structured error.
- `POST /api/cursor/migrate-legacy` - bulk; body `{lifecycle?: 'archived'|'all', dryRun?: boolean, keepSource?: boolean, forceArchiveRunning?: boolean}`. Returns a per-session result list.
- `hapi cursor migrate <session-id>` - per-session CLI.
- `hapi cursor migrate --all` / `--all-archived` / `--dry-run` / `--keep-source` / `--force-archive-running` - bulk CLI.
- Web: "Migrate to ACP" button on legacy-session rows (gated by `isLegacyCursorSession()` + non-running lifecycle).

## Future work (NOT in this PR)

A separate `hapi cursor import --from-cursor-disk` command could bring chat directories under `~/.cursor/chats/` that HAPI has no metadata row for (e.g. chats created from the Cursor IDE directly, never seen by HAPI's CLI) under HAPI management as fresh sessions. This is a different concern (discovery + metadata synthesis) and deserves its own design.

## Acceptance criteria

- [ ] Unit tests cover the 9-step recipe with mocked filesystem + mocked `agent acp` + mocked hapi.db, including all refusal paths and rollback paths.
- [ ] Integration test (opt-in via env flag) spawns real `agent acp` in an isolated `HOME=` with a synthetic fixture store.db; verifies `session/load` succeeds.
- [ ] No mutation of any session's on-disk store outside the migrator path.
- [ ] `--keep-source` preserves the legacy store; non-flag run removes it AFTER successful verify only.
- [ ] `--force-archive-then-migrate` archives a running session first, waits for lock release, then migrates.
- [ ] `lastUsedModel` from the legacy meta record is re-applied via `session/set_model` post-load.
- [ ] Bulk endpoint/CLI takes a hapi.db backup automatically.
- [ ] No `cp ~/.hapi/hapi.db` writes happen without backup.
- [ ] Web UI button is hidden on already-ACP sessions and on currently-running legacy sessions.

## Why not alternatives

- **Replay-migration** (paste prior transcript as a preamble user message in a fresh ACP session): cosmetic-only continuity (model re-derives state from text), 5-30K input tokens per resume forever, loses tool-call structure. Strictly worse than transplant given §4 evidence.
- **Sunset-only** (UI badge + warning, no migration): doesn't deliver what the request asks for.
- **In-place flip without on-disk move**: blocked - `session/load` literally fails because the file isn't where the ACP server looks.
- **Upstream RFC against cursor-agent to teach `session/load` to fall back to `~/.cursor/chats/`**: unnecessary given §2-§4. The cursor-agent's behavior is already consistent; the directory difference is a layout convention HAPI can adapt to itself.

heavygee:
Closes #824

When the user reopens a legacy stream-json Cursor session in HAPI, the
hub now transparently transplants its `~/.cursor/chats/<wsh>/<uuid>/store.db`
into `~/.cursor/acp-sessions/<uuid>/`, verifies it loads via `agent acp`,
flips `metadata.cursorSessionProtocol = 'acp'`, and removes the legacy
source — all before `resumeSession` returns. Subsequent opens are pure ACP.

## Why migration (not "leave legacy alone")

The primary justification is safety, not feature parity. #784 (`cursor-agent`
fabricates `Questions skipped by the user` responses in legacy stream-json
mode) still fires regularly in dogfood despite #801's mitigation: the agent
ships destructive side effects against fabricated consent. Migration to ACP
closes the protocol-level door because the `AskQuestion` tool does not exist
on the ACP side — there is nothing to fabricate.

#799 deliberately left existing sessions on legacy because they kept
working. That tradeoff was reasonable at the time. The accumulated #784
evidence makes legacy sessions actively unsafe; this PR makes the upgrade
path invisible enough that users stop avoiding it.

## How: transplant → verify → flip → remove

A pre-PR spike established that legacy and ACP `store.db` files use the
identical SQLite schema; only the directory layout differs. The migrator
therefore:

1. Sanity-checks the source store and pre-flips state (`session.active`,
   `lifecycleState`, on-disk presence, target collision)
2. Optionally archives a stale-running row (`forceArchiveRunning: true` is
   the default for the auto-migrate path because the caller already
   verified `session.active === false`)
3. Atomically creates `~/.cursor/acp-sessions/<uuid>/` with mode `0o700`
4. Copies `store.db` and chmods to `0o600` (multi-user-host hardening)
5. Writes a minimal `meta.json` sidecar (`schemaVersion`, `cwd`, optional
   `title`) with mode `0o600`
6. Spawns `agent acp` under HAPI_HOME isolation and verifies the session
   loads via `session/load`. On long histories the verify also drives a
   trivial single-turn prompt; on short ones load-only is enough
7. Flips `cursorSessionProtocol = 'acp'` AND clears the
   `cursorMigrationState` banner flag in a SINGLE metadata write
8. Removes the legacy source store (only after verify succeeded and the
   protocol flip committed). The legacy `~/.cursor/chats` parent dir is
   left as-is

Every failure leaves the legacy state intact. No `rm` fires without a
verify success AND a committed protocol flip.

## In-flight upgrade banner

The transplant takes 15-20s on long histories (copy a multi-hundred-MB
store, spawn `agent acp`, replay thousands of notifications, tear down
the probe). Without a progress indicator the wait reads as "broken" to
a fresh user. A minimal banner ships alongside the migrator:

- Hub sets `metadata.cursorMigrationState = 'in_progress'` **before**
  the long-running transplant. The session-cache refresh emits the
  existing `session-updated` SSE event (no new event type), so the web
  client picks it up in milliseconds. No client-side polling needed.
- Hub clears the flag in the **same** metadata write that flips
  `cursorSessionProtocol` to `'acp'` on success, so the banner disappears
  in the same render tick the chat re-renders as ACP — no flicker window.
- Hub clears the flag explicitly in the auto-migrate helper's `finally`
  on failure/exception, so the banner never gets stuck if migration
  falls back to the legacy launcher.
- Web renders an accessible (`role="status"`, `aria-live="polite"`)
  banner with an indeterminate spinner. Deliberately **no** fake
  percentage — we do not have phase data and a fake progress bar
  would lie.

## Coordination with the ACP refcount lock (#835 / #836 / #837)

This PR is sequenced **after** @swear01's three ACP mop-up PRs (merged
today as `ad038bbf`, `8094b500`, `fa363c2f`), all of which are
prerequisite for safe concurrent ACP launches.

The verify probe spawns `agent acp` directly via `AcpVerifyProbe` under
HAPI_HOME isolation (the migrator overrides `HOME` to a temp dir for
the verify pass), so it never touches `<real-HAPI_HOME>/locks/agent-acp-active/`
at all. Per @swear01's #835 design note, the post-flip ACP launcher claims
the lock through the standard `registerActiveAcpTransport` entry and
behaves like any other concurrent ACP start. The migrator itself never
writes `pid` or `count` files directly.

## Kill-switch

The auto-migrate path is gated by `HAPI_CURSOR_LEGACY_AUTO_MIGRATE`.
Set to `0`, `false`, `no`, or `off` to suppress it entirely (legacy
sessions keep running through the existing stream-json launcher).
Default is on.

## Single-session escape hatch

A REST endpoint at `POST /api/sessions/:id/migrate-to-acp` allows
explicit migration of a single session outside the sync-on-open path
(e.g. for a specific cold archived session a user wants to re-engage).
Bulk migration surfaces (CLI subcommand, web button, bulk REST endpoint,
candidate-listing API) were deliberately stripped per reviewer feedback;
**per-session sync-on-open + this escape hatch are the only two paths**.

## Test budget

- 389 hub unit tests pass against the rebased branch (53 for the
  migrator core, 32 for the verify probe, 15 for the auto-migrate
  helper guard matrix, 4 for the migration banner flag transitions,
  plus all upstream suites)
- 3 integration tests against a real `agent acp` (skipped by default
  unless `HAPI_CURSOR_LEGACY_MIGRATOR_INTEGRATION=1`)
- 10 web unit tests for the banner component (visibility paths + a11y)
- Typecheck clean across `cli` + `web` + `hub`

## Files

- `hub/src/cursor/{acpVerifyProbe,cursorLegacyMigrator}.{ts,test.ts}`:
  migrator core + verify probe + 53+32 unit tests
- `hub/src/cursor/cursorLegacyMigratorIntegration.test.ts`: 3 real-`agent acp` integration tests
- `hub/src/cursor/fixtures/buildSyntheticLegacyStore.ts`: synthetic store.db builder
- `hub/src/sync/syncEngine.ts`: `flipCursorSessionProtocolToAcp`,
  `maybeAutoMigrateLegacyCursorSession`, `setCursorMigrationStateInProgress`,
  `clearCursorMigrationState`, `buildMigratorForRequest`
- `hub/src/sync/syncEngineAutoMigrate.test.ts`: helper guard matrix
- `hub/src/web/routes/{sessions,cli}.ts`: per-session escape-hatch endpoint
- `hub/src/store/index.ts`: `getCursorSessionByCursorIdAndProtocol` lookup
- `shared/src/{schemas,apiTypes}.ts`: `cursorMigrationState` field + request/outcome types
- `web/src/components/CursorMigrationBanner.{tsx,test.tsx}`: banner + 10 tests
- `web/src/components/SessionChat.tsx`: render banner
- `web/src/lib/locales/{en,zh-CN}.ts`: banner copy
- `web/src/types/api.ts`: `Metadata` re-export
- `web/src/api/client.ts`: `migrateCursorSessionToAcp` client method

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1074,7 +1074,7 @@
 
     "@twsxtd/hapi-linux-x64": ["@twsxtd/hapi-linux-x64@0.20.1", "", { "os": "linux", "cpu": "x64", "bin": { "hapi": "bin/hapi" } }, "sha512-VWPCKdAgwfUNBRI9Xy14CKjx1d7JS1irOja5l6zufpaTi139jc51gyDcWFfygMwttQlNimmh2qHTfaFqqvcdNg=="],
 
-    "@twsxtd/hapi-win32-x64": ["@twsxtd/hapi-win32-x64@0.20.0", "", { "os": "win32", "cpu": "x64", "bin": { "hapi": "bin/hapi.exe" } }, "sha512-1GWfncMeaZvBIfSB0RY4UI4ywiKUtOAi41nRHxqUI/VdWS9Rw3syCRa4bH2gFJzrdRtDdi0kfSib9YRHs1uQgg=="],
+    "@twsxtd/hapi-win32-x64": ["@twsxtd/hapi-win32-x64@0.20.1", "", { "os": "win32", "cpu": "x64", "bin": { "hapi": "bin/hapi.exe" } }, "sha512-kHsA3aV9LlIbI0kpqeF8oFeSCLCubaGVHnC3l5AH51KbvlDGeYzXyr1S8KEdgVgd1Gg3cS6LmwYL/xnyr6WO5Q=="],
 
     "@types/aria-query": ["@types/aria-query@5.0.4", "", {}, "sha512-rfT93uj5s0PRL7EzccGMs3brplhcrghnDoV26NqKhCAS1hVo+WdNsPvE/yb6ilfr5hi2MEk6d5EWJTKdxg8jVw=="],
 
--- a/bun.lock
+++ b/bun.lock
@@ -1074,6 +1074,8 @@
 
     "@twsxtd/hapi-linux-x64": ["@twsxtd/hapi-linux-x64@0.20.1", "", { "os": "linux", "cpu": "x64", "bin": { "hapi": "bin/hapi" } }, "sha512-VWPCKdAgwfUNBRI9Xy14CKjx1d7JS1irOja5l6zufpaTi139jc51gyDcWFfygMwttQlNimmh2qHTfaFqqvcdNg=="],
 
+    "@twsxtd/hapi-win32-x64": ["@twsxtd/hapi-win32-x64@0.20.0", "", { "os": "win32", "cpu": "x64", "bin": { "hapi": "bin/hapi.exe" } }, "sha512-1GWfncMeaZvBIfSB0RY4UI4ywiKUtOAi41nRHxqUI/VdWS9Rw3syCRa4bH2gFJzrdRtDdi0kfSib9YRHs1uQgg=="],
+
     "@types/aria-query": ["@types/aria-query@5.0.4", "", {}, "sha512-rfT93uj5s0PRL7EzccGMs3brplhcrghnDoV26NqKhCAS1hVo+WdNsPvE/yb6ilfr5hi2MEk6d5EWJTKdxg8jVw=="],
 
     "@types/babel__core": ["@types/babel__core@7.20.5", "", { "dependencies": { "@babel/parser": "^7.20.7", "@babel/types": "^7.20.7", "@types/babel__generator": "*", "@types/babel__template": "*", "@types/babel__traverse": "*" } }, "sha512-qoQprZvz5wQFJwMDqeseRXWv3rqMvhgpbXFfVyWhbx9X47POIA6i/+dXefEmZKoAgOaTdaIgNSMqMIU61yRyzA=="],
--- a/hub/src/cursor/acpVerifyProbe.test.ts
+++ b/hub/src/cursor/acpVerifyProbe.test.ts
@@ -1,0 +1,358 @@
+/**
+ * Unit tests for the AcpVerifyProbe lock-acquisition primitives.
+ *
+ * The probe spawns a real `agent acp` in production. These tests only
+ * cover the lock dance (start/stop side effects on the agent-acp-active
+ * lock dir), NOT the RPC behaviour — that's covered by the integration
+ * tests with a real agent binary.
+ */
+
+import { describe, expect, it, beforeEach, afterEach } from 'bun:test'
+import { mkdtempSync, rmSync, writeFileSync, existsSync, mkdirSync } from 'node:fs'
+import { join } from 'node:path'
+import { tmpdir } from 'node:os'
+
+import { AcpVerifyProbe, tryAcquireAcpActiveLock } from './acpVerifyProbe'
+
+describe('AcpVerifyProbe — agent-acp-active lock acquisition (Codex #34 P2 v2)', () => {
+    let hapiHome: string
+    beforeEach(() => {
+        hapiHome = mkdtempSync(join(tmpdir(), 'hapi-acp-lock-test-'))
+    })
+    afterEach(() => {
+        try { rmSync(hapiHome, { recursive: true, force: true }) } catch {}
+    })
+
+    function lockDir(home: string): string {
+        return join(home, 'locks', 'agent-acp-active')
+    }
+
+    it('acquires the lock atomically when no holder exists (and releases on stop)', async () => {
+        const probe = new AcpVerifyProbe({
+            agentBinary: '/usr/bin/env', // any executable; we'll stop() before sending RPC
+            hapiHome
+        })
+        // Spawn would normally fail RPC, but the start path itself should
+        // succeed up through agent spawn — what we care about here is the
+        // lock side-effect.
+        probe.start()
+        expect(existsSync(join(lockDir(hapiHome), 'pid'))).toBe(true)
+        await probe.stop()
+        expect(existsSync(lockDir(hapiHome))).toBe(false)
+    })
+
+    it('throws when the lock is held by another live process (atomic refusal)', () => {
+        // Pre-create the lock dir with a pid that IS alive (our own pid).
+        mkdirSync(lockDir(hapiHome), { recursive: true })
+        writeFileSync(join(lockDir(hapiHome), 'pid'), String(process.pid))
+
+        const probe = new AcpVerifyProbe({
+            agentBinary: '/usr/bin/env',
+            hapiHome
+        })
+        expect(() => probe.start()).toThrow(/agent-acp-active lock is held/)
+        // Pre-existing lock dir must NOT be removed by the failed acquire.
+        expect(existsSync(lockDir(hapiHome))).toBe(true)
+    })
+
+    it('clears a stale lock (dead pid file present) and acquires on retry', async () => {
+        // Pre-create with a pid that is virtually certain to be dead.
+        mkdirSync(lockDir(hapiHome), { recursive: true })
+        writeFileSync(join(lockDir(hapiHome), 'pid'), '999999') // typical max_pid; if collides, test is slightly flaky but unlikely on CI
+
+        const probe = new AcpVerifyProbe({
+            agentBinary: '/usr/bin/env',
+            hapiHome
+        })
+        probe.start()
+        // We acquired by clearing the stale dir and re-creating it.
+        expect(existsSync(join(lockDir(hapiHome), 'pid'))).toBe(true)
+        await probe.stop()
+        expect(existsSync(lockDir(hapiHome))).toBe(false)
+    })
+
+    it('refuses on a pidless lock dir (mid-startup race, Codex #34 P2 v3)', () => {
+        // Pre-create an EMPTY lock dir without a pid file — this is what
+        // the CLI guard's registerActiveAcpTransport looks like in the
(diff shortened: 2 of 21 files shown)


heavygee:
Thanks for the review. Both Major findings addressed in `e3e1c0b3`:

**Finding 1 — verify probe `agentLookupHome` ignored `metadata.homeDir`:** Widened `CursorLegacyMigratorDeps.createProbe` signature from `(env) => AcpVerifyProbe` to `(env, agentLookupHome) => AcpVerifyProbe`. The default factory now uses the passed arg; `verifyInTempHome` threads `opts.sourceHome` (already the resolved session-owner home from `metadata.homeDir`) through. The earlier rounds plumbed the type into the factory default but never threaded the resolved home through the call path — exactly the gap you spotted. Two regression tests added in `cursorLegacyMigrator.test.ts`:

- Service-account case (`metadata.homeDir != deps.homeDir()`): captured `agentLookupHome` MUST equal `metadata.homeDir`
- Legacy session record (no `metadata.homeDir`): falls back to `deps.homeDir()` correctly

**Finding 2 — `bun.lock` win32-x64 pin mismatch:** Rebase artifact. The squash-rebase onto upstream/main `1f92a31b` (v0.20.1 release) updated `cli/package.json` to `0.20.1` but the win32-x64 lock entry stayed at `0.20.0`. Regenerated with `bun install`; `bun install --frozen-lockfile` is now clean.

Status:
- 391 hub unit tests pass (+2 new)
- Typecheck clean across cli + web + hub

Ready for re-review.

PR #882, opened 2026-06-12
feat: interactive PTY mode for Claude Code

junmo-kim:
## What

Adds an opt-in **interactive PTY mode** for Claude Code. Instead of spawning the agent headlessly over the SDK, the runner launches `claude` inside a real PTY and the web UI shows **both** a live terminal view and the usual structured chat — driven from the same session.

Enable it per-session with the new **"PTY mode"** checkbox in the new-session form (Claude only; off by default).

## Why

A real PTY runs Claude Code through its interactive entrypoint, so the agent behaves exactly as it does in a local terminal (TUI, slash commands, model switching) while remote users keep the structured chat they already rely on. The live terminal makes long-running or interactive turns observable from the phone.

There's also a usage-accounting reason to prefer the interactive entrypoint. Anthropic now meters interactive Claude Code separately from programmatic Agent-SDK use: subscription usage limits "stay reserved for interactive use of Claude Code," whereas "Claude Agent SDK and `claude -p` usage no longer counts toward your Claude plan's usage limits" and instead draws on a separate Agent-SDK monthly credit or pay-as-you-go API rates ([policy](https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan)). HAPI is a remote control over a Claude Code session a person is actively driving, not a headless automation product — in this mode it is essentially no different from an SSH client: a transport that relays an interactive terminal to and from the user. Launching the genuine interactive entrypoint (PTY) instead of the SDK simply keeps that usage categorized as the interactive session it really is.

## How it works

- The runner launches the real `claude` interactive entrypoint in a PTY and drives it from chat — the first message, in-place `/model` and `/effort` switches, and `--resume`. First-run folder-trust is pre-accepted in a disposable `CLAUDE_CONFIG_DIR`, so the user's own config is never touched.
- The structured chat is reconstructed by tailing Claude's JSONL transcript, so remote users get the same message/tool view as a non-PTY session; the live terminal is streamed alongside it (with scrollback replayed on reconnect).
- **Permissions and questions stay in the chat.** The SDK path routes tool approvals through `canUseTool`, which a PTY agent doesn't have — so a PreToolUse hook forwards each tool call to the runner, which surfaces it in the existing web approval modal and returns the decision to claude. Read-only tools auto-allow; everything else asks the web. The decision is always allow/deny (never `ask`, which would drop back to the TUI) and fails closed on errors, so the flow never silently auto-runs or hangs. `AskUserQuestion` is answered in the chat the same way.

## Screenshots

| Live agent terminal (Claude Code TUI) | Structured chat in a PTY session |
|---|---|
| agent terminal | structured chat |

| PTY mode opt-in | In-session model / permission change |
|---|---|
| new-session option | model change |

| Tool approval in chat (PTY) | AskUserQuestion answered in chat |
|---|---|
| permission modal | question card |

## Testing

- `cli` / `web` (vitest), `hub` / `shared` (`bun test`): full suites green; TypeScript strict clean across all packages.
- New unit coverage for the PTY driver (spawn/ready/submit-retry/trust/idle), the resume verify-resubmit, terminal room routing + scrollback replay, the disposable config dir, and the PreToolUse approval bridge (auto-allow / web round-trip / deny / AskUserQuestion answers / fail-closed).
- Manually verified end-to-end on desktop and mobile.

## Scope

Claude-only and opt-in. The driver is written generically so other PTY-based flavors can follow without reworking it. No schema migrations, and no change to existing (non-PTY) sessions. A second flavor (Antigravity CLI) will follow on the same driver in a separate PR.


junmo-kim:
Closing this for now due to a shift in priorities.

PR #892, opened 2026-06-13
fix(runner): surface dangling-symlink errors instead of misleading EEXIST

### LINKED ISSUE #890: Bug: misleading EEXIST error when resuming a session whose workspace path is a dangling symlink
## Summary

Resuming a session whose `metadata.path` is a symbolic link to a directory that no longer exists fails with a misleading kernel-level error. The user sees:

```
HTTP 500 : {"error":"Unable to create directory at '/home/<user>/coding/<workspace>'. System error: EEXIST: file already exists, mkdir '/home/<user>/coding/<workspace>'. Please verify the path is valid and you have the necessary permissions.","code":"resume_failed"}
```

The actual situation: the symlink exists, its target is gone. The error tells the user nothing useful about why or how to recover.

## Repro

1. Create a session with `metadata.path` pointing at a path that's a symlink to a worktree directory:
   ```
   /home/u/coding/proj-feature -> /home/u/coding/proj/worktrees/feature
   ```
2. Delete the worktree target (`rm -rf /home/u/coding/proj/worktrees/feature` or `git worktree remove ...`).
3. Leave the symlink in place.
4. Try to resume the session through the web UI.

Observed: HTTP 500, `EEXIST` error message, no actionable guidance.

Expected: a clear diagnostic explaining that the workspace path resolves to a deleted target, and what the user can do (recreate the directory at the target, remove the dangling symlink, or archive the session).

## Root cause

`cli/src/runner/run.ts` (lines ~297-339, the `simple` sessionType branch):

```ts
if (sessionType === 'simple') {
  try {
    await fs.access(directory);
    logger.debug(`[RUNNER RUN] Directory exists: ${directory}`);
  } catch (error) {
    // ... approval flow ...
    try {
      await fs.mkdir(directory, { recursive: true });
      directoryCreated = true;
    } catch (mkdirError: any) {
      // ... build error message and return ...
    }
  }
}
```

The bug:

- `fs.access(directory)` follows symlinks (default mode `F_OK`). When the symlink target is missing, it throws `ENOENT`.
- The catch block then calls `fs.mkdir(directory, { recursive: true })`.
- `fs.mkdir` with `recursive: true` only tolerates `EEXIST` when the existing entry is a *directory*. A dangling symlink (or a regular file at the path) is not a directory, so `mkdir` throws `EEXIST`.
- The error reporter renders the kernel error verbatim: `file already exists, mkdir '...'`.

Net result: a path that looks accessible to a human (`ls -la /home/u/coding/proj-feature` shows a symlink) is rejected with an unhelpful EEXIST.

## Proposed fix

In `cli/src/runner/run.ts`, treat the dangling-symlink case as its own diagnostic class. Two small changes:

**A. Detect the dangling-symlink case after `fs.access` fails.**

Use `fs.lstat` to check whether the path is a symlink. If yes, follow it once with `fs.realpath` and check whether the resolved target exists. If the target is missing, return an error with code `resume_failed` and a message that names the dangling symlink, names the missing target, and suggests recovery options:

```
Workspace path '/home/u/coding/proj-feature' is a symlink to '/home/u/coding/proj/worktrees/feature' which no longer exists.
This usually means the worktree was deleted (e.g. ` + '`git worktree remove`' + `) without removing the symlink.
Recovery: recreate the worktree at the target path, remove the dangling symlink, or archive this session.
```

**B. Handle the `EEXIST`-on-non-directory case in the existing mkdir error switch.**

Add a branch that recognises `EEXIST` from the kernel and stat's the path to determine whether it's a regular file, dangling symlink, or directory, and writes a targeted message instead of leaking the raw error.

## Out of scope

- Auto-cleanup of dangling symlinks. The user may have intent here (e.g. they're about to recreate the worktree); HAPI shouldn't unilaterally delete.
- Auto-recreation of the worktree. Same reason.
- Resume of sessions whose `lifecycleState` is `archived` (separate concern; that's a UI/session-management decision).

## Files touched (expected)

- `cli/src/runner/run.ts` (the directory-validation block at 297-339)
- New unit test (likely a co-located helper + `*.test.ts` next to it; only `runner.integration.test.ts` exists today)

## Definition of done

- A new unit test that:
  - Creates a tmp directory containing a symlink whose target does not exist.
  - Calls the directory-validation logic with that path.
  - Asserts the error result names a dangling symlink (or equivalent) and points at both the symlink path and the missing target.
- Manual repro: a session with a dangling-symlink workspace path produces a useful error in the web UI's "Resume failed" toast.
- The existing kernel-EEXIST path is preserved for true file-collision cases (regular file at the directory path) with a similarly clear message.

## Reporter environment

Found while resurrecting an archived session whose worktree had been pruned days earlier. The dangling symlink had survived the worktree's removal.

heavygee:
Closes #890.

## Summary

When a session's workspace path is a symbolic link to a directory that no longer exists, the runner currently fails with a misleading kernel-level error:

```
Unable to create directory at '/path'. System error: EEXIST: file already exists, mkdir '/path'.
```

The user has no way to tell that the dangling symlink is the actual problem.

## Root cause

In `cli/src/runner/run.ts` (the `simple` sessionType branch around lines 297-339):

- `fs.access(directory)` follows symlinks. When the target is missing, it throws `ENOENT`.
- The catch block falls through to `fs.mkdir(directory, { recursive: true })`.
- `mkdir` with `recursive: true` only tolerates `EEXIST` against an existing **directory**. A dangling symlink is not a directory, so `mkdir` throws `EEXIST`.
- The error reporter renders the kernel error verbatim.

## Fix

Replace the inline `fs.access` + `fs.mkdir` pair with a small `validateWorkspaceDirectory` helper that uses `fs.lstat` to inspect symlinks without following them. The helper distinguishes:

- missing path -> approval flow / mkdir as before
- existing directory -> ok
- regular file at the workspace path -> "non-directory file" error
- symlink to an existing directory -> ok
- symlink to a non-directory -> "not a directory" error
- **dangling symlink** -> diagnostic naming both the symlink path and the missing target, with recovery options (recreate the target, remove the symlink, archive the session)

The existing `mkdir` errno switch (EACCES, ENOTDIR, ENOSPC, EROFS) is preserved and extended with an EEXIST race-recheck so the kernel error code never leaks.

## Tests

`cli/src/runner/validateWorkspaceDirectory.test.ts` covers:

- 7 fs-touching cases against a real tmpdir + symlinks (missing path, approval gate, existing directory, regular file, symlink-to-dir, dangling symlink, symlink-to-file).
- 6 cases for the pure errno-to-message mapper (EACCES, ENOTDIR, ENOSPC, EROFS, EEXIST, unknown code).

Locally: ` + '`bun typecheck`' + ` clean across cli/web/hub; ` + '`bun run test`' + ` clean (977 cli + 416 hub tests pass; the only failure is the pre-existing flaky ` + '`runner.integration.test.ts > should detect version mismatch and kill old runner`' + ` which fails on ` + '`upstream/main`' + ` without my changes too).

## Manual repro

Against the actual broken path that triggered this report:

```
/home/u/coding/hapi-scratchlist-per-session -> /home/u/coding/hapi/worktrees/scratchlist-per-session
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^ deleted
```

Before: ` + '`HTTP 500 EEXIST: file already exists, mkdir ...`' + `

After:

```
Workspace path '/home/u/coding/hapi-scratchlist-per-session' is a symbolic link to
'/home/u/coding/hapi/worktrees/scratchlist-per-session', which no longer exists. This
usually means the target was deleted (e.g. via ` + '`git worktree remove`' + `) without removing
the symlink. Recovery: recreate the directory at the target path, remove the dangling
symlink (` + '`rm \'/home/u/coding/hapi-scratchlist-per-session\'`' + `), or archive this session.
```

## Out of scope

- Auto-cleanup of dangling symlinks (the user may have intent).
- Auto-recreation of the worktree.
- Behavior changes for archived sessions (separate concern).
- The ` + '`worktree`' + ` sessionType branch is intentionally untouched - its existing "Worktree sessions require an existing Git repository" error message is already a useful diagnostic for missing base paths.

### DIFF
--- a/cli/src/runner/run.ts
+++ b/cli/src/runner/run.ts
@@ -21,6 +21,7 @@ import { isRetryableConnectionError } from '@/utils/errorUtils';
 import { cleanupRunnerState, getInstalledCliMtimeMs, isRunnerRunningCurrentlyInstalledHappyVersion, stopRunner, waitForRunnerHandoff } from './controlClient';
 import { startRunnerControlServer } from './controlServer';
 import { createWorktree, removeWorktree, type WorktreeInfo } from './worktree';
+import { validateWorkspaceDirectory } from './validateWorkspaceDirectory';
 import { join } from 'path';
 import { buildMachineMetadata } from '@/agent/sessionFactory';
 import { resolveWorkspaceRoots } from '@/utils/workspaceRoot';
@@ -295,47 +296,28 @@ export async function startRunner(options: { workspaceRoots?: string[] } = {}): 
       let happyProcess: ReturnType<typeof spawnHappyCLI> | null = null;
 
       if (sessionType === 'simple') {
-        try {
-          await fs.access(directory);
+        const validation = await validateWorkspaceDirectory(directory, {
+          approvedNewDirectoryCreation
+        });
+        if (validation.type === 'requestApproval') {
+          logger.debug(`[RUNNER RUN] Directory creation not approved for: ${directory}`);
+          return {
+            type: 'requestToApproveDirectoryCreation',
+            directory
+          };
+        }
+        if (validation.type === 'error') {
+          logger.debug(`[RUNNER RUN] Workspace directory validation failed: ${validation.errorMessage}`);
+          return {
+            type: 'error',
+            errorMessage: validation.errorMessage
+          };
+        }
+        directoryCreated = validation.created;
+        if (validation.created) {
+          logger.debug(`[RUNNER RUN] Successfully created directory: ${directory}`);
+        } else {
           logger.debug(`[RUNNER RUN] Directory exists: ${directory}`);
-        } catch (error) {
-          logger.debug(`[RUNNER RUN] Directory doesn't exist, creating: ${directory}`);
-
-          // Check if directory creation is approved
-          if (!approvedNewDirectoryCreation) {
-            logger.debug(`[RUNNER RUN] Directory creation not approved for: ${directory}`);
-            return {
-              type: 'requestToApproveDirectoryCreation',
-              directory
-            };
-          }
-
-          try {
-            await fs.mkdir(directory, { recursive: true });
-            logger.debug(`[RUNNER RUN] Successfully created directory: ${directory}`);
-            directoryCreated = true;
-          } catch (mkdirError: any) {
-            let errorMessage = `Unable to create directory at '${directory}'. `;
-
-            // Provide more helpful error messages based on the error code
-            if (mkdirError.code === 'EACCES') {
-              errorMessage += `Permission denied. You don't have write access to create a folder at this location. Try using a different path or check your permissions.`;
-            } else if (mkdirError.code === 'ENOTDIR') {
-              errorMessage += `A file already exists at this path or in the parent path. Cannot create a directory here. Please choose a different location.`;
-            } else if (mkdirError.code === 'ENOSPC') {
-              errorMessage += `No space left on device. Your disk is full. Please free up some space and try again.`;
-            } else if (mkdirError.code === 'EROFS') {
-              errorMessage += `The file system is read-only. Cannot create directories here. Please choose a writable location.`;
-            } else {
-              errorMessage += `System error: ${mkdirError.message || mkdirError}. Please verify the path is valid and you have the necessary permissions.`;
-            }
-
-            logger.debug(`[RUNNER RUN] Directory creation failed: ${errorMessage}`);
-            return {
-              type: 'error',
-              errorMessage
-            };
-          }
         }
       } else {
         try {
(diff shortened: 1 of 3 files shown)


heavygee:
Addressed in 073add02.

The Major finding (`rm '${directory}'` was a copy-pasteable shell command that broke if the path contained a single quote) is fixed by dropping the literal command shape and describing the recovery action in prose. The path is still mentioned for clarity but no longer presented as something to paste into a shell.

Added a regression test (`does not produce a copy-pasteable rm command when the path contains a single quote`) that mkdir's a directory whose name contains `'`, creates a dangling symlink under it, and asserts the resulting error message never contains the literal `\`rm ...\`` shape.

Local gates: `bun typecheck` clean across cli/web/hub; `bun run test` clean (978 cli + 416 hub pass; the only failure is the pre-existing flaky `runner.integration.test.ts > should detect version mismatch` which fails on `upstream/main` without my changes too).

PR #900, opened 2026-06-13
feat: silent in-place patch loop for mermaid render failures + GFM table repair

### LINKED ISSUE #850: feat: in-place silent correction loop for agent render failures (mermaid, tables)
## Problem

Agents (Claude, Cursor, Codex) frequently produce structured content that the web renderer fails to parse correctly:

- **Mermaid diagrams** — typo in diagram type keyword, invalid syntax — web falls back to raw `<pre><code>` dump
- **Markdown tables** — missing header separator row, inconsistent column counts, wrong pipe alignment — GFM parser falls back to a paragraph or produces garbled HTML

In both cases: the user sees broken output, the agent has no idea, and the conversation continues. If feedback eventually reaches the agent, the correction appears as a separate out-of-context reply turn — clunky and jarring.

PR #845 attempted a hub-side mermaid type-check approach but was closed: the hub broadcasts the original message before it can intercept it, so the user always sees the broken block first. It also impersonated the user via a `role:'user'` system message.

## Proposed solution: in-place silent patch loop

The web client is the right place for detection — it's where rendering actually happens. On failure, show a placeholder and fire a silent targeted correction prompt to the agent. The corrected content splices in-place. The user never sees the broken output as a permanent state; no correction turn appears in the conversation.

### Flow

```mermaid
sequenceDiagram
    participant Agent
    participant CLI
    participant Hub
    participant Web

    Agent->>CLI: streams response (broken mermaid or table)
    CLI->>Hub: socket 'message' — stored, broadcast
    Hub->>Web: new-message (broken content delivered)
    Web->>Web: render failure detected
    Web->>Web: show placeholder "⟳ rendering…" over broken block
    Web->>Hub: socket 'patch-request' {msgId, blockIndex, type, failedCode}
    Hub->>Hub: dedup (only first connected client fires patch)
    Hub->>CLI: socket 'patch-prompt' {msgId, blockIndex, type, failedCode}
    CLI->>Agent: silent system prompt — "return only the corrected block"
    Agent->>CLI: corrected block only (no prose)
    CLI->>Hub: socket 'patch-response' {msgId, blockIndex, correctedCode}
    Hub->>Hub: store patch delta (original message not mutated)
    Hub->>Web: socket 'message-patched' {msgId, blockIndex, correctedCode}
    Web->>Web: splice corrected block into original response
    Web->>Web: renders correctly — placeholder replaced
```

### Detection per content type

| Type | Where detected | Failure signal |
|---|---|---|
| Mermaid diagram | `mermaid-diagram.tsx` | `mermaid.parse()` returns false / throws |
| Markdown table | markdown AST (remark/rehype pipeline) | Node parsed as `paragraph` when lines contain `\|` + intended separator row; or `table` node with inconsistent column counts |

The patch-prompt loop is **identical** for both — only detection differs.

### Common table failure modes

- Missing header separator row (`| --- | --- |`)
- Inconsistent column counts across rows
- Missing leading/trailing `|` on rows
- Escaped pipes (`\|`) breaking column alignment
- Prose paragraph where a table was clearly intended (heuristic: 3+ consecutive lines with `|`)

### Key properties

- **User never sees broken output as permanent** — placeholder during correction, not a code dump or garbled table
- **No impersonation** — `patch-prompt` is a new first-class protocol message type, not a `role:'user'` fake turn
- **Invisible to conversation history** — the correction loop doesn't appear as a user or assistant turn
- **In-place** — corrected content appears where the original was, in full context
- **Deduped at hub** — multiple connected clients don't each fire a correction
- **Extensible** — any future render-failure type (code block with unsupported language, LaTeX parse error, etc.) plugs into the same loop

### New pieces required

| Piece | Where | What |
|---|---|---|
| Failure placeholder | `web/src/components/assistant-ui/mermaid-diagram.tsx` + table renderer | Show "⟳ rendering…" instead of broken output |
| Table failure detection | Markdown AST / table renderer component | Detect malformed GFM tables before/after parse |
| `patch-request` event | Web → Hub socket | `{ msgId, blockIndex, type, failedCode }` |
| Dedup + routing | `hub/src/socket/handlers/` | Only first client fires; route to active CLI socket |
| `patch-prompt` event | Hub → CLI socket | New event type; CLI injects as silent system context |
| Silent system injection | `cli/src/api/apiSession.ts` | Narrow prompt: "return only the corrected [mermaid/table] block" |
| `patch-response` event | CLI → Hub socket | `{ msgId, blockIndex, correctedCode }` |
| Patch storage | Hub message store | Delta alongside original (audit trail preserved) |
| `message-patched` event | Hub → Web socket | Broadcast to all clients for this session |
| In-place splice | Web renderer components | Replace placeholder with corrected rendered output |

### What this is NOT

- Not unlimited retries — cap at 2 attempts per block, then show a permanent error state
- Not a mutation of stored messages — original preserved, patch is a delta
- Not a new conversation turn — `patch-prompt`/`patch-response` are system-level events

### Relationship to #829

This supersedes the hub-side approach in PR #845 (closed). That PR solved the wrong problem (notifying the agent after the user saw breakage). This solves the right problem: user never sees breakage; correction is silent and in-place.

heavygee:
## Summary

Closes #850.

Two related improvements for rendering resilience in the web client.

### Commit 1 — silent in-place patch loop for mermaid render failures

When `mermaid.render()` throws a parse error the diagram slot previously showed a permanent error state with no recovery path. This PR wires up a silent background correction loop:

1. `MermaidDiagram` detects the failure, claims a `blockIndex` for the block within its message, transitions to a `'patching'` spinner state, and fires `POST /sessions/:id/patch-request` with `{ msgId, blockIndex, type: 'mermaid', failedCode }`.
2. Hub (`SyncEngine.requestPatch`) deduplicates by `sessionId:msgId:blockIndex`, caps retries at 2, and auto-evicts stale entries after 30 s. It emits a `patch-prompt` socket event to the active CLI.
3. CLI (`registerPatchHandler`) runs a silent single-turn `query()` call (`maxTurns: 1`, `permissionMode: 'default'`, `canCallTool` hard-denies all tools) — no user turn is created or injected into conversation history.
4. Hub broadcasts `message-patched` SSE → `MermaidDiagram` splices the corrected diagram code in-place and re-renders.
5. Web component falls back to error state after 15 s if no `message-patched` arrives. Patch loop is gated on `flavor === 'claude'` (other flavors have no registered CLI handler) and skips blocks > 16 k chars.

Hub endpoints: `POST /sessions/:id/patch-request` (validated with `PatchRequestBodySchema`, `failedCode` capped at 16 k chars) returns 200 (sent/duplicate) or 503 (no CLI connected).

### Commit 2 — mechanical repair of GFM tables with off-by-one separator rows

`remark-gfm` follows the GFM spec: the delimiter row controls the column count. When an agent emits a table whose separator has fewer cells than the header, `remark-gfm` silently discards the extra columns. Analysis of real session data showed this is the dominant table failure pattern (~15 % of tables, almost all off-by-one or off-by-two in the separator).

A new remark plugin (`remarkRepairTables`) runs immediately after `remarkGfm` in `MemoizedMarkdown`. It uses `file.value` source + `node.position` to read the original separator line, counts header columns vs `node.align.length`, pads the separator with `---` cells, and re-parses just that table block. Only broken tables are touched; valid tables, prose with pipes, and code spans are unaffected.

9 vitest unit tests cover: valid table pass-through, off-by-one, off-by-N, no surrounding pipes, alignment hints preserved, multiple mixed tables, and prose/code-span false-positive immunity.

## Test plan

- [x] `bun typecheck` — clean (cli / hub / web)
- [x] `bun run test:web` — 9 new table-repair tests pass, no regressions
- [x] Cold self-review of the diff before submitting

## Disclosure

This PR was developed with assistance from Claude Sonnet 4.6 (claude-sonnet-4-6) acting as a peer agent in a multi-agent workflow.

### DIFF
--- a/cli/src/api/apiSession.ts
+++ b/cli/src/api/apiSession.ts
@@ -163,6 +163,7 @@ export class ApiSessionClient extends EventEmitter {
     private pendingMessages: { message: UserMessage; localId?: string }[] = []
     private pendingMessageCallback: ((message: UserMessage, localId?: string) => void) | null = null
     private cancelQueuedMessageCallback: ((localId: string) => boolean) | null = null
+    private patchPromptCallback: ((payload: { msgId: string; blockIndex: number; type: 'mermaid' | 'table'; failedCode: string }) => void) | null = null
     private readonly incomingFilter = new IncomingMessageFilter()
     private backfillInFlight: Promise<void> | null = null
     private needsBackfill = false
@@ -332,7 +333,25 @@ export class ApiSessionClient extends EventEmitter {
             }
         })
 
+        this.socket.on('patch-prompt', (data: { msgId: string; blockIndex: number; type: 'mermaid' | 'table'; failedCode: string }) => {
+            if (!data || typeof data.msgId !== 'string' || typeof data.blockIndex !== 'number') {
+                return
+            }
+            this.patchPromptCallback?.(data)
+        })
+
         this.socket.connect()
+    }
+
+    onPatchPrompt(callback: (payload: { msgId: string; blockIndex: number; type: 'mermaid' | 'table'; failedCode: string }) => void): void {
+        this.patchPromptCallback = callback
+    }
+
+    sendPatchResponse(payload: { msgId: string; blockIndex: number; correctedCode: string }): void {
+        this.socket.emit('patch-response', {
+            sid: this.sessionId,
+            ...payload
+        })
     }
 
     onUserMessage(callback: (data: UserMessage, localId?: string) => void): void {
--- a/cli/src/claude/runClaude.ts
+++ b/cli/src/claude/runClaude.ts
@@ -22,6 +22,7 @@ import { formatMessageWithAttachments } from '@/utils/attachmentFormatter';
 import { normalizeClaudeSessionModel } from './model';
 import { normalizeClaudeSessionEffort } from './effort';
 import { getInvokedCwd } from '@/utils/invokedCwd';
+import { registerPatchHandler } from '@/claude/utils/patchHandler';
 
 export interface StartOptions {
     model?: string
@@ -353,6 +354,14 @@ export async function runClaude(options: StartOptions = {}): Promise<void> {
         const removed = messageQueue.cancelByLocalId(localId);
         logger.debug(`[claude] cancelByLocalId(${localId}): ${removed ? 'removed' : 'not found (best-effort)'}`);
         return removed;
+    });
+
+    // Register silent in-place patch handler for mermaid/table render failures.
+    registerPatchHandler(session, {
+        cwd: workingDirectory,
+        model: currentModel ?? null,
+        claudeEnvVars: options.claudeEnvVars,
+        claudeArgs: options.claudeArgs
     });
 
     const resolvePermissionMode = (value: unknown): PermissionMode => {
--- a/cli/src/claude/utils/patchHandler.ts
+++ b/cli/src/claude/utils/patchHandler.ts
@@ -9,8 +9,13 @@
  */
 
 import { query } from '@/claude/sdk/query'
+import type { SDKUserMessage } from '@/claude/sdk'
 import { logger } from '@/ui/logger'
 import type { ApiSessionClient } from '@/api/apiSession'
+
+async function* singleTurnPrompt(text: string): AsyncGenerator<SDKUserMessage> {
+    yield { type: 'user', message: { role: 'user', content: text } } as SDKUserMessage
+}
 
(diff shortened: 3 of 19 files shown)


heavygee:
Closing this for a redesign. The patch loop as written gates on `flavor === 'claude'`, but mermaid render failures occur primarily in non-Claude agent sessions. Claude sessions rarely produce broken diagrams, so the feature as implemented solves a problem that doesn't arise in the sessions it covers. Will revisit when there's a cross-flavor one-shot repair path available.

PR #925, opened 2026-06-16
feat(web): add download button to file viewer

### LINKED ISSUE #924: feat(web): add download button to file viewer
## Summary

The web file browser currently has no way to download a file to the user's device. When viewing a file via the **Directories** tab → file viewer (`/sessions/:id/file`), the only available actions are:

- Copy path
- Copy content (text files only)

There is no download button, making it impossible to save binary files (images, compiled assets, etc.) or large text files that exceed the copy size limit.

## Proposed Solution

Add a download button in the file viewer header toolbar. Since the existing `GET /api/sessions/:id/file` endpoint already returns the full file content as base64, the download can be implemented entirely on the client side:

1. Decode the base64 response into a `Blob`
2. Create a temporary object URL and trigger `<a download>` click
3. Revoke the object URL

No new backend endpoint is required.

## Expected Behavior

- A download icon button appears next to the copy-path button in the file viewer toolbar
- Clicking it downloads the file with the correct filename
- Works for both text and binary files (images, etc.)
- Button is disabled / hidden while the file is still loading

## Current Behavior

No download option exists anywhere in the file browser UI.

swear01:
## Summary

Closes #924

- Adds a **download icon button** to the file viewer toolbar, placed next to the existing copy-path button
- Clicking it decodes the existing base64 file content (`GET /api/sessions/:id/file`) into a `Blob` and triggers a browser `<a download>` — no new backend endpoint required
- Works for text files, binary files, and images
- Button is hidden until the file has loaded successfully (`fileContentResult.success === true`)

## Test plan

- [ ] Open the file browser → Directories tab → click any text file → download button appears in toolbar
- [ ] Download triggers with correct filename
- [ ] Works for image files (PNG, JPG, etc.)
- [ ] Works for binary files
- [ ] Button is absent while file is loading or on error

🤖 Generated with [Claude Code](https://claude.com/claude-code) via [HAPI](https://hapi.run)

### DIFF
--- a/web/src/components/SessionFiles/DirectoryTree.tsx
+++ b/web/src/components/SessionFiles/DirectoryTree.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import type { ApiClient } from '@/api/client'
 import { FileIcon } from '@/components/FileIcon'
 import { useSessionDirectory } from '@/hooks/queries/useSessionDirectory'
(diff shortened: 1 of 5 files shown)


swear01:
Closing in favor of a clean branch rebased on main.

PR #978, opened 2026-06-29
[codex] document Xiaomi microphone troubleshooting

quecai-niu:
## Summary

- document a Xiaomi/MIUI-specific microphone permission troubleshooting step
- call out overlay permissions from Xiaomi Wallet and similar apps
- explain how to retry microphone authorization after disabling interfering overlays

## Why

On affected Xiaomi/MIUI devices, browser microphone authorization can be disrupted by apps allowed to display over other apps. This is an environment-specific permission interaction rather than a HAPI microphone implementation defect.

## Impact

Users get a direct recovery path without adding device-specific permission workarounds to the HAPI codebase.

## Validation

- documentation diff reviewed
- `git diff --check` passed

### DIFF
--- a/docs/guide/voice-assistant.md
+++ b/docs/guide/voice-assistant.md
@@ -109,9 +109,9 @@ Set `ELEVENLABS_API_KEY` in your environment and restart the hub.
 - Ensure no other app is using the microphone
 - Try refreshing the page
 
-### 小米/MIUI 手机上麦克风授权失败
+### Microphone permission fails on Xiaomi/MIUI devices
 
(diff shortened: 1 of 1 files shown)


tiann:
Which browser do you use? We only test on Chrome.

quecai-niu:
> Which browser do you use? We only test on Chrome.

I tested this on a Xiaomi Android phone using both Google Chrome and Microsoft Edge, and the issue occurred in both browsers.  
I found that the voice feature stopped working because the bookkeeping feature in Xiaomi Wallet was enabled. I submitted this documentation update because other users may encounter the same issue without knowing how to resolve it.

heavygee:
Merging as estate **lane B** (auto: docs-only / non-product) under @tiann's low-impact blessing ([#1268 comment](https://github.com/tiann/hapi/pull/1268#issuecomment-5141575753)) — taking a tiny docs PR off the maintainer plate. Thanks @quecai-niu!

PR #985, opened 2026-07-01
feat(web): global word-wrap toggle for code and diff views

### LINKED ISSUE #905: Mobile Web Terminal Output could wrap

When we used the hapi in mobile web. I view the terminal output and anything tools execution result. 
The output content could be long. Could we add wrap option/function to handle it? so that we could 
view them at full picture no need to scroll vertically.

### Additional information

_No response_

junmo-kim:
Fixes #905.

## Problem

The web UI renders code blocks, unified diffs, and long tool output/input with horizontal scrolling only (`overflow-x-auto`). On narrow screens — Android PWA, phones — long lines run off the right edge and you have to scroll sideways to read them, which is awkward mid-conversation.

## Solution

Add a **word-wrap toggle** next to the copy button in the code block header. The preference is stored globally in `localStorage` (`hapi-code-wrap`, default **off**), so one toggle applies to every code block, diff, and tool input/output at once, and it survives reloads.

To keep line numbers aligned when a line wraps, code blocks now render **one grid row per logical line** — a line-number cell and a code cell as siblings in the same row. When the code cell wraps, the row grows taller and the number stays pinned to the first visual line, matching the GitHub / VS Code gutter behavior. This replaces the previous two independent `<pre>` columns, whose separate text flows drifted out of alignment when wrapped. Rendering falls back to plain-text line splitting (with the same normalization) while highlighting is pending or the language is unsupported, so line numbers always match.

This PR also adds a lightly shaded **gutter background** behind the line numbers, consistent in light and dark themes, so the line-number column reads as a distinct gutter.

Default is **off**, so existing horizontal-scroll rendering is unchanged unless the user turns wrap on.

## Usage

Click the wrap icon in any code block header. The setting is remembered across the session and all code/diff/tool views.

|  | Wrap off (default) | Wrap on |
|---|---|---|
| **Light** |  |  |
| **Dark** |  |  |

## Tests

- Unit tests (vitest + Testing Library): `useCodeWrap` persistence + same-tab sync; per-line rendering (line count, wrap on/off, plain-text fallback); wrap-toggle suppression when a code block is nested inside an interactive ancestor (avoids nested `<button>`); `<pre>` preformatted semantics.
- Manually verified end-to-end on desktop and mobile (390px) viewports, in light and dark themes.

## Reviewing

Commits are split for review:
1. `refactor(web): add per-line shiki line-splitting helper`
2. `feat(web): add global word-wrap toggle for code, markdown, and diff views`
3. `feat(web): add a shaded gutter background behind line numbers`

If the gutter background (commit 3) is out of scope for you, it drops cleanly on its own without touching the wrap feature.

### DIFF
--- a/web/src/components/CliOutputBlock.test.tsx
+++ b/web/src/components/CliOutputBlock.test.tsx
@@ -14,4 +14,20 @@ describe('CliOutputBlock', () => {
         expect(screen.getByRole('button', { name: /npm test/i })).toBeInTheDocument()
         expect(screen.queryByTitle('Copy')).not.toBeInTheDocument()
     })
+
+    it('does not render a nested wrap toggle button inside the preview trigger', () => {
+        // The preview CodeBlock renders inside a DialogTrigger <button>, so
+        // its wrap toggle <button> must be suppressed to avoid nesting
+        // interactive elements (invalid HTML / hydration violation).
+        render(
+            <I18nProvider>
+                <CliOutputBlock text={'<command-name>npm test</command-name><local-command-stdout>ok</local-command-stdout>'} />
+            </I18nProvider>
+        )
+
+        // The wrap toggle is the only button carrying aria-pressed; asserting
+        // its absence by role avoids coupling to the localized title.
+        expect(screen.queryByRole('button', { pressed: false })).not.toBeInTheDocument()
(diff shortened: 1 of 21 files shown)


junmo-kim:
@tiann This PR addresses the confirmed wrap-support issue. Could you take a look when you have a chance?

PR #988, opened 2026-07-01
fix(web): keep abort next to Send by default

junmo-kim:
## Summary

The composer toolbar is customizable, but defaults still matter. Asking each user to manually reposition Abort for a predictable layout shifts a basic UX decision onto them.

Abort is a transient control used while a turn is running, so the default now keeps it at the end of the toolbar, immediately next to Send. Saved custom orders remain unchanged.

## Change

- derive the default toolbar order from the registered items with Abort appended last
- leave item registration and missing-item normalization order unchanged
- add a regression test for the exact default order

## Before / After

The red callout is added only to make the Abort position easy to compare.

| Before — mid-toolbar | After — next to Send |
|---|---|
| Before: Abort in the middle of the default composer toolbar | After: Abort next to Send in the default composer toolbar |

## Test plan

- [x] `bun typecheck`
- [x] `bun run test`
- [x] Isolated Hub + Vite + Playwright render of the default Chat settings toolbar, with saved toolbar state cleared


junmo-kim:
Updated this to work with the recently merged customizable toolbar. Abort ended up in the middle of the default toolbar when I added Scheduled in #590. Since it is only relevant while a turn is running, keeping it next to Send makes the default action area more predictable; saved custom layouts remain untouched.

PR #989, opened 2026-07-03
fix(web): recover silently-dead SSE connections quickly, stop banner noise

longcw:
SSE connections can die without the browser ever noticing (device suspend, tab freeze, network path loss — no FIN/RST arrives). The webui only detected this via the 90s heartbeat watchdog, so every silent death meant up to 90s of stale UI followed by an alarming "Reconnecting... (heartbeat timeout)" banner popping mid-use. Verified against a live hub: heartbeats are sent every 30.0s without fail, so these banners were purely a client-side detection problem.

- useSSE: on visibility resume, distrust the connection after one missed heartbeat interval (45s) instead of 90s — a connection killed while the device was suspended is now replaced the moment the user returns
- useSSE: abandon connection attempts that don't open within 10s (hung on a dead pooled socket after resume) instead of waiting for the watchdog
- useSSE: first reconnect attempt is immediate; exponential backoff now starts from the second attempt
- App: only show the reconnecting banner when the disconnect persists beyond 3s (useDelayedTrue), so sub-second recoveries stay invisible while real outages still surface

via [HAPI](https://hapi.run)

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -8,6 +8,7 @@ import { initializeThemeColors } from '@/hooks/useThemeColors'
 import { useAuth } from '@/hooks/useAuth'
 import { useAuthSource } from '@/hooks/useAuthSource'
 import { useServerUrl } from '@/hooks/useServerUrl'
+import { useDelayedTrue } from '@/hooks/useDelayedTrue'
 import { useSSE } from '@/hooks/useSSE'
 import { useSyncingState } from '@/hooks/useSyncingState'
 import { usePushNotifications } from '@/hooks/usePushNotifications'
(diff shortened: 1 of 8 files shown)


swear01:
Hi @longcw — this exact problem is being reported by users in the wild: on mobile, backgrounding the tab and returning leaves a persistent "Reconnecting... (stream error)" banner even though the UI keeps working (reconnect hangs on the dead pooled socket; see the analysis in https://github.com/tiann/hapi/issues/1559).

Your PR contains exactly the missing pieces (45s visibility-resume threshold, 10s connect-open timeout, immediate first retry), and it was closed without merging — was there a specific reason (e.g. concerns about the thresholds), or would you be OK with it being revived/carried in a trimmed form? The banner grace-period half of the fix has since landed as #1230, so only the reconnect-reliability half (#1559) remains.

swear01:
Hi @longcw — following up: I opened https://github.com/tiann/hapi/pull/1562 which reimplements the reconnect-reliability half of this PR (45s visibility-resume threshold, 10s connect-open timeout, immediate first retry) on top of the current scheduler, with tests. The banner grace-period half already landed as #1230. Context in https://github.com/tiann/hapi/issues/1559 — happy to adjust if you closed this for a specific reason.

PR #995, opened 2026-07-07
fix(codex): improve local session import compatibility

Himehane:
fix(codex): improve local session import compatibility

- fix duplicated user messages during transcript import
- fix duplicated assistant messages during transcript import
- normalize imported messages before deduplication
- improve compatibility with recent Codex transcript format
- support session titles from session_index.jsonl
- preserve fallback behavior for older Codex versions
- add regression tests for transcript parsing and title resolution


Himehane:
Verified locally.

All tests passed successfully.

- 134 test files passed
- 1131 tests passed

Additionally verified the Codex import behavior manually:
- Duplicate user messages are no longer imported.
- Duplicate assistant messages are no longer imported.
- Session titles are correctly loaded from session_index.jsonl.

Issue #1015, opened 2026-07-11
web(codex): SessionHeader missing reasoning effort; create page missing Fast/Personality/Plan


swear01:
## Summary

Codex remote sessions surface reasoning effort inconsistently across the web UI, and the create-session form cannot set several Codex runtime options that the in-chat settings panel already supports.

Reproduced on a local isolated hub (`:43006`) + Vite web against current upstream `main`, with Codex `gpt-5.6-sol`.

### Current

| Surface | Shows `reasoning …`? |
|---|---|
| Top `SessionHeader` (back / title / flavor / model) | **No** |
| Composer `StatusBar` (above input) | **Yes** (`reasoning default` / `reasoning xhigh` / …) |

Users looking at the top bar cannot see the active effort. Expected: **both** SessionHeader and StatusBar show the same reasoning label for `codex` / `opencode`.

### Notes

- StatusBar already formats via `formatCodexReasoningLabel(session.modelReasoningEffort)`.
- SessionHeader currently only shows flavor, model label, and worktree branch.
- Related: when `modelReasoningEffort` is unset (`null`), the UI shows `reasoning default` even if `~/.codex/config.toml` has e.g. `model_reasoning_effort = "xhigh"`. StatusBar reflects HAPI session state, not Codex config defaults.

## Bug / gap 2 — Create Session missing options available in chat Settings

On **Create Session** with agent = Codex, the form exposes Model + Reasoning effort + YOLO.

Chat composer **Settings** additionally exposes:

- Collaboration Mode (Default / Plan)
- Personality
- Permission Mode
- Fast Mode (Standard / Fast) when the model catalog advertises a Fast tier

`SpawnSessionRequestSchema` also does not accept `serviceTier` / personality / collaboration, so Create cannot pass Fast even if UI were added.

### Live probe (create vs chat)

Create Codex page:

- Reasoning effort: present (catalog can include Ultra)
- Fast Mode / Personality / Collaboration: **absent**

Chat Settings on an active Codex session:

- Fast Mode / Personality / Collaboration / Reasoning Effort: **present**

## Expected

1. SessionHeader shows the same Codex/OpenCode reasoning label as StatusBar (and an explicit Fast badge only when `serviceTier === 'fast'`, without effort/model heuristics — see #1004).
2. Create Session can set the same Codex options that chat Settings can (or document intentional create-time limits). Short-term: at least surface Fast Mode if spawn plumbing is extended.

## Related

[...]

Issue #1003, opened 2026-07-10
feat(codex): expose CLI personality selection in remote sessions


swear01:
## What feature would you like to see?

Expose Codex CLI personality selection in remote HAPI sessions.

Codex CLI supports `friendly`, `pragmatic`, and `none`, and the app-server protocol already accepts `personality` on `thread/start`, `thread/resume`, and `turn/start`. HAPI declares those protocol fields but currently has no session state, composer selector, slash handling, or launcher wiring for them.

## Expected behavior

- Add a Codex-only personality setting to the session/composer runtime controls.
- Pass the selected personality to Codex app-server turns (and thread start/resume where appropriate).
- Support `/personality friendly|pragmatic|none`; querying without an argument should report the current value.
- Default/unset must preserve the user's Codex configuration instead of forcing a personality.
- Keep this scoped to Codex CLI parity; it is unrelated to HAPI voice personality.
- Add focused state propagation, slash parser, and turn-param tests.

PR #1017, opened 2026-07-11
fix(web): expose Codex Fast and Plan on Create Session

swear01:
## Summary

- Extend spawn (`SpawnSessionRequest` → hub → runner → `hapi codex`) to accept optional `serviceTier` and `collaborationMode`.
- Add Create Session selectors for Collaboration (Default/Plan) and Fast Mode (gated by catalog `serviceTiers`, same as chat Settings).
- Closes the Create-page gap called out in #1015 (Bug/gap 2), excluding Personality.

**Personality:** still blocked on #1003 / #1009 — upstream chat Settings does not expose Personality yet, so Create deliberately matches current chat Settings (Fast + Plan only). Follow-up after #1009 merges.

## Test plan

- [ ] Create Codex session with Collaboration = Plan → new session starts in plan mode (StatusBar / Settings show Plan)
- [ ] Create Codex session with Fast Mode = Fast (when model advertises Fast) → `serviceTier` is `fast`
- [ ] Fast Mode selector hidden when catalog does not advertise Fast for selected/default model
- [ ] Non-Codex agents: selectors absent; spawn payload omits the new fields
- [ ] `bun run test -- src/commands/codex.test.ts src/runner/buildCliArgs.test.ts` (cli)
- [ ] `bun run test -- src/components/NewSession/newSessionFormDraft.test.ts` (web)

### DIFF
--- a/cli/src/api/apiMachine.test.ts
+++ b/cli/src/api/apiMachine.test.ts
@@ -441,7 +441,7 @@ describe('ApiMachineClient Codex transcript handlers', () => {
 
         client.setRPCHandlers({
             spawnSession,
-            stopSession: vi.fn(async () => {}),
+            stopSession: vi.fn(() => true),
             requestShutdown: vi.fn()
         })
 
--- a/cli/src/api/apiMachine.test.ts
+++ b/cli/src/api/apiMachine.test.ts
@@ -470,6 +470,7 @@ describe('ApiMachineClient Codex transcript handlers', () => {
         }
     })
 })
+})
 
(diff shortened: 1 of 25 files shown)


swear01:
Fixed HAPI Bot Major: `collaborationMode` now forwarded in `cli/src/api/apiMachine.ts` (with regression test). Also pass it on hub resume spawn so Plan survives restart.

swear01:
Addressed follow-up Minor: do not clear restored `serviceTier: 'fast'` while the Codex model catalog is still loading.

PR #1029, opened 2026-07-12
fix(web): prioritize session names in narrow sidebar

swear01:
## Summary

- reserve a minimum-width primary column for the agent icon, session name, and live status
- let trailing metadata use the remaining width, with the timestamp truncating before the session name becomes unreadable
- expose the full session name through the native hover title

## Why

At the sidebar's minimum width, the session row used a flex layout whose trailing metadata was `shrink-0`. Todo progress and timestamps therefore forced the session name down to only a few visible characters, even though the name is the row's primary identifier.

The new two-column grid gives the primary content a `9rem` minimum track while the secondary track can shrink to the available space. This applies to both the resizable desktop sidebar and narrow mobile layouts without relying on viewport breakpoints.

## Impact

Session names remain recognizable in narrow sidebars. Status, todo progress, and pending-request indicators remain visible when possible; the lower-priority timestamp yields space first.

## Testing

- `bun typecheck`
- `bun run test` — 2,975 passed, 3 skipped, 0 failed

### DIFF
--- a/web/src/components/SessionRowSummary.tsx
+++ b/web/src/components/SessionRowSummary.tsx
@@ -150,10 +150,13 @@ export function SessionRowSummary(props: {
 
     return (
         <div className={`flex w-full min-w-0 flex-col gap-1 ${className ?? ''}`}>
-            <div className={`flex items-center justify-between gap-3 ${!s.active ? 'opacity-50' : ''}`}>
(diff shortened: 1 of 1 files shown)


heavygee:
Rebased onto current `main` to clear merge conflicts.

`SessionList` now renders through `SessionRowSummary`, so the narrow-sidebar name-priority layout (minmax primary track + truncating timestamp) is applied there instead of the old inline row markup. Intent unchanged.

PR #1051, opened 2026-07-16
fix(web): prefer live Hub connectivity over navigator status

### LINKED ISSUE #1050: Offline banner shows while the Hub is reachable
## Description

The web app can display `You are currently offline. Some features may be limited.` even while the HAPI Hub is reachable and chat requests and responses work normally.

`OfflineBanner` currently relies exclusively on `navigator.onLine`. That browser hint is known to report `false` in otherwise-working environments, especially with system proxies, VPN/TUN networking, embedded browsers, or platform connectivity checks.

## Steps to reproduce

1. Access a HAPI Hub through the official relay from a browser where `navigator.onLine` reports `false`.
2. Open an existing session and send a message.
3. Observe that the message and response work normally through the Hub.
4. Observe that the offline banner remains visible.

Reproduced on macOS in both a regular browser and the WeChat embedded browser while controlling a Windows-hosted HAPI Hub through `relay.hapi.run`.

## Expected behavior

A proven live Hub connection should take precedence over the `navigator.onLine` hint. If the Hub SSE connection drops, the UI should show its reconnecting state regardless of the browser hint.

## Actual behavior

The offline banner remains visible solely because `navigator.onLine` is `false`, and it also suppresses the more accurate SSE reconnecting banner.

## Suggested fix

- Hide `OfflineBanner` when the global Hub SSE subscription is connected.
- Let the reconnecting banner take precedence during an SSE outage.
- Retain `navigator.onLine` only as a fallback before Hub connectivity has been established.

Aqi77su:
## Summary

- suppress the offline banner when the global Hub SSE subscription is connected
- show the reconnecting banner from the actual SSE state even when `navigator.onLine` is false
- add focused coverage for false offline reports and banner precedence

Closes #1050.

## Rationale

`navigator.onLine` is only a browser hint and can remain false with working proxy, VPN/TUN, embedded-browser, or platform networking. A live Hub SSE subscription is stronger evidence of connectivity and should take precedence.

## Testing

- Added focused Vitest coverage for `OfflineBanner` and `ReconnectingBanner`.
- Local execution was attempted with the repository's CI Bun version, but dependency installation stalled on the Windows host before the test runner became available. GitHub CI should run the full typecheck and test suite.

## AI disclosure

Implementation and tests were produced with OpenAI Codex (GPT-5) and manually scoped against the existing Hub/SSE state flow.


Aqi77su:
Addressed the banner-precedence finding in 452e4c4. `OfflineBanner`, `ReconnectingBanner`, and the PWA status offset now share `showReconnectingBanner`, so the offline fallback is only suppressed when the reconnecting banner is actually eligible to render.

Aqi77su:
Addressed the follow-up test failure in 9f2e50d. Both banner test suites now render their components inside the repository's `I18nProvider`, so `useTranslation()` has the required context.

Aqi77su:
The approved CI run failed during the repository-wide `bun typecheck` after the branch had fallen behind `main`. I reproduced both `bun run typecheck:web` and the full `bun typecheck` locally with Bun 1.3.14; both pass. The PR branch is now updated to current `main` at 019f15c so CI can rerun against the current base.

PR #1090, opened 2026-07-19
feat(host): add managed providers and workspace controls

zulinliu:
## Summary

Adds runner-backed host management for API providers, Git repositories, and workspace files. The feature is available through the existing Hub RPC boundary and Web settings/workspace surfaces, so remote users can manage the selected runner without exposing host credentials to the Hub.

- manage provider profiles and switch the active provider/model for Claude Code, Codex, and Grok
- manage repositories through clone, status, commit, push, remote, fetch, and branch workflows
- manage workspace files through create, copy, move, delete, preview, edit, and download workflows

## Provider management

- Add protocol-aware provider profiles with runner-local credential storage, health checks, and model discovery.
- Keep native agent model choices alongside dynamically discovered provider models. A configured 1M-capable model exposes an additional [1M] choice while preserving the provider's base model ID for requests.
- Apply provider and model selection to both new sessions and existing session settings, including Codex and Grok launch paths.
- Add typed shared RPC contracts and Hub route validation for provider, repository, and file operations.

### Git

- Clone repositories; inspect status; stage, commit, and push changes; update remotes; fetch updates; and create, switch, update, or delete branches.
- Prefer gh when available, with Git fallback, and provide conventional commit templates in the UI.

### Files

- Create files and directories, copy or move them to a selected or manually entered destination, and choose overwrite or keep-both behavior for conflicts.
- Preview supported files, edit text files, and download individual files.
- Package directory downloads as ZIP archives, including the selected root directory and empty directories.

## Compatibility

The merge with current main deliberately keeps both session-creation paths: existingSessionId for Codex import/recovery and providerProfileId for provider selection. It also preserves the current macOS Desktop/CLI Codex selection while retaining Windows command handling and custom-provider startup arguments.

## Tests

- bun typecheck
- git diff --check
- CLI targeted suites: 111 passed
- Hub native suite: 97 passed
- Web targeted suites: 49 passed
- Provider, repository, file-management, and ZIP-download focused coverage
- Manually verified against a local Hub and Runner, including LAN access and folder ZIP contents with unzip -Z1

The full root bun run test was also attempted during development, but the runner integration phase did not produce a complete final summary. Its temporary package-version mutation was restored and is not part of this PR.

## AI disclosure

Implementation and tests were assisted by OpenAI Codex (GPT-5).

PR #1091, opened 2026-07-19
feat(providers): add managed agent API profiles

zulinliu:
## Summary

Adds machine-local managed API provider profiles for Claude Code, Codex, and Grok Build.

- configure protocol-aware profiles, secrets, optional endpoints, and default or custom models
- test a provider, refresh its model list, and expose its models beside native Agent choices
- apply the selected profile to new sessions, resumed sessions, and terminal launches
- preserve the raw provider model ID when HAPI displays a separate model[1M] selection

## Security

- store secrets only in the selected runner's HAPI_HOME/providers.json with private permissions
- never persist provider secrets in the Hub or return them from profile-list responses
- inject credentials only into the selected Agent process
- validate Agent/protocol/credential combinations before the profile reaches the runner

## Documentation

- add a dedicated API Providers guide and sidebar entry
- document protocol support, native-login fallback, model health checks, 1M selection markers, terminal overrides, and secret boundaries
- update the root, CLI, Hub, and Web READMEs

## Tests

- bun typecheck
- git diff --check
- CLI Vitest: 37 passed
- Hub machine routes: 13 passed
- Web provider model tests: 9 passed
- Shared provider schemas: 5 passed
- reviewed the complete staged diff for credentials, accounts, private addresses, and credential-bearing URLs; no real secrets are included

## Scope

This is the provider-management half of the split for #1090. The scoped workspace file and Git controls are in a separate companion PR so each vertical feature can be reviewed independently.

## AI disclosure

Implementation and tests were assisted by OpenAI Codex (GPT-5).


zulinliu:
Fixed in 0cf0d70. Explicit system selection now clears only managed-provider metadata, preserving native Claude/Codex/Grok credentials. Added environment regressions plus local-resume coverage for providerProfileId: null.

swear01:
The session model picker can split the native catalog into duplicate groups when the persisted model is no longer returned by the catalog. `mergeModelOptions()` inserts a synthetic `Current` row at index 1, so the rendered order becomes `Native → model → Current → persisted model → Native → …`.

I do not think the session picker needs either a `Current` fallback category or source-group headers: the selected radio state already communicates the current choice. Please keep this picker as a flat, current model catalog; omit a persisted model that is no longer in that catalog rather than injecting it into the list.

PR #1092, opened 2026-07-19
feat(workspace): add scoped file and Git controls

zulinliu:
## Summary

Adds scoped workspace controls for files and Git repositories on connected runner machines.

- create, upload, copy, move, delete, preview, edit, and download files or directories
- upload one file at a time into the current workspace directory, up to 20 MiB
- prevent default upload overwrite by creating a numbered copy for a duplicate name
- download directories as ZIP archives, including the selected root and empty directories
- inspect repositories; stage, commit, push, update remotes, fetch, and manage branches
- run long file and Git operations as cancellable runner jobs with progress reporting

## Security

- require every file and Git path to resolve inside configured workspace roots
- canonicalize paths and reject symlink escapes
- reject .git metadata changes through file operations
- validate uploaded file names and Base64 content before runner-side writes
- use fixed Git argument arrays and reject remote URLs containing inline credentials

## Documentation

- add a dedicated Workspace Management guide and sidebar entry
- document workspace-root opt-in, file conflict policies, uploads, ZIP downloads, GitHub CLI preference with Git fallback, and access boundaries
- update the root, CLI, Hub, and Web READMEs

## Tests

- CLI, Web, and Hub TypeScript checks
- git diff --check
- File upload: CLI FileManager Vitest, 7 passed
- File upload: Hub machine routes, 12 passed
- File upload: Web WorkspaceBrowser Vitest, 3 passed
- File upload: Shared workspace schemas, 2 passed
- ZIP contents, branch operations, workspace scope, and inline-credential rejection are covered by focused tests

## Scope

This is the workspace-management half of the split for #1090. Managed API provider profiles are in #1091 so file and Git controls can be reviewed separately from agent credential handling.

## AI disclosure

Implementation and tests were assisted by OpenAI Codex (GPT-5).


zulinliu:
Fixed in f4cee30. Recursive delete/copy/move and replace now preflight directory trees and reject nested .git metadata. Directory ZIPs exclude .git while direct Git-metadata downloads are rejected. Added focused regression coverage and documentation.

zulinliu:
Follow-up fixed in 3e8f51a. create-file now shares the 2 MiB text-file schema limit, and changed downloads fail on a zero-byte chunk read instead of returning a non-advancing offset. Added focused schema and truncation regressions.

zulinliu:
Follow-up pushed in 88267aa.

- reject overlapping batch paths before filesystem mutation
- enforce the 2 MiB text limit in UTF-8 bytes
- serialize shared Git preference updates
- fix containment checks for valid names beginning with `..`

Added focused regressions for each case. Typecheck and CLI/Web/Shared suites pass; the one load-sensitive Hub migration timeout passes with an extended local timeout.

zulinliu:
Addressed the latest workspace-boundary review finding in c5f03f2. Local clone sources and set-remote paths, including `file:` URLs and relative paths, now resolve through WorkspaceScope before reaching Git. Network URLs and GitHub shorthand remain supported. Added regressions for out-of-root rejection and in-root local remotes.

zulinliu:
Addressed the latest file-read boundary findings in 5db8f88. Host listings and previews now reject Git metadata paths (including canonicalized symlink aliases), and direct downloads retain the validated file handle through release so path replacement cannot retarget chunk reads. Added focused `.git` visibility/preview and symlink-retarget regressions.

zulinliu:
Addressed the latest review finding in `df80271`:

- reject leading-dash Git remote names before they reach `git push`/`fetch`/`remote`
- reject the special remote names `.` and `..`
- add schema regression coverage across push, fetch, set-remote, and delete-remote-branch

Verification:
- `bun typecheck`
- `bun run test:shared` (119 passed)
- `cd cli && bun test src/host/repositoryManager.test.ts` (8 passed)

The full local suite also reached unrelated Hub migration/Cursor tests, where existing fixed 5s timeouts fired under local load; the affected focused suites pass.

zulinliu:
Addressed the branch-name follow-up in `febbd3a`.

The existing schema did accept slash-separated names (the existing `feature/workspace-controls` assertion and direct checks passed), but its compact regex was ambiguous and did not enforce the full Git ref rules. The replacement now:

- keeps slash-separated branch names valid
- rejects empty/dot-prefixed/`.lock` path components, `//`, `..`, `@{`, special `@`, leading dashes, control bytes, and Git-forbidden characters
- retains Git-valid characters such as `]`
- adds negative regression coverage, including `feature//workspace`

Verification:
- `bun typecheck`
- `bun run test:shared` (120 passed)
- `cd cli && bun test src/host/repositoryManager.test.ts` (8 passed)

zulinliu:
Codified the review lessons in `403c558`:

- added `docs/agents/host-operation-security.md` with the trust model, eight required invariants, an operation matrix, adversarial test requirements, and a pre-push workflow
- linked the checklist from `AGENTS.md` for filesystem/workspace/download/Git/subprocess changes
- added the same high-risk checks to the automated PR review prompt

Verification:
- `bun typecheck`
- `cd docs && bun run docs:build`
- reference and diff checks passed
- full local suite: 492 passed; 10 existing Hub migration/Cursor tests exceeded fixed 5s timeouts under local I/O load (no code paths changed by this documentation commit)

zulinliu:
Addressed the `.git` symlink destination finding in `312d437`:

- upload destinations now preserve and reject an original `.git` path before canonicalization
- copy/move execution and lock-key resolution use the same protected destination resolver
- added red/green regressions for upload, copy, and move through `.git -> metadata`, including no-write/no-move assertions
- audited remaining `resolveReadable()` calls; file-plane reads/writes use protected resolvers, while remaining generic calls are WorkspaceScope internals or structured Git operations

Verification:
- `bun typecheck`
- CLI: 150 files, 1399 tests passed
- focused host scope/operation/file tests: 24 passed

zulinliu:
Addressed both remaining host-operation findings in `957234c`:

- directory archives now open and validate each file, compare path/opened/current inode identity, and stream directly from the retained handle instead of letting yazl reopen a path
- archive streams carry the validated size, so truncation/growth fails rather than bypassing the 256 MiB accounting
- traversal revalidates directory identity and containment; failures close handles and remove the temporary archive
- `resolveWritableExisting()` now rejects original and canonical `.git` paths through the protected resolver
- added deterministic regressions for external symlink retargeting, in-flight size changes, and `.git` directory symlinks

Verification:
- `bun typecheck`
- focused DownloadManager/WorkspaceScope: 9 passed
- full CLI: 1400 passed; one unrelated fixed-5s RepositoryManager timeout, which passed alone in 3.13s

zulinliu:
Addressed both point-of-use findings in `ad9bd78`:

- previews now open once, verify opened/current inode identity, read through the retained handle, and reject byte-count changes
- recursive copies validate each entry through the protected workspace resolver in `fs.cp`'s filter, blocking `.git` inserted after preflight and symlink escapes at traversal time
- generalized the concurrent `.git` regression to copy, move, and delete; move/delete recheck immediately before their filesystem call
- added a deterministic preview retarget regression that replaces the validated path with an out-of-root symlink

Verification:
- `bun typecheck`
- CLI: 150 files, 1405 tests passed
- focused host tests: 30 passed

zulinliu:
Addressed recursive mutation use-time protection in `41d2f09`:

- replaced recursive `rm` with a scoped walker using `lstat` + `unlink`/`rmdir`; it never follows a final symlink and rejects `.git` before and during traversal
- rechecks each parent before child removal and verifies directory identity/emptiness before `rmdir`
- directory moves now use protected copy followed by safe removal instead of whole-tree rename
- failed source removal rolls back the newly copied move target
- conflict replacement and cross-device cleanup use the same safe remover
- added later-timing regressions for `.git` appearing after copy and after recursive deletion starts

Verification:
- CLI typecheck
- FileManager: 23 passed
- full CLI: 1406 passed; one unrelated fixed-5s RepositoryManager timeout, which passed alone in 2.71s

zulinliu:
Addressed both resource-bound findings in `4427e6e`:

- added Hono streaming body-limit middleware before every machine host route, sized for the 20 MiB upload plus base64/JSON overhead
- host JSON parsing now preserves body-limit stream errors instead of converting them to a generic invalid-body 400
- added 413 regressions for both oversized `Content-Length` and a 32 MiB stream without a length header; neither reaches RPC
- archive limits now count every ZIP entry, including the root and empty directories, not only files
- added a real prepare-path test with an injected low entry limit

Verification:
- `bun typecheck`
- machines route tests: 14 passed
- DownloadManager tests: 7 passed

PR #1096, opened 2026-07-19
fix(cli): stop skill_lookup $name user-turn prepend (prompt-injection FP)

### LINKED ISSUE #1095: fix(cli): skill_lookup $name instruction prepended to Cursor user turns triggers prompt-injection false positive
## Summary

Cursor (and other ACP) remotes prepend HAPI's `skill_lookup` `$name` instruction to the **first user message**. Cursor agents treat that out-of-band line as **prompt injection**, waste reasoning on it, and ignore it.

## Instruction text

```text
When a user message starts with "$name", call HAPI's skill_lookup tool with "name" (without "$") before acting.
```

## Provenance

- PR #1035 (`feat(cli): add skill_lookup MCP for non-native agents`, closes #752)
- Constant: `cli/src/modules/common/skillLookupInstruction.ts`
- User-message prepend for: Cursor ACP (`cursorAcpRemoteLauncher.ts`), Kimi (`kimiRemoteLauncher.ts`), generic ACP (`agent/runners/runAgentSession.ts`)
- System-prompt path (better): Grok / OpenCode

## Root cause

Two stacked problems:

1. **User-turn prepend is prompt-taint.** The same class of bug was already fixed for inline media (`HAPI_MCP_BRIDGE_PROMPT` removed from ACP user turns; tool descriptions carry discovery instead). `SKILL_LOOKUP_INSTRUCTION` was left as a user prepend.

2. **Cursor ACP ignores `session/new` `mcpServers`.** See [Cursor forum](https://forum.cursor.com/t/acp-agent-silently-ignores-mcpservers-in-session-new/153623). So on stock upstream, #1035's "wire the tool into Cursor ACP" half does not reach the agent via ACP — only the instruction does. That matches the classic injection shape: glued to user text + references a tool the agent does not see.

Fork/soup dogfood already works around (2) with a native `.cursor/mcp.json` overlay + `agent mcp enable hapi` (open PR #958 / related). Even with the overlay (tool listed in `--tools` including `skill_lookup`), the user-message prepend still triggers injection false-positives.

## Expected

- No `$name` / `skill_lookup` instruction glued onto user turns.
- Where the tool is available, discovery lives in the MCP tool description (and/or system prompt), matching OpenCode/Grok and the media-tool pattern.
- Cursor actually receives `skill_lookup` via a channel it honors (native MCP overlay or future ACP fix), not only via ignored `session/new` mcpServers.

## Suggested fix

1. Remove `SKILL_LOOKUP_INSTRUCTION` user-message prepend from Cursor / Kimi / generic ACP.
2. Keep (or add) the `$name` guidance on the `skill_lookup` MCP tool description.
3. For Cursor tool reachability: land / keep the `.cursor/mcp.json` overlay path (#958 territory); do not reintroduce user-turn prompts.

## Not the same as

- #752 / #1035 feature request itself — this is a regression/gap in how the instruction was delivered on Cursor ACP.

heavygee:
## Summary
- Stop prepending `SKILL_LOOKUP_INSTRUCTION` onto the first user turn for Cursor ACP, Kimi, and generic ACP remotes.
- That glued-on line was treated as prompt injection (and on Cursor, `session/new` mcpServers are ignored anyway — tool discovery belongs on the MCP tool description / system prompt).
- OpenCode and Grok keep the instruction in their system-prompt helpers. The `skill_lookup` tool description already carries the `$name` guidance.

Fixes #1095

## Test plan
- [x] `bunx vitest run` on cursor / kimi / runAgentSession / opencode systemPrompt suites (23 passed)
- [x] Dogfood: Cursor remote session first user message no longer starts with the `$name` instruction line

### DIFF
--- a/cli/src/agent/runners/runAgentSession.test.ts
+++ b/cli/src/agent/runners/runAgentSession.test.ts
@@ -180,8 +180,10 @@ describe('runAgentSession', () => {
 
         const firstPrompt = JSON.stringify(harness.prompts[0])
         const secondPrompt = JSON.stringify(harness.prompts[1])
-        expect(firstPrompt).toContain('$name')
-        expect(firstPrompt).toContain('skill_lookup')
+        expect(firstPrompt).toContain('first')
+        expect(firstPrompt).not.toContain('skill_lookup')
+        expect(firstPrompt).not.toContain('$name')
+        expect(secondPrompt).toContain('second')
         expect(secondPrompt).not.toContain('skill_lookup')
     })
 })
--- a/cli/src/agent/runners/runAgentSession.ts
+++ b/cli/src/agent/runners/runAgentSession.ts
@@ -16,8 +16,6 @@ import { PermissionModeSchema } from '@hapi/protocol/schemas';
 import { isPermissionModeAllowedForFlavor } from '@hapi/protocol';
 import { RPC_METHODS } from '@hapi/protocol/rpcMethods';
 import type { SessionEndReason } from '@hapi/protocol';
-import { SKILL_LOOKUP_INSTRUCTION } from '@/modules/common/skillLookupInstruction';
-
 function emitReadyIfIdle(props: {
     queueSize: () => number;
     shouldExit: boolean;
@@ -101,8 +99,6 @@ export async function runAgentSession(opts: {
     let thinking = false;
     let shouldExit = false;
     let waitAbortController: AbortController | null = null;
-    let skillLookupInstructionSent = false;
-
     const syncKeepAlive = () => {
         session.keepAlive(thinking, 'remote', {
             permissionMode: currentPermissionMode
@@ -180,15 +176,11 @@ export async function runAgentSession(opts: {
                 continue;
             }
 
-            let messageText = batch.message;
-            if (!skillLookupInstructionSent && !messageText.trimStart().startsWith('/')) {
-                messageText = `${SKILL_LOOKUP_INSTRUCTION}\n\n${messageText}`;
-                skillLookupInstructionSent = true;
-            }
-
+            // skill_lookup discovery lives on the MCP tool description — do not
+            // prepend instructions onto user turns (prompt-injection false positive).
             const promptContent: PromptContent[] = [{
                 type: 'text',
-                text: messageText
+                text: batch.message
             }];
 
             thinking = true;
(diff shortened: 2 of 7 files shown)


heavygee:
@tiann - thanks for the collaboration model that made this PR possible 🙏 

I went ahead and squash-merged #1096 after CI and the review bots were clean. That was me (heavygee) using write on a fix I believe is correct (#1095 - stop prepending the skill_lookup $name line onto Cursor/ACP user turns).

However I want to be transparent about process: this landed via an automated babysit agent following an internal brief that said “babysit to merge.” I had not written an explicit rule that write collaborators should not merge on tiann/hapi without maintainer say-so. That ambiguity was mine, not an assumption that fork maintainers should routinely merge upstream.

Going forward on my side: agents will continue to treat "babysit" as get-merge-ready + report only - but explicitly now, and will not run gh pr merge on this repo without explicit human approval.

Happy to follow whatever merge convention you prefer (maintainer-only merges, ping-before-merge, or fine as-is for small green fixes). Thanks again - and sorry for the process noise around a otherwise straightforward fix.

heavygee:
@swear01 — context note on scope, so #1035 doesn’t look “half deleted.”

This PR only removes the **user-turn prepend** of `SKILL_LOOKUP_INSTRUCTION` on Cursor ACP, Kimi, and generic ACP (`runAgentSession`). It does **not** remove:

- the `skill_lookup` MCP tool / resolver / auto-approve from #1035
- the `$name` guidance already on the tool description
- OpenCode / Grok keeping the same instruction text inside their existing first-prompt instruction bundles (`TITLE_INSTRUCTION` / `GROK_TITLE_INSTRUCTION`)

**Why cut that delivery path:** on Cursor (and similar remotes), gluing host instructions onto the first user message gets treated as prompt injection. That burns reasoning and is ignored. We’d already learned the same lesson for inline-media bridge prompts: discovery belongs on the tool (or a real instructions channel), not on user text.

**Why your approach made sense:** for agents without a durable system-prompt hook, first-prompt prepend was the available lever. The painful part is the surrounding game — Cursor ACP has historically ignored `mcpServers` on `session/new` even while advertising MCP capabilities ([forum report](https://forum.cursor.com/t/acp-agent-silently-ignores-mcpservers-in-session-new/153623)), so the instruction could land without the tool. Instruction-without-tool is exactly the shape injection detectors hate.

**What “correct” looks like (not fully open yet):** ACP actually connects the HAPI bridge tools → agent sees `skill_lookup` → description (already shipped in #1035) is enough. Alternate: a real system/instructions surface. Fork-side native MCP overlay is a workaround for Cursor reachability; we kept that out of this PR on purpose.

Hate the ACP/MCP mismatch + user-turn taint, not the #1035 design. Thanks for landing the tool half — that stays.

PR #1099, opened 2026-07-20
fix(web): avoid session misclicks during live reordering

NPUlrk:
## Summary

- prevent a session row from being selected solely because it moved under the pointer during a live sidebar reorder
- use the browser's native click targeting for normal activation while preserving long-press behavior
- cancel pending long-press timers even when the pressed element does not receive the release event
- add regression coverage for live session reordering and the long-press hook

## Root cause

Active sessions can reorder whenever SSE updates change `updatedAt`. The long-press hook previously invoked `onClick` from each row's `mouseup` handler without verifying that the press began on that same row.

If a row moved between mouse down and mouse up, the row newly positioned under the stationary pointer handled `mouseup` and navigated to the wrong session.

## Testing

- `bun typecheck`
- `bun run test`

## AI disclosure

This change was developed with assistance from OpenAI Codex. The reported behavior was reproduced with a regression test, and the resulting diff was reviewed and validated with the repository's full typecheck and test suites.

### DIFF
--- a/web/src/components/SessionList.directory-action.test.tsx
+++ b/web/src/components/SessionList.directory-action.test.tsx
@@ -219,6 +219,53 @@ describe('SessionList action menu parity', () => {
     })
 })
 
+describe('SessionList selection', () => {
+    it('does not select the row that moves under the pointer during a live reorder', () => {
+        const onSelect = vi.fn()
+        const first = makeSession({
+            id: 'session-first',
+            active: true,
+            updatedAt: 200,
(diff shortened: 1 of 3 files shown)


techotaku39:
Thanks for the original implementation. I have prepared a refreshed PR based on the current `main`: #1656. It keeps the touch ghost-click protection from #1185, prevents mouseup-driven selection during live session-list reordering, and adds regression coverage for the misclick scenario. The reported Edge PWA flow was manually verified on the task deployment; the focused Web tests pass. If #1656 is accepted, it supersedes this PR.

Issue #1117, opened 2026-07-21
bug(web): valid Mermaid diagrams (CLI strict+loose pass) silently fall back to raw source in chat


heavygee:
## Summary

Some **valid** Mermaid diagrams - ones that pass the Mermaid CLI (`mmdc`) render check in **both** `securityLevel: default` and `securityLevel: strict` - do **not** render in the HAPI chat UI. Instead they drop to the raw-source fallback (`MermaidFallback`, `<pre><code>`). The failure is **silent**: the operator sees the raw block, and there is no surfaced reason (no error text, no parser-version hint, no signal to the agent).

This is a **correctness / parity bug** (a false-negative fallback for input the reference renderer accepts), distinct from the two open feedback-loop enhancements:

- **#829** - route parse failures back to the agent (assumes the diagram is genuinely *broken*).
- **#850** - in-place silent correction loop for *broken* mermaid/tables (assumes the diagram is genuinely *broken*).

Both of those start from "the diagram is invalid, help the agent fix it." This issue is the opposite: the diagram is **valid** per the CLI, yet HAPI rejects it. The existing hardening from **#785 / #813** (pre-`parse()` gate + `suppressErrorRendering: true` + no-op parse-error handler + catch-all fallback) is exactly what makes the failure impossible to diagnose from the UI.

## Evidence (behavior only)

- A diagram renders successfully with the Mermaid CLI:
  - `mmdc -i diagram.mmd -o out.svg` (default security level) - **passes**
  - `mmdc -i diagram.mmd -o out.svg` with `--securityLevel strict` (or an equivalent `-c` config) - **passes**
- The identical fenced ` ```mermaid ` block in HAPI chat - **falls back to raw source**, no diagram, no error text.

(Proprietary diagram content withheld deliberately; the evidence here is behavioral: strict + loose both pass in the reference CLI, both fail in HAPI chat.)

## Why it is currently impossible to diagnose (the weak-feedback part)

In `web/src/components/assistant-ui/mermaid-diagram.tsx` the render gate swallows every diagnostic:

- `mermaid.setParseErrorHandler(() => undefined)` - discards parse errors.
- `mermaid.initialize({ securityLevel: 'strict', suppressErrorRendering: true, ... })`.
- `renderMermaidSvg()` does `const isValid = await mermaid.parse(code, { suppressErrors: true }); if (!isValid) return null` - a `false` return produces the fallback with no captured reason.
[...]

Issue #829, opened 2026-06-07
feat(hub+cli): surface mermaid parse failures back to the agent so it can self-correct


heavygee:
## Summary

The web client already validates mermaid blocks before rendering (explicit `mermaid.parse()` + `mermaid.render()` wrapped in try/catch, `securityLevel: 'strict'`, `suppressErrorRendering: true`, fallback to raw source on failure - see `web/src/components/assistant-ui/mermaid-diagram.tsx`). That stack prevents broken charts from XSS-ing or crashing the page (closed by #785) and is the foundation the lightbox in #737 / #741 sits on top of.

What it does **not** do: tell the **agent** that its diagram failed to parse. The validation loop is one-way. The user sees raw mermaid source in a code-block fallback; the agent sees nothing and happily emits another broken chart next turn. Over a long session, this produces a quiet UX regression - the operator stops getting diagrams entirely until they manually point out "your charts have been raw text for 20 messages."

This is the same class of gap that #675 / #676 closed for OpenCode errors ("invisible failure"), applied to the agent-side feedback path instead of the user-side display.

## Current behavior

1. Agent emits a markdown ```mermaid``` code block in assistant text.
2. Hub stores + forwards the message verbatim. No mermaid awareness.
3. Browser tries `mermaid.parse(code, { suppressErrors: true })`.
4. Parse returns falsy or `mermaid.render()` throws → `MermaidFallback` renders the raw source in a `<pre><code>` block.
5. User sees the raw text. Agent receives no signal. Hub has no record that anything failed.

No SSE event, no tool-result reply, no `system`-channel hint into the transcript.

### A. CLI-side post-emit re-parse (smallest)

In the CLI's outgoing assistant-text path, scan for ` ```mermaid ` fences, run a lightweight mermaid-parse (or call the existing client validator via `bun`/Node), and on failure inject a single `system`-channel turn back into the agent transcript:

> The previous mermaid block did not parse:
> ```
> <parser error excerpt, 1-3 lines>
> ```
> The block was preserved in the user-visible message as raw text. Please re-emit it with corrected syntax if the diagram was intended.

- Pros: hub stays mermaid-unaware; same diff in one place; works for every UI client.
- Cons: CLI carries a mermaid dependency (it does not today); version drift between CLI parser and browser parser could disagree.

### B. Client → hub → CLI signal (most accurate to what the user actually sees)

[...]

PR #1118, opened 2026-07-21
feat(web): surface Mermaid parse/render failure reason in fallback

heavygee:
## Summary

Addresses the weak-feedback half of #1117: when a Mermaid block fails in the HAPI chat UI, the fallback previously showed **only the raw source** with no reason. A diagram that renders fine in the Mermaid CLI but falls back in HAPI was therefore impossible to diagnose from the running UI, because every diagnostic was swallowed:

- `mermaid.setParseErrorHandler(() => undefined)` (no-op)
- `suppressErrorRendering: true`
- `mermaid.parse(code, { suppressErrors: true })` returning `false` with no message
- an empty `catch {}` around `render()`

This PR captures and surfaces the failure reason without weakening any of the #785 / #813 safety hardening.

## Changes

- `renderMermaidSvg()` now returns a `MermaidRenderOutcome` (`{ svg, error }`) instead of `string | null`.
  - The primary validation gate is unchanged: `parse(code, { suppressErrors: true })` still runs first, so Mermaid never injects its own error SVG and never fires global parse-error side effects.
  - On a falsy parse result it re-parses **once** without suppression **purely to capture the thrown reason**. The parse-error handler is a no-op and `suppressErrorRendering` stays on, so this has no rendering cost or side effect - it just lets the message surface.
  - `render()` failures capture their message too.
- `MermaidFallback` surfaces the reason (`mermaid.renderError` label + the concrete message) and exposes a `data-mermaid-error` attribute for automation / e2e. The raw source is still preserved verbatim in the `<pre class="aui-mermaid-fallback">`.
- The failure notice is **gated on an actual error**, so the brief async load window (svg not ready yet) does not flash a false "could not render" banner.

Uses the already-present-but-unused `mermaid.renderError` locale key (en + zh-CN), so no new strings.

## Why this and not the existing issues

#829 and #850 route/repair **broken** diagrams back to the agent. This is the prerequisite they both need: a captured, human-readable reason at the point of failure. It's also the tool that lets an operator tell a genuinely-broken diagram apart from a HAPI-only false-negative (the correctness half of #1117, deferred until a repro pins securityLevel vs mermaid-version drift).

Out of scope here (tracked in #1117): mermaid version-parity bump and any `securityLevel` default change.

## Test plan

- [x] `bun typecheck` - clean (cli / web / hub)
- [x] `bun run --cwd web test` - 158 files / 1299 tests pass
- [x] New unit tests: parse-rejection surfaces the reason (+ `data-mermaid-error`, render not called); `render()` throw surfaces the reason; raw source preserved verbatim in both
- [x] `mermaid-diagram.live.test.tsx` (real mermaid render in jsdom) still passes

## Related

- #1117 - valid-diagram false-negative + weak feedback (this PR is the feedback half)
- #785 / #813 - render-failure hardening this preserves
- #829 / #850 - agent-side feedback / correction loops that depend on a captured reason

## Disclosure

Developed with assistance from an AI coding agent (Opus 4.8) acting as a peer in a multi-agent workflow.

### DIFF
--- a/web/src/components/assistant-ui/mermaid-diagram.test.tsx
+++ b/web/src/components/assistant-ui/mermaid-diagram.test.tsx
@@ -145,6 +145,45 @@ describe('MermaidDiagram', () => {
         expect(document.querySelector('[data-mermaid-lightbox]')).toBeTruthy()
     })
 
+    it('surfaces the parse error reason (not just raw source) when a diagram is rejected', async () => {
+        // First call is the suppressed validation gate (returns falsy); the
+        // second is the diagnostic re-parse that throws the human-readable reason.
+        mermaidMocks.parseMock.mockReset()
+        mermaidMocks.parseMock.mockResolvedValueOnce(false)
+        mermaidMocks.parseMock.mockRejectedValueOnce(new Error('No diagram type detected in text'))
+
+        const code = 'grph TD\nA --> B'
+        renderDiagram({ code, language: 'mermaid', components: defaultComponents })
+
+        await waitFor(() => {
+            const reason = document.querySelector('.aui-mermaid-fallback-reason')
+            expect(reason?.textContent).toContain('No diagram type detected in text')
+        })
+
+        // Raw source is still preserved verbatim in the fallback <pre>.
+        expect(document.querySelector('.aui-mermaid-fallback')?.textContent).toBe(code)
+        // Machine-readable hook for automation / e2e.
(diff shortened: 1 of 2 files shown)


heavygee:
Closing as premature. Per the fork's feature-work-lifecycle, upstream PRs come **after** peer-stack/soup dogfood + explicit operator approval - I opened this before that step. The branch (heavygee:feat/mermaid-render-diagnostics) and the tracking issue #1117 stand; I'll reopen once the change is dogfooded and approved. Apologies for the noise.

PR #1141, opened 2026-07-23
fix(web): use dedicated split breakpoint for compact tablets

teekalpha:
Some compact Android tablets report a landscape CSS viewport below
1024px despite having enough physical screen space, so the sessions
layout falls back to a single column. This adds a dedicated
tablet-friendly split breakpoint (920px) for the sessions layout
instead of relying on Tailwind `lg`.

- New `split: 920px` screen in tailwind.config.ts
- Sessions split columns + resize handle use `split:` instead of `lg:`
- The sidebar width/resize `@media` in index.css is moved to 920px in
  lockstep, so there's no 920–1024px band where the split shows but the
  sidebar width fails to apply
- Global `lg` breakpoint is unchanged; other pages are unaffected

Tested: full typecheck, all workspace test suites, and the production
web build pass. The generated CSS confirms that the `split:` utilities
and sidebar rules are emitted at `min-width: 920px`.

Manually verified on OPPO Pad mini in Chrome landscape at a 960px CSS
viewport with page zoom set to 100%.

### DIFF
--- a/web/src/hooks/useSidebarResize.ts
+++ b/web/src/hooks/useSidebarResize.ts
@@ -27,9 +27,16 @@ export function useSidebarResize() {
 
     const onPointerDown = useCallback((e: React.PointerEvent) => {
         e.preventDefault()
+        // The sidebar (the handle's previous sibling) can render narrower than the
+        // stored width when the viewport cap in index.css kicks in on a compact
+        // split. Seed the drag from the actual rendered width so there's no dead
+        // zone before the sidebar responds. Falls back to the stored width when the
+        // element or its measured width is unavailable (e.g. non-DOM test env).
+        const sidebarEl = e.currentTarget.previousElementSibling as HTMLElement | null
+        const renderedWidth = sidebarEl?.getBoundingClientRect().width
         activePointerIdRef.current = e.pointerId
         startXRef.current = e.clientX
-        startWidthRef.current = width
+        startWidthRef.current = renderedWidth || width
         setIsDragging(true)
     }, [width])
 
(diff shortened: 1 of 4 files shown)


teekalpha:
Good catch. Fixed in ada0896 by capping the sidebar width against the
viewport:

```css
width: min(var(--sidebar-w), calc(100vw - 424px)) !important;
```

`424px = 420px min detail pane + 4px resize handle`, so the detail pane
keeps at least 420px — the same worst case as the old 1024px breakpoint
— all the way down to 920px, while staying a no-op on desktop.

Verified in a headless Chromium render of the built CSS with a persisted
600px sidebar (the worst case):

| viewport | sidebar | detail |
|---:|---:|---:|
| 919px | hidden | 919 (single column) |
| 920px | 496 | 420 |
| 960px | 536 | 420 |
| 1200px | 600 | 596 |

teekalpha:
Good point — fixed in 7b73aea. `onPointerDown` now seeds the drag from
the sidebar's rendered width instead of the stored value:

```ts
const sidebarEl = e.currentTarget.previousElementSibling as HTMLElement | null
const renderedWidth = sidebarEl?.getBoundingClientRect().width
startWidthRef.current = renderedWidth || width
```

So when the viewport cap has shrunk the sidebar (e.g. 600px stored ->
496px rendered at a 920px viewport), dragging responds immediately with
no dead zone. On desktop, rendered and stored widths match, so behavior
is unchanged. Verified in a headless Chromium render that
`handle.previousElementSibling` resolves to the sidebar and measures the
capped 496px at a 920px viewport.

PR #1212, opened 2026-07-28
feat(web): quote selected text from a message into the composer

hqhq1025:
Select any text in a message, hit the quote button, and it becomes a managed chip above the composer. On send it serializes into a Markdown blockquote.

> Stacked on #1211. That fix is included here because `QuoteSelectionPopover` is a Radix Popover — without the React dedupe it white-screens the page and the e2e suite can't run. Once #1211 lands, that commit drops out of this diff.

## Flow

```
drag-select text in a message
  │ selectionchange (debounced) + mouseup
  ▼
bubble appears above the selection      [ ❞ quote │ ⧉ copy ]
  │ click quote
  ▼
① source text gets a tint + numbered superscript
② a chip appears above the composer
③ composer takes focus, caret at the end
```

Roughly what it looks like with two quotes taken (chip 1 hovered, so marker ¹ lights up):

```
┌────────────────────────────────────────────────────────────┐
│ The converter passes unknown SDK message types…¹           │  ← tinted + marker
│ Do not delete the normalize fallback. It is deliberate.²   │
├────────────────────────────────────────────────────────────┤
│ ▏1 The converter passes un… ✕   ▏2 Do not delete the nor… │  ← chips
│ ┌────────────────────────────────────────────────────────┐ │
│ │ Why are these two handled differently?                 │ │
└─┴────────────────────────────────────────────────────────┴─┘
```

## Serialization

One quote — no numbering, since a lone `[1]` is just noise:

```markdown
> The converter passes unknown SDK message types straight through…

Why?
```

Two or more — numbered, so the agent can count them and refer to them:

```markdown
> **[1]**
> The converter passes unknown SDK message types straight through…

> **[2]**
> Do not delete the normalize fallback. It is a deliberate last line of defence.

Why are these two handled differently?
```

Three things worth noting about the format:

- **Line-by-line `> ` prefix**, not a wrapped block — keeps fenced code inside a quote valid and preserves indentation. Blank lines become a bare `>` so the blockquote doesn't split in two.
- **Numbering markers are `[N]`, not "Quote N"** — zero language dependency, no new i18n key, and no English label wedged into a Chinese message.
- **Serialization happens at send time**, not when quoting. So adding a second quote retroactively numbers the first. A plain-text-append approach can't do this: by the time you know there's a second quote, the first one's text is already sitting in the textarea.

## Design notes

**Quotes are objects, not text.** The obvious cheaper implementation is to append `> …` straight into the composer. That makes a mis-quote expensive to undo (hunt down and delete the right markdown lines), mixes quoted text with what you're typing, and rules out the retroactive numbering above.

**Not built on assistant-ui attachments.** `Attachment.type` is a closed union and `PendingAttachment.file` must be a real `File`, so quotes would need fabricated File objects, type-branching in the adapter's `add`/`remove`/`send`, synthetic blobs in the attachment-draft IndexedDB store, and an exemption from `hasAttachments` (which gates scheduled send — quotes are plain text and shouldn't be restricted). Splitting `hasAttachments` alone is most of the work of a separate layer.

**Colors reuse existing tokens** — `--app-md-quote-*` for the source tint and chips (it's literally the blockquote palette), `--app-chat-user-chip-*` for accents. No new theme variables; light/dark/amoled all covered already.

**Serialization has exactly one chokepoint.** Send has three entry points (`handleSend` plus two paths in `handleKeyDown`). Composing the text in the composer would mean touching all three, and missing one would produce "quotes work unless you press Cmd+Enter". `onNew` in `assistant-runtime.ts` is their common downstream, so it happens there once.

## Two constraints that shaped the implementation

**Source highlighting can't touch the DOM.** Messages are React-rendered markdown; injected `<mark>` nodes get wiped on the next re-render. It uses the CSS Custom Highlight API (`CSS.highlights` + `::highlight()`), which paints at render time without a DOM node. Where that API is unavailable (Firefox < 140), highlighting and markers are skipped and everything else works — deliberate progressive enhancement, with a dev-mode `console.info` so it isn't silent.

**Markers must be an overlay.** CJK text can break between any two characters, so an inline superscript regularly gets orphaned onto its own line. They're positioned from `getClientRects()` instead, and recomputed on resize.

## Testing

- **Unit** — serialization (incl. fenced code, blank lines, pre-existing `>`), storage round-trip and corruption handling, hook state incl. the cross-session persist race
- **Component** — selection resolution: inside a message, outside, cross-message, whitespace-only, touch
- **e2e (8 cases)** — driven with real pointer events via `page.mouse`, not `element.click()`

That last point isn't stylistic. An early version had the copy-confirmation tied to the selection staying alive; a real user moves the mouse after clicking, the selection dies, and the confirmation vanished instantly. A script-driven click leaves the pointer parked and never reproduces it. There's a regression test for that, and one for the Radix autofocus clearing the selection (which silently breaks the whole feature).

```bash
cd web
npm run test                       # 1560/1560
npx playwright test e2e/quote-selection.spec.ts   # 8/8
```

## Desktop only

Gated on `!isTouch`. Native selection handles fight a custom bubble, `useLongPress`'s contextmenu `preventDefault` can interfere with selection, and `interactive-widget=resizes-content` means the keyboard resizes the viewport out from under a floating layer. Each needs its own answer, so mobile is deliberately out of scope here rather than half-done.

One consequence to flag: the chip's remove button appears on hover, so a mobile version would need it always visible.

## Open questions

I'd rather ask than guess:

1. **Should mobile be in scope at all?** If yes I'm happy to do it as a follow-up — I just didn't want to ship a compromised version of it silently.
2. **Numbering order** — currently by the order you quoted, not document order. "The one I picked first is [1]" felt closer to intent, but footnote convention would argue the other way.
3. **Should the source message id go to the agent?** Right now it's frontend-only (for jump-to-source). Sending it would let the agent disambiguate, at the cost of noise in the prompt.
4. **Cap on quote count** — `QUOTES_MAX = 50` purely as a runaway guard, not a product decision.


hqhq1025:
Thanks for merging #1211.

Heads-up on this PR's diff: `web/vite.config.ts` still shows up here. It's the same 17 lines you just merged — this branch still carries the original commit, and the squash gave it a different SHA on main, so GitHub can't tell they're the same change. GitHub reports the PR as mergeable, so it isn't blocking anything.

I'll rebase onto current main to drop it. Holding off for the moment because main has moved 8 commits ahead and scratchlist v2.2 touched `SessionChat.tsx` and both locale files, so the rebase needs conflict resolution plus a full re-verify. If you have feedback on the feature itself, I'd rather fold the rebase into that round than do it twice.

Say the word if you'd prefer the clean diff first.

hqhq1025:
Both findings were real. Fixed in 9dd4a700.

**[Major] Selection leaving the container** — confirmed and fixed with the suggested guard. Worth spelling out why the old check missed it: it validated the *message node* was inside the container, and on a drag that starts in a message the anchor node genuinely is. So the guard passed while `selection.toString()` had already swallowed whatever chrome the drag ended on, and that text went straight into the outbound prompt.

Added a regression test for it. I also wrote a mirror-image test (selection *starting* outside, ending inside) and then checked whether it actually fails without the fix — it doesn't. That path was already covered by the "no message node found" branch. Kept it as a behavioural lock with a comment saying exactly that, so nobody later assumes it's guarding the endpoint check.

**[Minor] Nested interactive controls** — also fixed, restructured to two sibling buttons inside a plain div as suggested. My original `role="button"` span only dodged the HTML validation error; the ambiguity the rule exists to prevent was still there. Added an e2e case asserting both controls are real `<button>`s, neither contains the other, and remove is reachable and activatable by keyboard alone.

**Verification** (bun is available locally, so I ran what the runner couldn't):
- `npx vitest run` — 1591/1591 pass
- `npx tsc --noEmit` — clean
- `npx playwright test e2e/quote-selection.spec.ts` — 9/9 pass

hqhq1025:
Rebased onto current main — the `vite.config.ts` duplication is gone, so this PR is now 21 files of nothing but the feature.

Git skipped the dedupe commit on its own (`skipped previously applied commit`), since the patch-id matched what you merged. No conflicts either, despite scratchlist v2.2 having touched `SessionChat.tsx` and both locale files.

Re-verified after the rebase rather than assuming a clean rebase means working code:
- `npx vitest run` — 1629/1629 pass
- `npx tsc --noEmit` — clean
- `npx playwright test e2e/quote-selection.spec.ts` — 9/9 pass

The open questions from the description still stand whenever you get to it — the mobile one especially, since that decides whether this lands as desktop-only or waits.

PR #1231, opened 2026-07-29
perf(hub): gzip the SSE stream without delaying delivery

hqhq1025:
## Why

SSE payloads are plain JSON that repeat the same field names on every event, so they compress well. Measured on real captured traffic from a hub with 15 active sessions:

| sample | plain | gzip (one-shot) | gzip (per-event flush) | ratio |
|---|---|---|---|---|
| normal load, 60s | 76 KB | 16.4 KB | 17.2 KB | **77.4%** |
| heavy load, 300s | 1.56 MB | 404 KB | 411 KB | **73.6%** |
| idle, 150s | 520 KB | 138 KB | 143 KB | **72.5%** |

JSON field names alone account for 38% of the raw stream.

This mostly helps clients on mobile data, weak links, or long-haul relays — it saves bytes and battery rather than fixing latency.

## Why it is not just `compress()`

Two obvious approaches both fail, and both fail **silently** — the stream still looks fine, events just never show up:

1. **Hono's `compress()` middleware is a no-op here.** It bails out when `Transfer-Encoding` is set (`hono/dist/middleware/compress/index.js:11`), and `streamSSE` always sets `chunked`. No `Content-Encoding` is ever emitted.

2. **`CompressionStream` buffers until the stream ends.** Measured on a 10-event stream: every event arrived at once, at the moment the stream closed (5.0s). On a connection that stays open for hours, that means events never arrive.

So this drives `zlib` directly and issues a `Z_SYNC_FLUSH` after each chunk. That costs about **one percentage point** of ratio (see table) and keeps delivery immediate.

## Verification

Verified in a real Chromium `EventSource`, not just unit tests:

```
first event   13ms      ✓ immediate
intervals     499, 500, 501, 501, 501 ms   ✓ one event per tick
onerror       none      ✓
```

Also confirmed at the socket level that the server flushes each event as it is written (first byte at 2ms, then one per 500ms tick).

Clients that do not send `Accept-Encoding: gzip` keep the uncompressed stream unchanged.

## Notes

No event payload or timing changes — transport only.

## Testing

- 5 new unit tests: pass-through without gzip, pass-through on unsupported encoding, byte-exact round trip, per-event flush (asserts each event is decompressible *before* the stream closes), and empty-body handling
- `tsc --noEmit -p hub/tsconfig.json` clean
- Full `hub` suite: 648 pass, 0 fail

### DIFF
--- a/hub/src/web/routes/events.ts
+++ b/hub/src/web/routes/events.ts
@@ -7,6 +7,7 @@ import type { SyncEngine } from '../../sync/syncEngine'
 import type { VisibilityState } from '../../visibility/visibilityTracker'
 import type { VisibilityTracker } from '../../visibility/visibilityTracker'
 import type { WebAppEnv } from '../middleware/auth'
+import { compressSseResponse } from '../sseCompression'
 import { requireSession } from './guards'
 
 function parseOptionalId(value: string | undefined): string | null {
@@ -77,7 +78,7 @@ export function createEventsRoutes(
             }
         }
 
-        return streamSSE(c, async (stream) => {
+        const response = streamSSE(c, async (stream) => {
             manager.subscribe({
                 id: subscriptionId,
                 namespace,
@@ -117,6 +118,8 @@ export function createEventsRoutes(
 
             manager.unsubscribe(subscriptionId)
         })
+
+        return compressSseResponse(response, c.req.header('Accept-Encoding'))
     })
 
     app.post('/visibility', async (c) => {
--- a/hub/src/web/sseCompression.test.ts
+++ b/hub/src/web/sseCompression.test.ts
@@ -113,3 +113,99 @@ describe('compressSseResponse', () => {
         expect(compressSseResponse(original, 'gzip')).toBe(original)
     })
 })
+
+describe('compressSseResponse cleanup', () => {
+    it('cancels the upstream stream when the client goes away', async () => {
+        // SSE clients disconnect mid-stream all the time; the source must be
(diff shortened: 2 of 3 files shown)


hqhq1025:
Self-review turned up three defects in the first version; pushed fixes in d65056b5.

**Cancelling the source threw.** The wrapper holds a reader for the whole life of the connection, and cancelling a locked stream is invalid — in Bun it throws `TypeError: Invalid state: ReadableStream is locked` synchronously out of the `cancel` callback. SSE clients disconnect mid-stream as a matter of course, so this fired on essentially every disconnect and the upstream cancel never ran. Now cancels through the reader, which is allowed to.

**Reads were not gated on downstream demand.** Only zlib's own buffer was consulted, and SSE compresses well enough that a slow client can be megabytes behind while the compressed queue still looks nearly empty — a test with a non-reading consumer pulled 1752 chunks before stalling. Reading now waits for `desiredSize` to go positive and is resumed from `pull()`.

**`Accept-Encoding` was a substring match**, so `gzip;q=0` — which means the client *refuses* gzip — was read as acceptance. Now parses the q-value.

Six new tests cover these (upstream cancel propagation, no rejection on client cancel, q=0 refusal, non-zero q, wildcard, and the backpressure stall).

Re-verified the property this PR exists for is intact: real Chromium `EventSource` still gets the first event at 13ms and each one after on its own 500ms tick, no error events. hub suite 654 pass / 0 fail.

PR #1242, opened 2026-07-29
feat(web): show status and platform on the machines settings page

hqhq1025:
Follow-up to #1214.

## Problem

Every row repeated itself:

```
msraima100gpu010            ← label, falls back to the hostname
msraima100gpu010 · linux    ← host · platform
```

The second line exists so a *renamed* machine can still be identified. With no custom name the label already **is** the hostname, so the line carries nothing. Since most machines never get renamed, that is the normal state rather than an edge case — the page reads as four rows of duplicated text.

## Changes

Restructure the row around what is actually distinct:

- **The hostname is the label** until a custom name exists, rendered in `--app-hint` so an unnamed machine reads as unnamed. Once named, the hostname drops to the second line, where it finally says something.
- **Platform becomes a chip** on the right, resolved through the existing `resolveMachineOsLabel`, so it reads `Linux` rather than the raw `linux`.
- **A status dot** distinguishes online from offline; rows are sorted online-first.

## Offline machines

A status dot is decoration if every row is online, and `GET /machines` returned `getOnlineMachinesByNamespace`. So it takes an optional `includeOffline=1`.

The default is unchanged. The session list, machine filter and new-session picker all want machines they can *spawn on*, and widening that for everyone would change what each of them shows. The settings page opts in, because the set of machines you can rename is not the set you can run on — `PATCH /machines/:id` already worked on offline machines (`requireMachine` never checked liveness), they just could not be listed.

Offline machines are cached under `['machines', 'withOffline']` so they cannot leak into the session list. The key stays prefixed with `queryKeys.machines`, so the existing `machine-updated` invalidation in `useSSE.ts:275` — which matches by prefix — keeps refreshing both entries with no change there.

## Touch

The name itself is now the tap target; the trailing button is demoted to a hover affordance and marked `aria-hidden`/`tabIndex={-1}` since it duplicates an action already reachable. In #1214 the hover button was one of two ways in, but hover does not exist on touch, so on a phone it was effectively invisible.

## Tests

Hub: 3 cases for the new query parameter (default stays online-only, `includeOffline=1` widens, any other value is ignored).

Web: 5 new cases — the page requests offline machines, an unnamed machine does not render its hostname twice, a named one reveals it underneath, online sorts before offline, and an offline machine can be renamed. The existing 10 still pass.

- `bun test` in `hub/`: 665 pass, 3 skip, **0 fail**
- `bun run typecheck` in `hub/`, `web/`: clean
- `bun run test` in `web/`: the 15 machines-settings tests pass. Note the repo currently has 20 failures across 5 unrelated files (`AgentSelector`, `AgentFlavorIcon`, `SessionList.directory-action`, `NewSession`) from the duplicate-React-instance problem; they fail on `main` at this commit with these changes stashed, and #1211 addresses them.


hqhq1025:
Both findings were real. Fixed in the latest commit.

**[Major] stale offline-inclusive cache** — worse than it first looked. Machine liveness does not travel through invalidation at all; `upsertMachine` writes directly via `setQueryData` against the exact `['machines']` key, so `['machines', 'withOffline']` was never touched. My claim in the PR description that prefix-based invalidation covers it was wrong for this path: prefix matching only helps the `invalidateQueries` call at `useSSE.ts:316`, and liveness never goes through it.

Writing to both keys would not have been enough either, because the two caches disagree about what offline means — the online-only list *drops* a machine that goes offline, while the settings list must *keep* it and render it as offline. So the fold is now an exported pure function with an explicit contract:

```ts
applyMachineToCache(previous, machine, { keepOffline: boolean })
```

applied once per cache, with the deletion path writing to both keys as well. Being a pure export makes it directly testable, matching how `isRenderIrrelevantPatch` and friends are already tested in that file. 8 new cases cover both contracts; I verified they bite by making the function ignore `keepOffline` — exactly the two offline-inclusive cases fail.

**[Minor] status hidden from assistive tech** — correct, the dot was `aria-hidden` with no text alternative. It now carries `role="img"` with a localised name (Online / Offline, en + zh-CN) plus a matching `title`, and there is a test asserting both are reachable by accessible name.

`bun run typecheck` clean; the 16 machines-settings and 24 useSSE tests pass.

PR #1257, opened 2026-07-30
fix(web): budget subagent messages separately from top-level history

hqhq1025:
## Problem

Scrolling up through a long session stalls after a few pages, and the history that *did* load is then thrown away.

The message window budgets raw `DecryptedMessage`s, but the timeline renders **folded** blocks. A subagent run emits one message per step and renders as a single Task card, so the two units disagree by orders of magnitude — and the budget is spent in the units the reader never sees.

Measured against a real session (`sqlite3` on the hub's `hapi.db`):

| | messages | share |
|---|---|---|
| sidechain (subagent) | 2655 | **96%** |
| top-level | 90 | 4% |
| total | 2745 | |

The 800-message budget bought roughly 43 top-level rows. The other ~1900 messages were unreachable — not because the session was too long, but because one subagent run can spend the entire budget on rows that collapse into a single card.

The distribution across sessions is wide enough that no single total works: 0%, 11%, 26%, 50%, 69%, 96% sidechain in the six largest sessions on this hub.

## Reproduction

Driving the real UI with Playwright, scrolling to the top and waiting 3.5s per round:

| round | `scrollHeight` | request |
|---|---|---|
| 1 | 2956 → 4209 | `beforeSeq=2508&limit=200` |
| 2 | → 5272 | `beforeSeq=2308` |
| 3 | 5272 (unchanged) | `beforeSeq=2108`, then nothing |
| 4 | 5272 (stalled) | — |
| 5 | **→ 2365** | — |
| 6 | 2365 | — |

Round 5 is the second half of the bug: loaded history is discarded, text length dropping 7356 → 3466. The scroll container itself is fine — `scrollTop` holds where you put it, and scrolling down to the bottom works.

## Approach

Give sidechain messages their own budget, exactly the way `AGENT_RUN_WINDOW_SIZE` already fences off Codex agent-run messages. `isSidechainMessage` reads the same stored field `normalizeAgent` reads — deliberately a per-message property, not the reducer's grouping, which is stateful and belongs in the reducer.

Two consequences that don't fall out for free:

**Sidechain messages must not outlive their Task card.** `tracer.ts:139` renders a parentless sidechain message at the top level, so a transcript whose `tool_use` was trimmed away would spill into the timeline as loose rows — inverting the bug rather than fixing it. `trimSidechain` drops anything older than the oldest surviving top-level row, keeping every retained sidechain message inside the span its parent can still occupy.

**Trimming them must not arm `requiresLatestReset`.** That flag means "the window no longer holds the live tail", and the next tail sync acts on it by refetching the latest page and replacing the window — which is precisely the discard observed in round 5 above. Trimming a subagent against its own budget always drops the newest of *them*, so the reset check now compares the newest **non-sidechain** message before and after the trim.

## Result

Same probe, after:

| round | `scrollHeight` | request |
|---|---|---|
| 1 | 14412 → 15466 | `beforeSeq=2637` |
| 2 | → 16719 | `beforeSeq=2437` |
| 3 | → 17782 | `beforeSeq=2237` |
| 4 | → 18887 | `beforeSeq=2037` |
| 5 | — | `beforeSeq=1837` |

Five pages instead of three, monotonically growing, no discard.

## Testing

Three tests in `message-window-store.test.ts`, each verified to fail when its own half of the change is disabled:

- `protects top-level rows from a subagent flood` — fails when the sidechain bucket is removed
- `drops subagent messages that fall behind the oldest surviving top-level row` — fails when the orphan boundary is removed
- `does not force a latest reset when only subagent messages overflow` — drives the real `syncTailMessages` path; fails when the reset check is removed

`31 passed` in this file (28 before). `bun run typecheck` clean.

### DIFF
--- a/web/src/lib/message-window-store.test.ts
+++ b/web/src/lib/message-window-store.test.ts
@@ -1030,6 +1030,67 @@ describe('history view and older pagination', () => {
         expect(kept.some((message) => message.id === 'top-0')).toBe(true)
         expect(kept.some((message) => message.id === 'top-9')).toBe(true)
     })
+
+    it('refetches the latest page when returning to the tail after a subagent overflow', async () => {
+        const id = sessionId('sidechain-recovery')
+        const getMessages = vi.fn(async () => latestResponse(
+            [makeAgentMessage({ id: 'top-0', seq: 1, at: 1 })],
+            { epoch: 7 }
+        )) as unknown as ApiClient['getMessages']
+        await syncTailMessages(createApi(getMessages), id)
+        expect(getMessageWindowState(id).epoch).toBe(7)
+
+        setMessageViewMode(id, 'history')
+        ingestIncomingMessages(id, Array.from({ length: SIDECHAIN_WINDOW_SIZE + 50 }, (_, index) =>
+            makeSidechainMessage(`side-${index}`, index + 100, index + 100)
+        ))
+        // Suppressing the reset while reading history leaves the window short of
+        // the transcript. Without redeeming that debt on the way back, the Task
+        // card would stay permanently incomplete.
+        expect(getMessageWindowState(id).epoch).toBe(7)
+
+        setMessageViewMode(id, 'tail')
+
+        expect(getMessageWindowState(id).epoch).toBe(null)
+    })
+
+    it('drops subagent messages left past the newest surviving top-level row', () => {
+        const id = sessionId('sidechain-orphan-tail')
+        ingestIncomingMessages(id, [makeAgentMessage({ id: 'anchor', seq: 1, at: 1 })])
+        setMessageViewMode(id, 'history')
+        // Prepend keeps the oldest top-level rows and drops the newest, so any
+        // subagent that started after the last survivor has lost its Task card.
+        // tracer.ts would render those parentless rows at the top level.
+        ingestIncomingMessages(id, [
+            ...Array.from({ length: HISTORY_WINDOW_SIZE + 20 }, (_, index) =>
+                makeAgentMessage({ id: `top-${index}`, seq: index + 2, at: index + 2 })
+            ),
+            ...Array.from({ length: 30 }, (_, index) =>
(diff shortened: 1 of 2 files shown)


hqhq1025:
Both findings were real. Fixed in 6957978.

**Sidechain-only tail truncation has no recoverable state** — correct, and the second half of that is the more serious one: suppressing the reset kept the reader in place but left nothing to ever refetch the missing transcript, so a Task card could stay permanently incomplete.

The suppression is now a *deferral*. `sidechainOverflowed` records that the window is short, and `enterTailMode` redeems it into `requiresLatestReset` — that being exactly the moment the reader chose to come back to the tail, so the refetch costs them nothing. `activateMessageWindow`'s early-return had to learn about the flag too, or re-opening the session would skip the redemption.

The reset condition itself was also the wrong shape. Comparing newest positions meant "no top-level rows at all" fell through to a reset on every overflow. It now reads the dropped set directly — *did any top-level row leave?* — which answers the question the flag actually asks and behaves consistently for a sidechain-only window.

**Prepend trimming can retain sidechain rows after their owning Task row is dropped** — also correct, I only guarded the old edge. Prepend drops the newest top-level rows, so a subagent that started after the last survivor has lost its parent too, and `tracer.ts:139` renders those at the top level.

Both edges are now enforced, with the new edge applied *only* under prepend. Under append it would be actively wrong: a subagent running right now legitimately sits past every top-level row, and cutting there would empty the live Task card.

**Coverage** for the three gaps called out, each verified to fail when its own logic is disabled:

- `refetches the latest page when returning to the tail after a subagent overflow` — history → overflow → tail, asserts the epoch is cleared
- `drops subagent messages left past the newest surviving top-level row` — prepend overflow with the parent Task trimmed
- `handles a window that holds nothing but subagent messages` — sidechain-only overflow

`34 passed` in this file. Full suite shows no change against baseline (same 128 environment-related failures before and after, +3 passing = the new tests). `bun run typecheck` clean.

PR #1269, opened 2026-07-31
test(cursor): remove per-file temp HAPI homes after cursor-models suites

heavygee:
Fast-follow to #1268 addressing the HAPI Bot `[Minor]` finding on that PR.

## What / why

The per-file HAPI homes created by the cursor-models test cluster were never removed, so every run leaked a directory into the system temp dir (`afterEach` only cleared the cache file *inside* them). Over many runs this litters `/tmp` with `hapi-cursor-models-*` dirs.

This restores the previous `HAPI_HOME` and recursively removes each per-file temp root in teardown:

- `cursorModels.test.ts` - `afterAll` restore + `rmSync` (the file the bot flagged)
- `cursorModelsSharedCache.test.ts` - `afterAll` restore + `rmSync` (the file the bot flagged)
- `cursorModelsStaleLock.test.ts` - same latent leak (`join(tmpdir(), 'hapi-cursor-models-lock-<pid>')` was never cleaned) - folded in so the whole cluster is leak-free

No source/runtime change - test hygiene only.

## Verification

- `bun typecheck` clean (cli+web+hub).
- 22 cursor-models tests pass.
- Ran the cluster 3x with a pre-cleaned `/tmp`: **0** leaked `hapi-cursor-models-*` dirs after each run (before the fix, one dir leaked per run per file).

The one unrelated `runner.integration.test.ts` failure on my machine reproduces identically on pristine `upstream/main` (live-runner-daemon collision, environment-specific) and passed in CI on the #1268 merge commit - not touched by this diff.

### DIFF
--- a/cli/src/modules/common/cursorModels.test.ts
+++ b/cli/src/modules/common/cursorModels.test.ts
@@ -1,13 +1,23 @@
-import { afterEach, describe, expect, test, vi } from 'vitest'
+import { afterAll, afterEach, describe, expect, test, vi } from 'vitest'
 import { setCursorAcpModelsSnapshot } from '@/cursor/utils/cursorAcpModelsBridge'
-import { mkdtempSync } from 'node:fs'
+import { mkdtempSync, rmSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 
 // Isolate the on-disk cursor-models cache to this file's own HAPI_HOME so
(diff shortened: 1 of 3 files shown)


heavygee:
Fast-follow to close the HAPI Bot [Minor] on #1268 - temp-dir cleanup, bot re-review is Findings:None. Merging this one too under the low-impact arrangement.

PR #1314, opened 2026-08-02
feat(web): pin running sessions in an in-progress section + fix stale push subscriptions

wu736139669:
## Summary
Session-list UX improvements for surfacing running sessions, plus a push-notification reliability fix.

## Changes
1. **Pinned "In progress" section** — active sessions are grouped at the top of the session list (before directory groups), sorted by recent activity, with a pulsing green indicator. Rows show their project name and the section is collapsible. Inactive sessions remain grouped by directory below.
2. **Running badge** — active sessions show a green pulsing dot ("运行中 / Running"); thinking sessions use a green spinner; pending permission/input attention still takes priority.
3. **Push re-subscribe on VAPID key change** — browsers keep a stale push subscription when the hub's VAPID keys change (e.g. hub migration), so push services reject sends with `VapidPkHashMismatch`. The web client now records the VAPID public key used at subscribe time and automatically unsubscribes/re-subscribes when it differs.
4. **Auto-expand fix** — opening a pinned running session no longer expands the directory group below it (auto-expand only triggers for sessions actually rendered inside a group).

## Notes
- Mobile-friendly (works with pull-to-refresh and machine filters).
- Tested against hub v0.25.3.


wu736139669:
Addressed all three findings from the HAPI Bot review:

1. **[Major] VAPID rotation leaves the obsolete endpoint registered in the hub** — `usePushNotifications` now calls `unsubscribePushNotifications({ endpoint: staleEndpoint })` after unsubscribing a stale subscription, so the old endpoint is pruned from `push_subscriptions` and no longer receives failed sends.
2. **[Minor] Filtered running-session matches hidden in a collapsed section** — the pinned "In progress" section now stays open while searching (`isFiltering` overrides the collapsed state), matching the existing group behavior.
3. **[Minor] Auto-expand guard stale across active/inactive transitions** — the guard key is cleared when the selected session is not found in any directory group (e.g. it moved to the pinned section), so it auto-expands again when the session transitions back into a group.

Tests: updated `SessionList.directory-action.test.tsx` for the pinned-section behavior (2 rewritten + 1 new case for collapsed search results). Full web suite passes: 208 files / 1789 tests.

wu736139669:
Addressed the two new findings:

1. **[Major] VAPID rotation commits replacement state before hub registration** — `writeStoredVapidKey` now runs only after `subscribePushNotifications` succeeds. A failed hub registration leaves the previous key in place, so the next load retries the replacement instead of reusing a subscription the hub never learned about.
2. **[Minor] Running-section toggle is pointer-only / stale state while filtering** — the pinned section header now has `role="button"`, `tabIndex`, `aria-expanded` and Enter/Space activation; the chevron and `aria-expanded` reflect the effective state (open while searching even when collapsed).

Tests: added keyboard-activation coverage and `aria-expanded` assertions in `SessionList.directory-action.test.tsx`; hardened the New Session custom-model fetch against partial api clients (caught a sync-throw that broke the suite). Full web suite: 1790 tests pass.

wu736139669:
Closing this combined PR in favor of focused PRs per concern:

- **#1315** — feat(web): pinned in-progress session section with state badges (session list UX)
- **#1316** — fix(web): re-subscribe push subscriptions when VAPID keys change
- **#1317** — fix(web): hide the voice button when no voice backend is configured
- **#1318** — feat(hub,web): support custom Claude models via settings.json

All four are rebased on the current upstream main.

PR #1315, opened 2026-08-02
feat(web): pin running sessions in an in-progress section with state badges

wu736139669:
Pinned "In progress" section for the session list, surfacing live sessions with distinct states.

- Active sessions are grouped at the top (before directory groups), sorted by recency, showing project + machine labels.
- Rows are split into **Running** (thinking / background tasks, green pulse), **Pending** (awaiting permission/input, amber) and **Idle** (alive but inactive, gray) sub-groups with matching row badges.
- The section is collapsible (chevron, Enter/Space activation, `aria-expanded`) and stays open while searching.
- Auto-expand of directory groups now only triggers for sessions actually rendered inside a group (opening a pinned running session no longer expands the group below).

Tests: updated `SessionList.directory-action.test.tsx` (pinned-section behavior, collapsed search results, keyboard activation). Full web suite passes.


heavygee:
The problem with this change, is that if you have a lot of sessions open, across many projects, it now becomes harder to differentiate between them. 

Before - you could see "at a glance" where an active agent belonged because it was included in their working directory set.

Now - you have to review the small "working/directory" texts that live below the session titles, to understand what the work is connected to. it's now more work than it used to be :/

I'm struggling to see how the existing "only show active sessions" didn't already suffice - with that enabled, you would get a likely small collection of active sessions, grouped by their working directory.

I think I will try adding an option next to that existing "only show active session" that will turn this from a default, to an optional toggle.

heavygee:
Follow-up on the Idle residual from this PR: opened https://github.com/tiann/hapi/issues/1362

Appreciate the pinned In progress + working/pending work here - that part is solid. The ask in #1362 is narrowly to stop labeling quiet active rows as **Idle** (defaults unmarked; exceptions marked), and to keep quiet actives in directory groups when pin-in-progress is on instead of an Idle subgroup.

PR #1318, opened 2026-08-02
feat(hub,web): support custom Claude models via settings.json

wu736139669:
Claude Code has no model catalog API like Codex, so sessions routed through a custom `ANTHROPIC_BASE_URL` (e.g. DeepSeek) could not select their models in the New Session picker.

- `settings.json` gains an optional `customClaudeModels: string[]` field.
- New hub route `GET /api/claude/custom-models` exposes the list.
- The New Session Claude model picker merges the configured names with the built-in presets; the fetch is guarded so partial API clients (older hubs, tests) don't break the form.

Tested against a hub routing Claude to `api.deepseek.com/anthropic` with `deepseek-v4-flash[1m]`.

### DIFF
--- a/docs/public/schemas/settings.schema.json
+++ b/docs/public/schemas/settings.schema.json
@@ -46,6 +46,15 @@
       },
       "description": "Allowed CORS origins. ENV: CORS_ORIGINS (comma-separated)"
     },
+    "customClaudeModels": {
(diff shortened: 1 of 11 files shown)


wu736139669:
Closing because this customization is no longer needed.

PR #1358, opened 2026-08-04
fix(web): prevent near-bottom chat scroll jump

teekalpha:
## Summary
- resume tail-following only after the viewport actually reaches the bottom
- prevent downward readers from being snapped past the final content
- add regression coverage for downward scrolling near the bottom

## Tests
- `cd web && bun run test -- HappyThread.test.tsx`
- `cd web && bun run typecheck`


teekalpha:
Addressed the review concern in 587aba43:

- latch explicit jump-to-tail intent across intermediate native smooth-scroll events
- cancel the latch when the user reverses upward
- clear it once the viewport reaches the actual bottom
- add a component regression test covering jump-to-tail, intermediate downward scroll events, and scrollHeight growth

Validated with `HappyThread.test.tsx`, `HappyThread.mobile-scroll.test.tsx` (26 tests), and web typecheck.

PR #1365, opened 2026-08-04
feat(pi): import and reconcile local sessions

KorenKrita:
## Summary

Add production-grade import and reconciliation for local Pi coding-agent sessions.

- scan local Pi JSONL sessions through the owning machine RPC
- import the active branch, including messages, tool activity, visible custom messages, compaction summaries, and image placeholders
- resume an imported session through HAPI's existing archived-session reopen transaction
- incrementally reconcile history written outside HAPI without duplicating entries
- expose both New Session import and an inactive-session **Sync Pi history** action

## Motivation

A Pi session may be used directly in Pi and later continued through HAPI, or used in HAPI and then updated elsewhere. Previously HAPI could resume a known native Pi ID, but it could not discover local sessions, hydrate their existing transcript, or manually reconcile history added outside HAPI.

### Machine-side discovery

- honor `PI_CODING_AGENT_SESSION_DIR`, `PI_CODING_AGENT_DIR/sessions`, and the default `~/.pi/agent/sessions`
- keep file access on the owning machine instead of assuming the Hub can read client storage
- resolve the current active leaf from Pi's `id` / `parentId` tree
- tolerate malformed and unknown JSONL records
- preserve stable Pi entry IDs in the RPC response

### Hub import and reconciliation

- identify imports by namespace + machine ID + native Pi session ID
- use stable message local IDs for idempotent retries
- persist imported messages as already invoked, preserving source timestamps
- rebuild HAPI-to-Pi user-entry locators so imported history supports native fork/rewind controls
- persist the latest observed native entry as the incremental cursor
- detect rewritten/diverged branches instead of silently merging incompatible history
- reject insertion into an active HAPI session
- retain partial imports with explicit failed/diverged state for safe retries
- scope Pi deduplication by machine so copied native IDs on different machines do not merge

### Web UX

- add a searchable, directory-filtered Pi session picker to New Session
- support single-session import + reopen and batch import
- refresh successful batch items even when another item fails
- add **Sync Pi history** to the session `⋮` menu for inactive Pi sessions
- report updated, unchanged, active-session, and diverged outcomes
- warn users to close standalone Pi processes before continuing the same native session because Pi does not expose a cross-process ownership lock

## Validation

- `bun run test`
  - CLI: 2,161 passed, 1 skipped
  - Hub: 918 passed, 3 skipped
  - Web: 2,124 passed
  - Shared: 197 passed
- `bun run typecheck:cli`
- `bun run typecheck:hub`
- `bun run build:web` including PWA inject-manifest generation
- two independent read-only reviews, followed by fixes for incremental cursors, oversized-message canonicalization, locator recovery, reopen routing, visible Pi metadata records, partial batch results, and image-only user entries

## Known limitation

Pi currently has no cross-process session ownership lock. Multiple processes can open the same native session, but a stale process may append from an older leaf and create an unintended parallel branch. The UI advises users to reload or close other Pi processes before continuing in HAPI when they want a single linear history.

## Review follow-up

The HAPI Bot findings are addressed in follow-up commits:

- native Pi resumes now require the recorded machine instead of hostname fallback
- batched `get_entries` synchronization persists one final metadata snapshot
- Pi discovery reads bounded headers first and materializes only the newest requested sessions

This PR also repairs the existing `RawSendError.deliveryMode` typecheck regression so the full `bun run typecheck` command passes.


KorenKrita:
@tiann ready for review

PR #1385, opened 2026-08-05
feat(web): add setting to keep reasoning collapsed by default

KorenKrita:
## Summary

Reasoning messages are currently auto-expanded while streaming and stay expanded afterwards. For chat-heavy usage this pushes the answer off-screen on every turn, especially on mobile.

This PR adds a **"Collapse reasoning by default"** setting (Settings → Chat → Tool cards) that keeps reasoning blocks collapsed at all times — including during streaming. The existing auto-expand behavior remains the default for everyone else; the setting is opt-in.

## Changes

- `web/src/hooks/useReasoningCollapse.ts` — new localStorage-backed preference hook, modeled after the existing `useCodexExplorationCollapse`
- `web/src/components/assistant-ui/reasoning.tsx` — `ReasoningGroup` only auto-expands while streaming when the preference is off
- `web/src/routes/settings/chat.tsx` — new toggle in the Tool cards section
- `web/src/lib/locales/{en,zh-CN}.ts` — labels for the new setting
- `web/src/hooks/useReasoningCollapse.test.ts` — tests for default, persistence, and cross-tab sync

## Verification

- `vitest run`: 240 files / 2139 tests passed
- `tsc --noEmit` clean

### DIFF
--- a/web/src/components/assistant-ui/reasoning.test.tsx
+++ b/web/src/components/assistant-ui/reasoning.test.tsx
@@ -1,0 +1,101 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest'
+import { render, fireEvent, act, cleanup } from '@testing-library/react'
+import React from 'react'
+
+// ReasoningGroup consumes `useMessage` from assistant-ui. Mock it so the
+// message status/content can be controlled per test.
+const { mockMessage } = vi.hoisted(() => ({
+    mockMessage: {
+        status: null as { type: string } | null,
+        content: [] as { type: string }[],
+    },
+}))
+
+vi.mock('@assistant-ui/react', () => ({
+    useMessage: () => mockMessage,
+}))
+
+import { ReasoningGroup } from './reasoning'
+
+const STORAGE_KEY = 'hapi-reasoning-collapsed'
+
+function renderGroup() {
+    return render(
+        <ReasoningGroup>
+            <div data-testid="reasoning-content">thinking text</div>
+        </ReasoningGroup>
+    )
+}
+
+// The collapsible region is the direct div child of .aui-reasoning-group
+// (the header is a button). Collapsed state is signalled by the max-h-0 class.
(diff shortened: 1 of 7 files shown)


KorenKrita:
Thanks for the review. Fixed: the effect now re-evaluates the preference during an active stream, collapsing the block when the setting is enabled from another tab. Verified with the full web suite (2139 tests) and `tsc --noEmit`.

KorenKrita:
Addressed the listener fan-out. Instead of lifting the hook into chat context, the preference hook now uses a module-level singleton store (`useSyncExternalStore`): a single `storage` listener is registered for the whole app, and every mounted `ReasoningGroup` shares the same snapshot. Consumers are unchanged.

Added a test that mounts two hook instances and asserts exactly one storage listener is registered, and zero after all unmount.

Full web suite: 2140 tests passed, `tsc --noEmit` clean.

KorenKrita:
Fixed: the setter now updates the in-memory snapshot directly and treats persistence as best-effort, so the preference stays functional even when `localStorage` rejects a write (unavailable or quota-limited).

Added the requested test coverage:
- hook tests: `setItem`/`removeItem` throwing still keeps the preference working in memory
- new `ReasoningGroup` component tests: default collapsed, expand on click, auto-expand while streaming, stays collapsed while streaming when the preference is on, and collapsing an already-expanded streaming block when the preference is enabled from another tab

Full web suite: 2146 tests passed, `tsc --noEmit` clean.

KorenKrita:
Fixed: the store now tracks `hasUnpersistedOverride`. When persistence fails the setter records the divergence and new subscriptions skip the storage resync; a real cross-tab storage event clears the flag and resyncs. The in-memory preference now survives subsequent card mounts.

Added the requested regression test: with `setItem` failing, a second hook instance mounts and both instances retain the in-memory value.

Full web suite: 2147 tests passed, `tsc --noEmit` clean.

PR #1392, opened 2026-08-06
feat(settings): onboard hub provider credentials for dictation and voice

### LINKED ISSUE #1384: Settings: onboard dictation (and voice) provider credentials in UI
## Rationale

Env-only / SSH `hub.env` configuration for dictation (and voice/agent) provider keys is a **shipping shortcut, not a design win**. HAPI is a local-first, single-operator hub: there is no advantage to requiring file edits and a hub restart before Settings can list providers, and no meaningful disadvantage to onboarding those credentials in Settings — as long as keys stay hub-side (authenticated write, masked read; env remains an optional override).

The hub already persists other config via `env > settings.json > default`. Voice/dictation keys being env-only is inconsistent with that model and makes the feature invisible (empty picker: "No transcription provider is configured on the hub").

**Fix:** Settings UI to add/edit/clear hub-stored provider credentials so providers appear without SSH.

Appreciate #1327 / #1329 for the transport; this is the missing first-run / onboarding layer, not a rejection of env-based ops for power users.

## Problem

Dictation landed in #1327 / #1329 (closes #462) with the correct security posture: provider API keys stay in the **hub process environment**, and the browser only talks to authenticated hub routes.

That leaves a first-run hole. Settings → Voice → Dictation shows:

> No transcription provider is configured on the hub.

There is **no** Settings field to paste an OpenAI / ElevenLabs / Deepgram / Groq / OpenAI-compatible key (or base URL + model). Docs tell operators to `export …` / edit systemd `EnvironmentFile` / hub env and **restart**. To a normal user the feature looks broken or missing.

The same invisible pattern applies to voice-assistant backends (`ELEVENLABS_API_KEY`, `GEMINI_API_KEY` / `GOOGLE_API_KEY`, `DASHSCOPE_API_KEY` / `QWEN_API_KEY`) and more broadly to agent provider credentials: pickers only list what the process already has.

## Expected

1. Authenticated Settings UI to **add / edit / clear** hub-side provider credentials (dictation first).
2. Keys stay on the hub: never returned in full to the client; UI shows masked hints only.
3. After save, providers appear in the dictation picker **without** SSH-editing env files. Prefer live reload via in-process env merge; document if a restart is still required for a given key.
4. Empty state teaches the path ("Add a provider…") instead of only "not configured on the hub."
5. Env vars remain a valid ops/bootstrap override (env wins over UI-persisted settings).

## Proposed shape

- Hub: authenticated `GET`/`PUT` credentials API under `/api/voice/transcription/credentials`.
- Persist UI-managed secrets in hub data dir (`settings.json` `providerCredentials`, same trust model as existing `telegramBotToken`).
- On load + on save: apply file-backed values into the effective env view used by `listConfiguredTranscriptionProviders` / `listConfiguredVoiceBackends` when the process env key is unset.
- Web: Settings → Voice dictation onboard form; share the same credential store shape for assistant backends (and later agent providers).

## Non-goals

- Changing the transport model from #1327/#1329 (browser must not hold long-lived provider API keys).
- Replacing env-based ops for power users.

## Refs

- #462 (dictation modes, closed)
- #1327 Add provider-backed dictation mode
- #1329 Add realtime dictation providers

heavygee:
## Summary
- Closes #1384: Settings → Voice can add / edit / clear hub-side credentials for dictation providers and voice-assistant backends without SSH-editing `hub.env`.
- Hub persists `providerCredentials` in `settings.json` (masked reads; authenticated writes; process env still wins when set).
- After save, pickers refresh so providers appear without a restart for UI-managed keys.

## Motivation
Env-only configuration for these keys was a shipping shortcut, not a design win. HAPI is local-first / single-operator; empty "No transcription provider is configured on the hub" with no UI path made the feature look broken. Env remains a valid ops override.

## Test plan
- [x] `bun typecheck`
- [x] Focused hub/web tests (`providerCredentials`, voice routes, client, voice settings)
- [x] Operator dogfood: Settings onboard → provider appears in picker
- [ ] Review: env-set keys show as locked / non-editable and win over settings.json
- [ ] Review: clearing a UI credential removes the provider from the picker
- [ ] Review: no full secrets returned on GET credentials


heavygee:
Addressed the three Major findings in 90912246c:

1. **Alias env locks** — `isLogicallyEnvLocked` treats GEMINI/GOOGLE and DASHSCOPE/QWEN as one lock so a settings alias cannot shadow the env sibling on apply.
2. **Save ≠ Clear** — empty fields send `undefined` (omit); only Clear sends `null`. OpenAI-compatible base URL/model are seeded from status so Save does not wipe them.
3. **Owner-only settings.json** — `writeSettings` now uses `0o600` file + `0o700` dir with post-write `chmod` (umask-safe).

Added regression tests for alias shadowing, omit-vs-clear, and POSIX mode `0o600`.

heavygee:
Second-pass Majors addressed:

1. **Owner namespace** — GET/PUT `/credentials` return 403 unless `ns === 'default'`.
2. **Atomic persist** — patches stage into a copy; `process.env` syncs only after `writeSettings` succeeds.
3. **Independent OpenAI-compatible editability** — base URL / model / API key use separate disabled flags; Clear only nulls editable fields.

Tests cover non-owner 403, locked multi-field no env mutation, write failure no env mutation, and mixed-lock editability flags.

heavygee:
Third-pass Majors addressed:

1. **Partial OpenAI-compatible Clear** — Clear appears when any editable field has a stored value (URL / model / API key), not only when both URL+model make `configured` true.
2. **Serialized settings RMW** — `withSettingsLock` chains concurrent updates per settings file so disjoint PUTs cannot clobber each other or race on `.tmp`.

Tests cover concurrent openai+groq updates and API-key-only partial status.

heavygee:
Fourth-pass findings addressed:

1. **Serialize every settings writer** — added `updateSettings` + locked `writeSettings`; migrated credentials, relay auth RMW, `getOrCreateSettingsValue`, `loadServerSettings`, and CLI token persist. Race test covers credential PUT vs relay-key write.
2. **Clear resets compatible fields** — after Clear/Save, base URL / model / API key inputs resync from the returned status so partial clears cannot be re-saved from stale React state.

heavygee:
Fifth-pass Majors addressed:

1. **Cross-process lock** — shared `@hapi/protocol/settingsFileLock` (`wx` + stale retry, same constants as the old CLI lock). Hub `withSettingsLock` = in-process chain + file lock; CLI `writeSettings`/`updateSettings` use the same helper.
2. **Owner-only after CLI rewrite** — CLI atomic writes now use `0o600` + post-`chmod` (and `0o700` home dir).

Integration test races hub credential PUT against a CLI-style lock writer and asserts both fields survive with mode `0600`.

heavygee:
Sixth-pass Majors addressed in ee9e5b57f:

1. **UI secrets stay out of `process.env`** — settings-backed credentials live in an in-memory overlay via `getProviderEnvironment()`; voice/dictation paths read that overlay. Tunnel/ACP/Codex children no longer inherit Settings-managed API keys.
2. **PID + token settings locks** — `withSettingsFileLock` records `{pid,token}`; reclaim only when the recorded PID is dead (or the sidecar is legacy/empty); release unlinks only when the token still matches ours. EPERM on `kill(pid,0)` counts as alive.

Tests cover overlay vs `process.env`, dead-PID reclaim, legacy empty locks, and successor-safe release.

heavygee:
Blocker addressed in the tip commit:

**Do not reclaim ownerless lock sidecars** — acquisition now uses sync `openSync('wx')` + `writeSync` for the owner payload, and on `EEXIST` only unlinks when `readLockOwner()` returns a parsed owner whose PID is dead. A transient empty/partial file (publication window) is waited out, not stolen. That was the CI `ENOENT` on `settings.json.tmp` rename in the hub-vs-CLI race test.

Coverage: empty-sidecar fail-closed, dead-PID reclaim, successor-safe release, 8-way concurrent writers.

heavygee:
Seventh-pass Majors addressed:

1. **Serialize stale reclaim** — dead owners are moved aside with `renameSync` to a unique `.break.*` path, re-verified (pid+token, still dead), then unlinked. Contenders that lose the rename never touch the live lock path. Owner read during reclaim is synchronous so waiters cannot all snapshot one dead owner across an `await` and stampede the exclusive create.
2. **Failed publication cleanup** — if writing the owner payload fails after `wx`, the sidecar is unlinked so settings writers are not wedged on an empty lock.

Tests cover write-failure cleanup, multi-contender dead-owner reclaim (max concurrency 1), and the existing empty-sidecar / successor-release cases.

heavygee:
Eighth-pass findings addressed:

1. **Reaper-serialized stale reclaim** — dead owners are only unlinked while holding a fixed `${lock}.reap` exclusive sidecar, after re-validating pid+token. Delayed contenders that still remember the old dead owner cannot rename/unlink a successor's live lock. Added a barrier test for that interleaving.
2. **Minor schema** — `providerCredentials` added to `docs/public/schemas/settings.schema.json`.

heavygee:
Ninth-pass Majors addressed:

1. **CLI fail-closed read** — locked `updateSettings` uses `readSettingsForUpdate()` which throws on unreadable/invalid JSON instead of synthesizing `{}` and wiping hub fields. Corrupt-file test asserts bytes unchanged.
2. **Reaper backoff** — when `${lock}.reap` is held, reclaim returns false and the waiter sleeps before retrying instead of spinning attempts synchronously.

heavygee:
Tenth-pass Majors addressed:

1. **Crash-safe lock publish** — acquisition writes a fully-formed `.candidate` then `linkSync`s onto the fixed lock path. A kill mid-publish leaves an orphan candidate at worst, not an empty live sidecar. Orphan-candidate test included.
2. **Vitest hoist** — CLI persistence test creates its temp dir inside `vi.hoisted` so the required test job can import the suite.

heavygee:
Eleventh-pass findings addressed by taking Codex's escape hatch:

1. **proper-lockfile** — replaced the bespoke `wx`/reaper sidecars with `proper-lockfile` (mkdir + mtime lease, `stale: 30s`, retries). Crash-recoverable without an unreclaimable empty file. Lock path is `settings.json.hapi.lock` so leftover file-shaped `*.lock` artifacts from earlier revisions cannot block mkdir.
2. **Minor** — voice settings credentials editor/link only render for the owner namespace (`getNamespaceFromToken` shared with SettingsNav).

heavygee:
Eleventh-pass findings addressed by taking Codex's escape hatch (fe71c2f50):

1. **proper-lockfile** — replaced the bespoke `wx`/reaper sidecars with `proper-lockfile` (mkdir + mtime lease, `stale: 30s`, retries). Crash-recoverable without an unreclaimable empty file. Lock path is `settings.json.hapi.lock` so leftover file-shaped `*.lock` artifacts from earlier revisions cannot block mkdir.
2. **Minor** — voice settings credentials editor/link only render for the owner namespace (`getNamespaceFromToken` shared with SettingsNav).

heavygee:
CI kick: no check-suites on tip after rebase (GitHub Actions lag/outage). Reopening to retrigger workflows.

PR #1405, opened 2026-08-07
feat(media): add audio and file display

NightWatcher314:
## Summary

This is a stacked follow-up to #958. Please review the final follow-up commit (`f57783e7`) for the changes introduced here; the earlier commits are inherited from #958.

- add a permission-gated `display_media` MCP tool while keeping `display_image` and `display_video` compatibility
- render generated audio with native controls and unknown files as downloads
- detect WAV, FLAC, Ogg, MP3, and M4A from file bytes; distinguish audio-only and video ISO-BMFF files through a bounded box walker
- lazy-load non-image media and keep the existing 25 MiB per-file/cache limits
- serve generic files with `Content-Disposition: attachment` and all generated media with `X-Content-Type-Options: nosniff`
- expose the tool consistently through Claude, Codex, Cursor, OpenCode, ACP, prompts, doctor output, and the shell helper

## Test plan

- `bun typecheck`
- `bun run test` — CLI 216 files / 2284 passed / 1 skipped; Hub, Web, and Shared passed
- `bun run build`
- targeted generated-media, ACP bridge, web renderer, and Hub route tests
- `git diff --check`
- two independent local reviews; final result: `Findings: None`

## Notes

- Depends on #958 and should be rebased after it lands.
- Generated with substantial assistance from OpenAI Codex; changes were manually inspected, tested, and independently reviewed.


heavygee:
@NightWatcher314 🤜 💥 🤛 - nicely done!

PR #1436, opened 2026-08-09
feat(web): persist and send voice input message across session navigation (#1435)

### LINKED ISSUE #1435: feat(web): persist and send voice input message across session navigation
### Summary
When a user inputs speech/voice in the web frontend and hits the send button, if the user switches to another session while the voice input is being processed or sent, the message processing and dispatching should continue bound to the target session where it was initiated.

### Desired Behavior
1. User speaks/inputs voice on Session A.
2. User presses the send button.
3. User navigates away to Session B (or another page).
4. The voice input processing / transcription / send pipeline completes in the background and correctly delivers the message to Session A.

swear01:
Closes #1435

## Summary

Finish an explicitly requested voice send even when the operator navigates to another session while transcription or delivery is pending.

- Bind delivery to the original target; resume inactive sessions before posting and avoid navigating away from a different page after background completion.
- Preserve standard and realtime finalization across unmount; canceled recordings without a pending send are not uploaded.
- Preserve source and destination follow-up drafts on success and failure. Resolve successful transfer text at persistence time so attachment transfer does not freeze an old destination draft.
- Realtime failure recovery replaces the merged composer text rather than appending it twice.
- Resolve the currently mounted composer by session ID during both successful transfer and late recovery; sample live edits and update its text/ref before a subsequent unmount can overwrite the merged draft. Both standard and realtime sends reuse the same success transfer helper.
- Preserve rich-input flushing and existing attachment, scheduling and delivery-mode guards.

## Maintenance verification — 2026-09-13

Integrated main `32ca6f54` without history rewrite. Latest head: `0a7bbc3b81d44050a53c8aca788fd074b12f5c57`.

- `bun typecheck` and full CLI/Hub/Web/Shared/Relay suites: passed.
- Latest focused voice/draft-transfer/hydration suite: 94 passed.
- Registered composers expose saved text until their animation-frame hydration applies it, then respect intentional empty edits. Both voice modes cover a failed background POST before destination hydration, followed by hydration and unmount, without losing destination text.
- Failed cross-session sends preserve the independent destination draft even when it existed before resume resolved, and merge source follow-ups separately instead of treating destination text as submitted content. Both modes retain regressions for pre-resolution, in-flight, mounted and unmounted recovery.
- Same-ID resumes keep the original failure-recovery baseline. Voice transfers preserve destination attachment drafts and their valid upload metadata, re-sample target edits during corrective writes, and refuse a destructive move after a target storage-read error. Standard/realtime regressions remount the destination and restore the saved attachment.
- Regression failures were reproduced before fixes for background destination drafts, production realtime replacement callbacks, and two successive composer mounts followed by a late rejection and another unmount.
- Playwright: 16 passed.
- Isolated runner integration on the initial integration head: 13 passed, 1 skipped; subsequent changes affect frontend recovery only. Latest-head GitHub CI and review remain authoritative.

Local verification uses bounded workers, independent temporary directories, Node experimental webstorage disabled, and a test-only macOS ps adapter. No live transcription-provider verification is claimed.

No upstream merge, release, or fleet deployment was performed.

AI disclosure: Codex assisted with integration, regression tests and verification.


swear01:
### Regression: standard dictation drops microphone processing constraints

While investigating a mobile browser/PWA report where standard dictation
produces poor recognition and only a small fraction of the spoken text, I
found that this PR changes:

```ts
navigator.mediaDevices.getUserMedia({
    audio: {
        echoCancellation: true,
        noiseSuppression: true,
        autoGainControl: true
    }
})
```

to:

```ts
navigator.mediaDevices.getUserMedia({ audio: true })
```

This change appears unrelated to the session-navigation behavior addressed by
this PR. It also makes standard dictation inconsistent with the realtime
transcription paths, which still request all three microphone constraints.

The maintained `v0.29.0.3` PWA carries this PR head and exposes the issue on
mobile. The code regression is confirmed; whether it fully explains the
reported recognition degradation still needs a Standard/Realtime A/B test.

Could the original microphone constraints be preserved and a regression test
added to assert the `getUserMedia` arguments?

PR #1451, opened 2026-08-09
feat(web): configure Create agent visibility

swear01:
Adds display settings to choose which agent buttons appear in Create Session. Defaults to showing all creatable agents.


techotaku39:
Hi @swear01, thanks for working on this.

After testing this PR with the latest `upstream/main`, I found the per-agent visibility switches a little too granular. The bigger product issue seems to be discoverability: the current main branch hides Agents that are not installed or available on the selected machine, so users may not notice newly supported Agents.

Would you consider changing this PR to use a single browser-local Display setting, such as **“Show unavailable Agents”**, disabled by default?

- Disabled: keep the current behavior and show only available Agents.
- Enabled: show all creatable Agents known by the current HAPI build; unavailable ones remain disabled/muted, with the reason shown where possible.
- Keep the Runner-side availability check authoritative, so this setting only changes presentation.

This seems simpler for users while preserving a clean default UI. Since this overlaps with the current per-agent visibility controls, would it be better to adapt this PR or handle it as a follow-up? If you agree with this direction, I’d also be happy to prepare a PR implementing the single-toggle behavior.

swear01:
Closing this superseded PR: current NewSession lists runner-reported available/installed agents only, so a separate visibility preference is not needed.

Issue #1374, opened 2026-08-04
feat(a2a): P1 work-graph ledger (events + principal/namespace)


heavygee:
## Summary

First implementation slice of the A2A control-plane RFC ([Discussion #1332](https://github.com/tiann/hapi/discussions/1332)).

**In scope (P1 only):**
- `events` + `event_links` tables (RFC shape)
- Structured `principal` + `namespace` on every write
- Write/query API (hub-internal store + minimal HTTP if needed for tests)
- Namespace isolation tests
- Hub capability flag (additive-only wire; old clients neither offer nor break on Layer 1)
- Refuse writes whose principal is non-human with no resolvable human owner

**Explicitly out of scope:**
- Handoff create/deliver/receipt (P2)
- Work-ad / `AGENT_NOTIFY_SUMMARY` elevation (P3)
- Manager UI / Overseer / inbox / attention flags
- `granting_event_id` validation (grant shape still open in the RFC — parse/hold, do not accumulate unvalidated pointers)

## Why this first

Layer 0 (`ping_peer` / `inspect_peer` / citations) already moves messages between sessions. P1 does not add a new collaboration verb — it adds a durable, attributable place to put structured cross-session truth so P2 handoffs are not amnesia.

## Acceptance (from RFC)

- [ ] Non-human principal with no human owner is refused
- [ ] Cross-namespace read/write is refused (isolation tests)
- [ ] Idempotent insert via `idempotency_key` does not duplicate
- [ ] Old client without capability flag neither offers nor breaks on Layer 1
- [ ] Query by `related_session_id` returns that session's ledger rows

## References

- Discussion: https://github.com/tiann/hapi/discussions/1332
- Parent strategy: https://github.com/tiann/hapi/discussions/1258
- Related Layer 0: #1195, #1228, #1370, #1371

tiann:
Closing the original P1 scope as completed by #1467, which also landed the ledger substrate: events/event_links, structured principal and namespace, idempotent writes, namespace-isolation coverage, session-scoped queries, rejection of non-human principals without a human owner, and the health capability flag. Further project indexing or later A2A phases are separate follow-up work, not blockers for this original P1 slice.

PR #1462, opened 2026-08-09
feat(web): render AGENT_NOTIFY_SUMMARY as compact metadata

techotaku39:
## Summary

- Split valid trailing `AGENT_NOTIFY_SUMMARY` footers from assistant prose at render time.
- Render status, summary, and action as a compact Hapi metadata row instead of raw JSON.
- Use a checkmark for completed summaries, localized status labels for other states, and aligned status/action markers.
- Preserve invalid or non-final footer-like text as normal assistant content.

## Scope

This is a presentation-only change. It does not change the `AGENT_NOTIFY_SUMMARY` contract, settings toggle, stored message content, or notification payloads.

## Testing

- `bun test shared/src/messages.test.ts`
- `bun run --cwd web test -- src/components/AssistantChat/messages/NotifySummaryText.test.tsx`
- `Invoke-HapiTaskPlaywright.ps1` terminal wrapping suite: 2/2 passed
- Live desktop and mobile smoke checks with historical `AGENT_NOTIFY_SUMMARY` fixtures: 2/2 passed


heavygee:
Thanks for this — the compact metadata row is genuinely good UX.

Also: sorry we shipped #1376 (opt-in emit) without the corresponding default chat removal on the same train. That left vanilla users staring at a raw machine footer when the toggle is on — your PR is exactly the kind of "wait, this is ugly" reaction that gap deserved 😂

Product context, tying into the A2A control-plane thread ([#1332](https://github.com/tiann/hapi/discussions/1332)):

`AGENT_NOTIFY_SUMMARY` started as FCM enrichment (#803). #1376 added an optional Settings toggle so agents can **emit** that footer. The A2A RFC treats the line as a **machine status contract** (Layer 0→1): best-effort worker self-report that should land in the work-graph ledger (P1 [#1374](https://github.com/tiann/hapi/issues/1374), then P3 notify→ledger ingest — separate PR), not as always-on chat chrome.

So the continuum looks like:

1. **Emit** — #1376 (shipped, opt-in today)
2. **Capture** — P3 ingest into `events` / work-graph ([#1465](https://github.com/tiann/hapi/issues/1465) / [#1467](https://github.com/tiann/hapi/pull/1467); includes A0 substrate aligned with #1374 contracts)
3. **Display** — by default **hide** from chat (render-time strip; store keeps the raw line for FCM/parse/ledger); optional **show** = exactly your compact row

Only once (2) and (3) both exist can we responsibly flip emit to **always on, optionally disable**. Until then, people who hate the raw line have no clean escape except turning emit off — and disabling emit also turns off the status feed that upcoming A2A abilities (#1332 P2+ / work-ads / stale detection) will consume. Display control lets chat stay clean (or nice) without starving the ledger.

Your presentation work is the right shape for the **show** path. If it lands alone as always-on, every enabled-emit user gets a chat affordance with no opt-out, while the durable ledger path was incomplete until #1467.

### Ask

Could you extend this PR (or a small follow-up on the same branch) with a Settings sibling to #1376, roughly:

- **Show session status summary in chat** (name bikeshed-ok)
- **Default: off** (hide / strip from chat + copy)
- **When on:** render your compact `NotifySummaryText` row (not raw JSON)
- **Invariant:** display never gates parse/FCM; display must not gate capture (#1467)

Happy to take strip/default-off ourselves via #1464 if you would rather keep this PR presentation-only — either way we want your compact row as the **on** renderer, not a competing always-on design.

Refs: #1332, #1374, #1376, #1464, #1465, #1467.

techotaku39:
Thanks for the detailed product context — I agree that emit, capture, and display should remain separate concerns.

I’d prefer to keep #1462 presentation-only: it provides the compact `NotifySummaryText` renderer for the eventual “show” path, without adding another user setting here.

I’m happy for #1464 to own the default-off display gate, including hiding it from chat and copy while preserving the raw footer for parsing, FCM, and future ledger capture. To avoid shipping an always-on row, #1464 should land before or alongside #1462, or stack on top of this branch.

The current compact row can then serve as the enabled “show” renderer.

heavygee:
Thanks — taking the default-off display gate in #1464 via https://github.com/tiann/hapi/pull/1477 (stacks your `NotifySummaryText` as the enabled show renderer). Merge train with capture #1467; will coordinate so we do not ship always-on compact row alone.

techotaku39:
Closing this PR because its implementation and follow-up fixes were incorporated into #1477, which has now been merged. Thanks for carrying the compact `NotifySummaryText` renderer forward.

techotaku39:
Closing this PR because its implementation and follow-up fixes were incorporated into #1477, which has now been merged.

PR #1497, opened 2026-08-11
fix(web): persist directory expansion locally

TimChesko:
## Summary
- store file-manager expanded folders in localStorage so directory state survives closing/reopening the Mini App
- keep the existing sessionStorage key as a legacy fallback and migrate it on write
- add regression coverage for remount restore and legacy sessionStorage migration

## Test plan
- cd web && bun run test src/components/SessionFiles/DirectoryTree.test.tsx
- cd web && bun run typecheck


TimChesko:
Addressed the storage fallback review in 08533ebb. Reads now try the session fallback safely before localStorage, so a failed local write cannot be shadowed by stale local state and a throwing local getter no longer prevents fallback restore.\n\nRan:\n- `cd web && bun run test src/components/SessionFiles/DirectoryTree.test.tsx`\n- `cd web && bun run typecheck`

PR #1512, opened 2026-08-11
fix(web): use latest assistant replies for session recency

### LINKED ISSUE #477: feat(web): improve session sidebar with pinning, sorting, and unread indicators
## Feature Request

The session list sidebar in the web UI currently has limited management capabilities. Several improvements would make it more practical for users managing multiple sessions.

### 1. Pin / Favorite sessions
- Allow users to pin important sessions to the top of the list
- Pinned sessions should stay at the top regardless of activity

### 2. Customizable sort order
- Currently, sessions auto-sort by most recent activity (the session that receives a new message jumps to the top), which can feel disorienting
- Offer sorting options: by last activity, by creation date, alphabetical, or manual order
- Consider making the auto-reorder behavior optional, as it can be confusing when a session suddenly jumps to the top just because a message was sent

### 3. Unread message indicators
- Show a badge or dot on sessions that have new/unread messages
- Distinguish between "new assistant reply" and "no new activity"
- This is especially important when running multiple sessions in parallel — users need to know which sessions have new responses

### 4. Better activity sorting behavior
- The current "most recently active session goes to top" behavior can be disorienting
- Consider a less aggressive approach: highlight the session with new activity but keep the list order stable, or only reorder on user-initiated actions

## Context

When working with HAPI across multiple sessions simultaneously, the sidebar is the primary navigation tool. These improvements would significantly help users manage and track their active sessions.

techotaku39:
## Summary

Use the latest visible assistant reply as the default session-list recency clock while preserving the existing activity and replay behavior based on `updatedAt` where explicitly required.

This PR also makes legacy reply-clock migration asynchronous and bounded, removes redundant fork transcript rescans, and keeps reply-clock updates ordered across dual SSE connections.

As a follow-up to the recency change, unread state and local read watermarks now use the same reply/activity clock. Lifecycle-only metadata updates, such as archive and reopen bookkeeping, no longer create false unread indicators.

## Problem / Motivation

The sidebar previously used `updatedAt` for both session activity and displayed recency. Prompt or activity updates could make a session look newer than another session with a more recent AI response, making the right-side timestamp and default ordering difficult to interpret.

Legacy transcript backfills also needed to avoid materializing large histories while holding SQLite write locks. Global and session-scoped SSE connections can additionally deliver full session records out of order while a merge or rewind recomputes the reply clock.

After introducing the independent assistant-reply clock, unread detection and local read watermarks still relied on raw `updatedAt`. Archive and reopen lifecycle metadata can update `updatedAt` without adding user-visible content, causing an already-read session to show a false white unread dot.

### Reply-clock derivation and persistence

- Add the durable assistant reply clock in schema v27:
  - nullable `last_assistant_message_at`
  - `assistant_reply_clock_backfilled` migration marker
- Detect visible assistant prose while ignoring tool calls/results, reasoning, hidden progress, and metadata-only events.
- Maintain the reply timestamp during message ingest, import, and copy operations.
- Mark merge and rewind transcript edits for asynchronous recomputation instead of decoding the full transcript inside their write transactions.
- Queue legacy transcript backfills after cache hydration, scanning newest-first in bounded 200-row pages and yielding between pages.
- Guard backfill completion with the session sequence so concurrent transcript changes leave the marker unset and are retried safely.
- Stop pending backfill work during SyncEngine shutdown and persist the marker even when no visible reply exists.
- Aggregate the newest assistant reply during fork batch hydration and update the destination session once, without a second full-transcript scan.

### Cache, sync, and SSE consistency

- Emit a narrow versioned reply-clock patch for exact recomputations, so backward/null corrections update only reply-clock fields and cannot overwrite newer metadata/runtime state.
- Carry the session sequence as the reply-clock version through structured patches and list summaries.
- Preserve replies received while legacy backfill is incomplete by lowering the local unread watermark before completion can seed it.
- Gate full records and versioned backward/null updates on both detail and sidebar caches to prevent stale dual-SSE delivery from restoring an old timestamp.
- Refresh the live session cache after creating a direct Codex import, because the import inserts transcript rows without per-message realtime events.
- Propagate reply-clock fields through stored sessions, the session cache, SSE patches, API summaries, and client rendering.

### Client and session-list behavior

- Sort the default session list by the latest visible assistant reply, falling back to `updatedAt` when no visible reply exists.
- Use the same reply/activity fallback for rendered session rows, including imported Codex sessions.
- Preserve global/project pin, active-session, and pending-request priority buckets.
- Keep explicit `order=updatedAt` behavior unchanged.
- Preserve existing consumers that explicitly use `updatedAt`, including replay and activity-order paths.

### Unread-state consistency follow-up

- Use `lastAssistantMessageAt ?? updatedAt` for unread checks, read watermarks, and explicit mark-unread actions across Web, Android, and iOS.
- Add `assistantReplyClockBackfilled` to full session and summary payloads so clients defer legacy unread baselines until reply-clock backfill completes.
- Move Web and native local read-state storage to v2 namespaces/snapshots so old raw-`updatedAt` watermarks are not interpreted as reply-clock watermarks.
- Track legacy sessions skipped during initial baseline creation and seed them once the Hub reports authoritative reply-clock data, without seeding sessions that appear after baseline initialization.
- Prevent lifecycle-only metadata writes, including archive, reopen, and repeated lifecycle stamps, from advancing `updatedAt`.
- Preserve normal activity updates when lifecycle metadata is bundled with a real user-facing metadata change.
- Reuse the Hub's metadata merge semantics when deciding whether sparse lifecycle payloads contain ordinary metadata changes.
- Add regression coverage for archive/reopen lifecycle writes, sparse metadata payloads, repeated lifecycle stamps, legacy baseline initialization, local read state, and native stores.

## User Impact

- Sessions with visible AI responses display and sort by their latest reply time.
- Sessions without visible replies fall back to activity time.
- A newly received AI reply makes the session unread.
- Replies received while legacy reply-clock backfill runs remain unread until opened.
- Archiving or reopening an already-read session does not create a false white unread dot.
- Legacy sessions do not produce historical unread indicators while reply-clock backfill is incomplete.
- Existing databases backfill incrementally after startup instead of blocking Hub construction.
- Rewinds and merges cannot be undone in connected browser caches by older SSE records.
- Web, Android, and iOS use the same reply-time and unread-state semantics.
- Pinning, active/pending prioritization, explicit `order=updatedAt`, and existing replay behavior remain unchanged.

## Migration / Rollback

- Existing databases migrate from the upstream schema to schema v27.
- Existing session rows receive an incomplete reply-clock backfill marker and are scanned asynchronously.
- Newly created sessions opt into the current write-through reply-clock path immediately.
- The additive fields preserve existing `updatedAt` values and do not rewrite historical activity timestamps.
- Web local read-state keys and Android/iOS snapshots use v2 formats because their clock semantics changed.
- No destructive down migration is included; if the code is reverted, the additive columns can remain unused under the repository's normal schema policy.

## Validation

- `bun typecheck` — passed.
- `bun run build` — passed.
- From `shared`: `bun test src/schemas.sessionPatch.test.ts src/sessionSummary.test.ts` — 39 passed across 2 files.
- From `hub`: `bun test src/sync/sessionCache.applySessionPatch.test.ts` — 16 passed.
- From `web`: `bunx vitest run src/hooks/useSSE.test.ts` — 46 passed.
- From `web`: `bunx vitest run src/hooks/useSSE.test.ts src/lib/sessionLastSeen.test.ts src/lib/sessionAttention.test.ts` — 83 passed across 3 files.
- `PLAYWRIGHT_WEB_PORT=3045 bun run test:e2e -- terminal-wrap-fidelity.spec.ts` — 2 passed in the fixture-only Root suite.
- `pwsh -NoProfile -File .\scripts\Invoke-HapiTaskPlaywright.ps1 -Name session-last-reply-time -Suite Root terminal-wrap-fidelity.spec.ts` — 2 passed.
- From `web`: `bunx vitest run src/lib/sessionAttention.test.ts src/lib/sessionLastSeen.test.ts src/components/SessionRowSummary.test.tsx src/components/SessionList.test.ts src/components/SessionRowSummary.time.test.ts src/hooks/useSSE.test.ts src/hooks/queries/useSessions.test.tsx` — 137 passed across 7 files.
- From `hub`: `bun test src/socket/handlers/cli/sessionHandlers.test.ts src/store/sessions.test.ts src/sync/sessionCache.applySessionPatch.test.ts src/store/messages.test.ts src/store/messages.truncate.test.ts` — 111 passed.
- From `shared`: `bun test src/sessionSummary.test.ts` — 29 passed.
- Isolated Full test environment — local and public health/auth checks passed with one active Runner machine.
- Android and iOS platform suites were not run locally because this Windows host has no JDK compiler or Xcode/Swift toolchain.

## Related Issues

Fixes #477

## AI Disclosure

Tool: OpenAI Codex.
Models used across the branch: GPT-5.6 and GPT-5.

The implementation, follow-up fix, diff review, and validation were performed in the HAPI task worktree.


techotaku39:
I found a concrete reproduction on HAPI v0.29.1. I first observed this after upgrading to this release.

A session that has already finished and received its final assistant reply can later:

- change its sidebar timestamp to “just now”, even though no new user message was sent and no new assistant reply appeared;
- become unread again after I opened it and it was marked as read.

The behavior is intermittent. I have screenshots showing completed/inactive sessions being displayed as “just now” with unread dots, and I can provide them if needed.

This appears consistent with the release using the same `updatedAt` clock for both sidebar recency and unread detection, while lifecycle/status metadata updates can advance that clock without adding visible chat content.

Could you confirm whether this case is covered by the current implementation? A regression test for a completed session whose lifecycle metadata changes after its final assistant reply may be useful.

PR #1525, opened 2026-08-12
test(cli): make the suite pass on macOS hosts

7bata:
## What

The `cli` test suite has 11 failures when run on a macOS host. This PR fixes them. **Test files only — no source changes.**

## These are not regressions from my tree

I found these while syncing a fork, so the first thing I checked was whether I had caused them. I hadn't: I reproduced all 11 identically on **pristine `upstream/main` (`fea42212`, v0.27.3)** in an isolated clone on this Mac, with no local changes applied.

Measured on macOS 26.6 (arm64), `bunx vitest run <file>`, one file per run:

| File | Before (pristine main) | After |
|---|---|---|
| `cli/src/agent/__tests__/runAgentPty.test.ts` | 1 failed / 34 passed | 35 passed |
| `cli/src/agy/utils/agyHookCarrier.test.ts` | 4 failed / 20 passed | 24 passed / 2 skipped |
| `cli/src/agy/utils/agyHookCarrierPlatformScope.test.ts` | 1 failed / 17 passed | 18 passed / 1 skipped |
| `cli/src/agy/utils/agyHookCarrierScopeCache.test.ts` | 5 failed / 1 passed | 6 passed |

## Root causes

Two kinds, neither of which is a product bug:

**1. Linux-host assumptions in tests (9 failures).** The agy hook-carrier tests probe `/proc` and expect Linux-shaped scope identities. On darwin the platform dispatch takes a different path, so the assertions cannot hold.

**2. One environment leak (2 symptoms, 1 fix).** `runAgentPty.test.ts`'s spawn-env isolation test inherits `CLAUDE_CONFIG_DIR` from the host shell — which is always set when the suite is launched from inside a Claude Code session. Its assertion failure also abandoned the in-flight `runAgentPty` call on the shared harness, which resurfaced later as the unhandled `testagent PTY exited before becoming ready` rejection. Clearing the inherited vars around that test fixes both symptoms.

## On the skips — please look here first

Three tests are now skipped on macOS, and I want to be explicit that this is **not** failures swept under a rug:

- Gating is `it.runIf(process.platform === 'linux')`, so the Linux-only tests still run on your Linux CI exactly as before.
- **Every Linux-gated test got a darwin twin covering the same behaviour on the other platform.** Net test count goes *up*, not down: 24 → 26 and 18 → 19 in the two affected files.
- One of the added tests is the real-macOS success path (`ioreg` + `sysctl` end to end into `owner.json`) that `agyHookCarrierPlatformScope.test.ts`'s own file header had deferred to "a live mac run". This is that run.

## Verification

Full suite on macOS after this change: cli 3617 passed / 4 skipped, hub 1113, web 2373, shared 253 — exit 0.

I do not have a Linux host to verify on, so the Linux side rests on the gating being a no-op there plus your CI. The Linux-path assertions themselves are unchanged in substance.

## AI disclosure

Per CONTRIBUTING's Vibe Coding policy: this was produced by a multi-agent workflow — Claude Fable 5 for planning and task decomposition, Claude Sonnet for implementation, and Claude Opus for code review and acceptance. A human reviewed the result and ran the verification above.

### DIFF
--- a/cli/src/agent/__tests__/runAgentPty.test.ts
+++ b/cli/src/agent/__tests__/runAgentPty.test.ts
@@ -394,26 +394,44 @@ describe('runAgentPty', () => {
     })
 
     it('injects envVars/extraEnv into the spawn env only (not process.env)', async () => {
-        const msg = deferred<{ message: string } | null>()
-        const opts = makeOpts({
-            envVars: { FLAVOR_TOKEN: 'tok' },
-            extraEnv: { CLAUDE_CONFIG_DIR: '/tmp/iso-cfg' },
-            nextMessage: () => msg.promise,
-        })
-        const promise = runAgentPty(opts)
-        await tick(0)
-        const spawnEnv = (harness.m.spawn.mock.calls[0][0] as { env: Record<string, string> }).env
-        expect(spawnEnv.FLAVOR_TOKEN).toBe('tok')
-        expect(spawnEnv.CLAUDE_CONFIG_DIR).toBe('/tmp/iso-cfg')
-        // TERM is always set so interactive TUI agents initialize (agy/bubbletea
-        // drops to its login menu without it).
-        expect(spawnEnv.TERM).toBeTruthy()
-        // process.env must stay clean so the parent's scanner is unaffected.
-        expect(process.env.CLAUDE_CONFIG_DIR).toBeUndefined()
-        expect(process.env.FLAVOR_TOKEN).toBeUndefined()
-        await reachReady()
-        msg.resolve(null)
-        await promise
+        // The host shell may legitimately carry these already (a test run
+        // launched from inside a Claude Code session exports CLAUDE_CONFIG_DIR,
+        // for example). Clear them for the duration of this test so the
+        // "no leak into process.env" assertions observe what runAgentPty did,
+        // not what the runner's environment happened to contain. Restoring in
+        // finally also keeps the tail of this test (await promise) on the
+        // non-throwing path — an assertion throw here would abandon the
+        // in-flight runAgentPty call on the shared harness, which then
+        // surfaces as an unrelated unhandled rejection in a later test.
+        const savedConfigDir = process.env.CLAUDE_CONFIG_DIR
+        const savedFlavorToken = process.env.FLAVOR_TOKEN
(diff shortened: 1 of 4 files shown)


7bata:
Heads-up on the red `test` check here: it is **not** caused by this PR.

The failing step is `bun typecheck` → `typecheck:web`, and the error is in `web/src/components/assistant-ui/markdown-a.test.tsx`:

```
src/components/assistant-ui/markdown-a.test.tsx(52,5): error TS2322:
  Types of property 'showSessionSummaryInChat' are incompatible.
    Type 'boolean | undefined' is not assignable to type 'boolean'.
```

This PR touches four `cli` test files and nothing under `web/`, so it cannot reach that code path.

Two independent checks that it is pre-existing on `main`:

1. I ran `bun run typecheck:web` on pristine `upstream/main` (`fea42212`, "Release version 0.27.3") in a clean clone with no local changes — same file, same line, same error.
2. Your own CI run for "Release version 0.27.3" on `main` is also failing.

For whoever picks it up: `HappyChatContextValue.showSessionSummaryInChat` is declared as a required `boolean` in `web/src/components/AssistantChat/context.tsx`, but the `chatContext()` test helper in that file builds the value literal without it, so the object is missing a required property. It looks like fallout from the `AGENT_NOTIFY_SUMMARY` chat-display settings change.

I left it alone rather than folding a `web/` fix into this PR, since CONTRIBUTING asks for one concern per PR.

7bata:
Addressed in 9341e3ae — good catch. The `finally` now deletes when the variable was originally absent instead of leaving a leaked value behind, which was exactly the failure mode on the path that matters (assertion throws because a real leak happened → saved value is `undefined` → old code restored nothing). `runAgentPty.test.ts`: 35 passed after the change.

7bata:
Addressed in 85e5ec88 — you're right, and it's the same failure path as the last one: `promise` was scoped inside the `try`, so `finally` couldn't reach it and the abandoned run kept the shared harness with nothing left to resolve it.

One deliberate deviation from the suggested snippet: I restore the environment *before* settling the run rather than after. Restoring is synchronous and can't fail, so doing it first leaves `process.env` correct even if the settle hangs; the other order leaves it wrong for the duration of the await.

`runAgentPty.test.ts`: 35 passed.

Issue #1526, opened 2026-08-12
Proposal: upstreaming cross-machine terminal attach, session survival, and multi-account support — how would you like it split?


7bata:
### What feature would you like to see?

I've been maintaining a fork that adds four subsystems on top of HAPI. CONTRIBUTING says no mega PRs and to open an issue first for significant features, so I'm asking before writing any PR: **which of these do you want at all, and how small do you want the chunks?**

Listed smallest-first. Each is self-contained and can land independently, in this order:

**1. Session survival + zombie reaping.** When the controlling terminal hangs up (SSH drop, laptop lid, `tmux kill`), the session switches to remote mode instead of dying; the hub reaps sessions left stuck in `running` after connection loss, and self-heals them when the CLI actually reconnects. This is the smallest piece and reads mostly as a bug fix — it touches existing runner/hub files rather than adding modules. **If you only want one thing from this list, I'd suggest this one.**

**2. `hapi attach` — control a session on any machine from any terminal.** A terminal client with a session picker, session view, composer, and permission banner, so you can approve tool calls and steer a session from a second machine without the web UI. This is the big one: ~66 new files, and almost entirely new modules rather than edits to existing code.

**3. Multi-account.** Choose which Claude account a session spawns under (either a config-dir profile or a token), show usage per account, and log in a remote account from the attach UI.

**4. Host registry.** A machine list so `attach --host <name>` works without retyping addresses.

On reviewability: across the whole fork, 121 files are new and only 35 are edits to existing files, so the conflict surface with `main` is much smaller than the raw line count suggests. I have already synced the fork onto current `main` (your latest 75 commits merged in) with all four packages' tests green, so this is not a branch that has drifted away from you.

I'm not proposing to upstream my internal design docs or my personal account-naming conventions — those stay in the fork.

### Additional information

**I've already opened one small PR as a starting point: #1525** (`test(cli): make the suite pass on macOS hosts`). It is test-files-only and unrelated to the four items above — I hit 11 cli test failures on a macOS host while syncing, reproduced all of them on pristine `upstream/main`, and fixed them. That one needs no decision from you here; it's just the smallest useful thing I could send first.

**Code visibility.** The four subsystems currently live in a private fork, so there is nothing for you to read yet beyond this description — I deliberately did not push the whole fork history to a public repo (it carries internal infrastructure details). Say which piece you want to look at first and I'll push a cleaned branch for it so you can read real code before committing to anything.

[...]

PR #1527, opened 2026-08-12
feat(cli): survive terminal hangup by switching the session to remote mode

7bata:
Addresses item 1 (first half) of #1526.

Closing the local controlling terminal (SSH drop, terminal app quit, tmux/screen detach without a persistent session) sends `SIGHUP` to the foreground process group. There was no handler for it, so the kernel default killed the runner before it could archive gracefully — leaving the hub session stuck at `lifecycleState='running'` forever (the split-brain case `hub/src/web/routes/sessions.ts` documents at the archive endpoint).

**What this does**

- **Opt-in, scoped to the claude flavor.** `registerProcessHandlers()` takes an optional `{ surviveTerminalHangup?: boolean }`. Only `runClaude.ts` opts in. Every other flavor (codex, cursor, copilot, grok, pi, agy, kimi, opencode) calls it with no options and keeps the platform default — SIGHUP still terminates the process, byte-for-byte unchanged from upstream.
- When opted in, SIGHUP marks the terminal as lost and installs an output guard on stdout/stderr/stdin, so in-flight writes from readline/ink don't crash the process via `EPIPE`/`EIO` once the terminal's far end is gone.
- Once the terminal is marked lost, the local claude child dying — by signal, non-zero exit, **or a clean `code=0` exit** (a real claude child can exit gracefully when its PTY disappears) — is treated as "the terminal died out from under the child," and is redirected into the existing local→remote switch path. The switch consumes one-time flags (`--fork-session` etc.) exactly like any other local→remote switch.
- `RemoteLauncherBase`'s `hasTTY` is evaluated live (`isTTY && !isTerminalLost()`), not cached at construction, so a terminal that dies mid-run stops `console.clear()`/ink render/`setRawMode` immediately.
- Escape hatch: `HAPI_EXIT_ON_HANGUP=1` (strict string match) makes SIGHUP do a graceful archive-and-exit instead — close to the old behavior, but reported to the hub properly. There's no canonical env-var reference page in `docs/` today; happy to document this wherever you prefer.

**Events covered (deliberately narrow for this first PR)**

`SIGHUP` only. stdin EOF/close, parent-process death, and PTY loss that doesn't produce a SIGHUP (e.g. a disowned/non-foreground process) are **not** detected here — the terminal-app-quit / SSH-drop / window-close family all arrive as SIGHUP via the kernel's foreground-process-group rule, which is the case this targets. The others are natural follow-ups if you want them.

**Platform note**

Effectively POSIX-only. Windows emulates SIGHUP on console close and the handler does run, but the OS then terminates the process unconditionally after a short grace period, so a session can't outlive its console there. The real-process integration tests are gated `describe.skipIf(process.platform === 'win32')`.

**Behavior matrix** — see the matrix in the #1526 comment accompanying this PR.

**Tests**

New/updated suites cover: opt-in gating (no SIGHUP listener without the flag — asserted at the `process.listeners` level and with a real spawned child that dies on SIGHUP without opt-in); survival with opt-in (real child process, real signal); `HAPI_EXIT_ON_HANGUP=1` graceful archive (unit + real process); output-guard behavior incl. stdin, unexpected error codes, and idempotency; clean-exit-vs-`/exit` disambiguation; `--fork-session` one-shot consumption across the terminal-loss switch; live `hasTTY` for launchers constructed before the terminal died. Full `cli` suite: 2483 passed / 10 failed — the 10 are the pre-existing macOS-host baseline addressed separately in #1525 (with #1525's commits stacked locally, the suite is fully green: 2493 passed / 0 failed).

**Known edges**

- Mid-generation terminal loss drops the current unsaved turn; the session resumes from the last save point via `--resume` (same as any local→remote handoff).
- `/exit` racing terminal death in the same tick resolves to "keep the session alive, switch to remote" — worst case is an extra remote session to end manually, judged safer than silently losing a live session.

**Heads-up on CI**

The `test` job will come up red on `typecheck:web` — a pre-existing breakage on upstream `main` itself (`web/src/components/assistant-ui/markdown-a.test.tsx:52`: the test's context object is missing the required `showSessionSummaryInChat` field). Reproducible on a clean `main` checkout, and the Release 0.27.3 CI run is red the same way. Unrelated to this PR.

**AI disclosure** (per CONTRIBUTING's Vibe Coding policy): this was produced by a multi-agent workflow — Claude Fable 5 for planning and task decomposition, Claude Sonnet for implementation, and Claude Opus for code review and acceptance. A human reviewed the result and ran the verification above.

### DIFF
--- a/cli/src/agent/__fixtures__/runnerLifecycleProcessChild.ts
+++ b/cli/src/agent/__fixtures__/runnerLifecycleProcessChild.ts
@@ -1,0 +1,42 @@
+// Standalone child process for runnerLifecycle.process.test.ts.
+//
+// Runs createRunnerLifecycle().registerProcessHandlers() in a *real*,
+// separately-spawned OS process — the only way to prove the SIGHUP handler
+// actually replaces the kernel's default terminate-on-SIGHUP action (a
+// same-process `process.emit('SIGHUP')` in a unit test only proves the
+// listener ran, not that it beat the default disposition).
+//
+// Uses fs.writeSync(1, ...) instead of console.log: stdout is a pipe when
+// spawned from the test, and Node/Bun buffer pipe writes asynchronously —
+// a subsequent process.exit() (the HAPI_EXIT_ON_HANGUP=1 path) can truncate
+// anything still in flight. Synchronous writes make every line observable
+// to the parent before exit.
+import fs from 'node:fs'
+import { createRunnerLifecycle } from '../runnerLifecycle'
+
+const session = {
+    updateMetadata: (fn: (m: Record<string, unknown>) => Record<string, unknown>) => {
+        const next = fn({})
+        fs.writeSync(1, `EVENT updateMetadata ${JSON.stringify(next)}\n`)
+    },
+    sendSessionDeath: (reason: string) => {
+        fs.writeSync(1, `EVENT sendSessionDeath ${reason}\n`)
+    },
+    flush: async () => true,
+    close: async () => {
+        fs.writeSync(1, 'EVENT close\n')
+    }
+} as unknown as Parameters<typeof createRunnerLifecycle>[0]['session']
+
+const lifecycle = createRunnerLifecycle({ session, logTag: 'child' })
+// SIGHUP survival is opt-in — this fixture exercises the opted-in flavor.
+// See runnerLifecycleProcessChildNoOptIn.ts for the un-opted-in sibling.
+lifecycle.registerProcessHandlers({ surviveTerminalHangup: true })
+
+fs.writeSync(1, 'READY\n')
+
+// Keep the event loop alive so the parent test can observe "still running"
+// well after SIGHUP — without this the process would exit on its own once
+// there is nothing left to do, which would make the survival assertion
+// meaningless.
+setInterval(() => {}, 1000)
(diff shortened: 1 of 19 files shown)


7bata:
Addressed in 597c28d6.

Both classification points now use `isTerminalLost() || await waitForTerminalLoss(100)` instead of a bare synchronous read. `waitForTerminalLoss()` (new in `terminalLossState.ts`) is a one-shot bounded wait woken by `markTerminalLost()`: it resolves `true` immediately if the terminal is already known lost, `true` the instant a delayed SIGHUP lands while the wait is pending, or `false` after 100ms with no timer/waiter leak.

Being explicit about the semantics:

- **This is a bounded wait, not an ordering proof.** If the parent's SIGHUP never arrives within the window, classification falls back to exactly the pre-change behavior (clean exit → genuine user `/exit`; abnormal exit → launch failure surfaces as before). The failure direction is "behave as this PR's baseline did", not something new.
- **Cost:** every local-child exit (clean or abnormal) can now take up to 100ms longer to classify. Accepted as the price of closing the race; the tradeoff is called out in a code comment at both call sites.
- **Ordering coverage:** `claudeLocalLauncher.childExitsFirst.process.test.ts` is a real two-process test that forces the flagged sequence — the child process is SIGHUP'd and exits first, the parent's own SIGHUP is delivered ~20ms later, still inside the window — and asserts the classification is `switch`. A control case (child dies, parent SIGHUP never arrives) asserts the fallback to `exit`, ruling out an always-true implementation. Four unit cases cover `waitForTerminalLoss` itself (already-lost / in-window flip / timeout / reset cleanup).

One semantic note worth flagging: a user who types `/exit` and has their terminal die within the same 100ms window now gets a surviving remote session rather than an ended one — same direction as the pre-existing same-tick tie-break already documented in the code, just with a wider window. Worst case is an extra remote session to end manually.

Issue #1486, opened 2026-08-10
feat(runner): hub-held operator-approved proof rebind + terminal resume grant


heavygee:
## Context

Follow-up from #1473 (A2A P0.5 peer delivery provenance).

Under the same-UID adversarial model, cold runner recovery cannot use anything stored under `HAPI_HOME` / `settings.json` as a possession proof (`machineTag`, file reenroll grants, offline tag rebind). Codex correctly Blocker'd those paths on #1473.

Memory-only `runnerProof` is lost on stop/crash/reboot. That leaves an impossible triangle on cold start:

1. **Rotate machine id** (`allowLegacyReenroll`) — runner recovers, but Cursor/Pi exact-id resumes strand on the retired id.
2. **Fail closed** (no rotate, no rebind) — runner never registers after ordinary cold start.
3. **Rebind via persisted tag/file** — same-UID sibling can forge session provenance.

#1473 ships (1) for runner availability and documents (2)/(3) as rejected. Residual Codex Majors on tip `07e775cd6` are intentional under that choice.

Related cold-start orphan discussion: #929 (process re-attach; complementary, not the same as proof/session machine-id remap).

## Proposal

Hub-held, **operator-authenticated** (web/Telegram — not CLI token from `settings.json`) one-shot flows:

1. **Proof rebind / session remap** — after cold start mints a new proof (and optionally a new machine id), operator approves an atomic remap of sessions from the retired machine id to the live one (or rebind proof on the same id).
2. **Terminal resume capability grant** — ordinary `hapi resume` is not a runner-tracked descendant, so peercred mint fails by design. Operator one-shot grant (or explicit fail-loud) so scoped RPC does not silently no-op.

## Acceptance

- Same-UID sibling cannot mint/rebind without the operator channel.
- Cursor/Pi exact-id resume works after reboot once operator approves (or sessions are remapped).
- Terminal resume either gets a capability or fails with an actionable error (no silent unattributed connect that permanently rejects RPC registration).

## Non-goals

- Restoring file reenroll grants or tag-gated offline rebind under `HAPI_HOME`.
- Auto-remap authenticated only by `machineTag` / CLI API token on disk.

heavygee:
**P0.5 decision (see #1473):** provenance is best-effort for this release. Same-UID "malicious sibling forges stamps" is not the threat model we are blocking ship on.

[...]

PR #1528, opened 2026-08-12
fix(hub): make session timeout and reconnect state durable and consistent

7bata:
Addresses item 1 (second half) of #1526 — the two defects called out there.

**Up front — this PR flips two deliberate upstream test assertions** (`hub/src/sync/permissionModePersistence.test.ts:66`, `hub/src/sync/sessionModel.test.ts:1361`). Both previously asserted that runtime `active` state must **not** persist across restarts — a deliberate, commented design decision (stale sessions would otherwise look falsely online after a restart). That concern is real; this PR bounds it instead of dismissing it: after a restart, a stale persisted `active=true` row is converged back to `false` by the existing `expireInactive` sweep within ~35s worst-case (30s heartbeat threshold + one 5s tick), and a new regression test pins exactly that convergence path. In exchange, genuinely-connected sessions no longer show as offline after every hub restart until their next keepalive lands. Context below.

**1. `active=true` was never persisted on reactivation** (pre-existing upstream bug). `markSessionActive` / `handleSessionAlive` flipped the in-memory flag but never called `store.sessions.setSessionActive(id, true, …)`, unlike both `active=false` convergence points which do persist. A session that reconnected after going inactive reverted to "inactive" on the next hub restart even though its CLI was still connected. Fixed on the false→true edge only (no store write per repeat keepalive).

**2. The timeout path didn't reconcile `metadata.lifecycleState`** — the known `inactive`/`running` split-brain. Deliberately **not** fixed at the 30s expiry point: a brief network blip that misses a couple of heartbeats must not escalate into an archive. Instead:

- A new `SessionReaper` (periodic sweep, **disabled by default** — see the review discussion below; opt in via `HAPI_REAPER_INTERVAL_MS`) archives sessions that have been disconnected **and** `lifecycleState='running'` for longer than a stale threshold (default 30 min) — `archivedBy: 'hub-reaper'`, explicit `archiveReason`, broadcast as a normal `session-updated`.
- A reconnect self-heal (`healReaperArchivedSession`) restores `lifecycleState='running'` the moment a reaper-archived session's CLI actually comes back, so the reaper can never permanently bury a live session.
- Staleness is measured from the later of the session's last heartbeat and the reaper's own start instant ("hub-downtime floor"), so a hub outage is never charged against sessions; everyone gets a fresh full window after the hub comes back.
- Residual window: between the 30s `active=false` flip and the stale threshold, a session reads `active=false` + `lifecycleState='running'` — i.e. "disconnected, not yet archived." Up to ~30 min wide at the defaults; that's a knob (`HAPI_REAPER_STALE_MS`), not code.
- Knobs: `HAPI_REAPER_INTERVAL_MS` / `HAPI_REAPER_STALE_MS` (non-negative integers; `0` disables; documented in `hub/README.md`). **Ships disabled by default** (interval `0`) per the review below — heartbeat age alone can't prove process death, so heuristic reaping is operator opt-in. Once enabled, the stale threshold defaults to 30 min.

**Behavior notes reviewers should know**

- `markSessionArchivedFromHub` now broadcasts `session-updated` on success — this also adds a broadcast to the existing `/archive` fallback path (previously silent DB update).
- Historical rows whose `metadata` predates `lifecycleState` entirely are never touched by the reaper (intentional compatibility boundary).

**On #1486 — no code coupling, but one behavioral interaction worth naming.**

Verified by review: nothing here touches operator auth, proof rebind, machine ids, or resume capability grants.

There is a behavioral overlap worth surfacing rather than leaving for someone to find. #1473 shipped machine-id rotation (`allowLegacyReenroll`) for cold-start runner recovery, so a runner returning after a reboot registers under a *new* machine id while its earlier sessions stay attached to the retired one. Those sessions are then disconnected with `lifecycleState='running'` — exactly the reaper's archive condition — and the reconnect self-heal cannot fire for them, because the CLI never reconnects under the old id. At the default threshold the reaper would archive precisely the sessions an operator-approved remap (#1486's proposal) would later want to rescue.

Three reasons I don't think that argues against the reaper: it isn't introduced by this change (without it those sessions stay `running` forever, which is the split-brain this PR exists to fix); archiving is reversible; and the threshold is an env knob, not a constant. The reaper is also already conservative about states it doesn't understand — rows whose metadata predates `lifecycleState` are never touched.

If #1486 lands and wants the reaper to leave sessions pending an operator remap alone, `SessionReaper`'s sweep predicate is the single place to add that exemption — happy to wire it once there's something to key on.

**Tests:** 1111 tests / 88 files green, `tsc --noEmit` clean. Coverage includes reaper threshold/idempotency/broadcast/downtime-floor, self-heal on both reconnect entry points (incl. retry-exhaustion under contended metadata versions), reactivation persistence (TDD, red-first), the restart-convergence guard described above, and env-knob parsing (floats rejected → defaults).

**Heads-up on CI**

The `test` job will come up red on `typecheck:web` — a pre-existing breakage on upstream `main` itself (`web/src/components/assistant-ui/markdown-a.test.tsx:52`: missing required `showSessionSummaryInChat` in a test fixture). Reproducible on a clean `main` checkout; the Release 0.27.3 CI run is red the same way. Unrelated to this PR.

**AI disclosure** (per CONTRIBUTING's Vibe Coding policy): this was produced by a multi-agent workflow — Claude Fable 5 for planning and task decomposition, Claude Sonnet for implementation, and Claude Opus for code review and acceptance. A human reviewed the result and ran the verification above.

### DIFF
--- a/hub/README.md
+++ b/hub/README.md
@@ -50,6 +50,8 @@ Dictation and voice-assistant provider keys can also be added from **Settings â†
 - `HAPI_RELAY_AUTH` - Explicit relay auth key. By default the hub obtains and persists an individually revocable key from the relay. A persisted key rejected with HTTP 403 is discarded and reissued once; an explicitly configured environment key must be updated manually.
 - `HAPI_RELAY_FORCE_TCP` - Force TCP relay mode (true/1).
 - `VAPID_SUBJECT` - Contact email/URL for Web Push.
+- `HAPI_REAPER_INTERVAL_MS` - Sweep interval for the session reaper, which archives sessions whose CLI connection dropped without a chance to exit cleanly. Default: `0` (disabled). A dropped connection plus a stale heartbeat is a suspicion the CLI process is dead, not proof - a client riding out a long network partition looks identical to a killed one from the hub's side, and archiving it while the CLI is still alive lets the reopen flow spawn a second agent on the same session before it reconnects. Set this to opt in (e.g. `300000` for 5 minutes) once you're confident that risk doesn't apply to your environment.
+- `HAPI_REAPER_STALE_MS` - How long a disconnected session must go without a heartbeat before the reaper (once enabled) considers it a candidate. Default: `1800000` (30 minutes). Only takes effect when `HAPI_REAPER_INTERVAL_MS` is also set.
 
 ## Running
 
--- a/hub/src/sync/permissionModePersistence.test.ts
+++ b/hub/src/sync/permissionModePersistence.test.ts
@@ -63,7 +63,7 @@ describe('permission mode persistence', () => {
         expect(reloadedSession?.permissionMode).toBe('yolo')
     })
 
-    it('shows persisted permission mode before keepalive after hub restart', () => {
+    it('shows persisted permission mode and active state after hub restart', () => {
         const store = new Store(':memory:')
         const engine = createEngine(store)
 
@@ -83,7 +83,9 @@ describe('permission mode persistence', () => {
         const reloadedEngine = simulateHubRestart(store)
         const reloadedSession = reloadedEngine.getSession(session.id)
 
-        expect(reloadedSession?.active).toBe(false)
+        // Reactivation is now persisted to the store (not just the
+        // in-memory cache), so it survives the hub restart too.
+        expect(reloadedSession?.active).toBe(true)
         expect(reloadedSession?.permissionMode).toBe('bypassPermissions')
     })
 
--- a/hub/src/sync/reaper.test.ts
+++ b/hub/src/sync/reaper.test.ts
@@ -1,0 +1,967 @@
+import { afterEach, beforeEach, describe, expect, it } from 'bun:test'
+import type { SyncEvent } from '@hapi/protocol/types'
+import { Store } from '../store'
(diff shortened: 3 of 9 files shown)


7bata:
Both findings addressed:

**Major — the reaper now ships disabled by default** (02056779). `REAPER_DEFAULT_INTERVAL_MS` is `0`; operators opt in with `HAPI_REAPER_INTERVAL_MS` (once enabled, the stale threshold still defaults to 30 min; `0` on either knob disables; both are documented in `hub/README.md`'s environment table with the rationale). Your reasoning is accepted as-is: heartbeat age can't prove process death, and archiving a CLI that's riding out a long partition — combined with the existing reopen path — could double-run an agent. This is also the option the PR description offered up front, so consider that offer taken; the description's "Behavior" section has been updated to match.

**Minor — the duplicate archive broadcast is removed** (3517c527). The success branch now relies on `refreshSession()`'s own full `session-updated` emit and returns; the hand-rolled metadata-patch emit is gone, and the stale comment that justified it is rewritten to say what actually happens. The broadcast test now pins "exactly one `session-updated` per reap, carrying the full rebuilt session".

Issue #1535, opened 2026-08-12
Security boundary: session-scoped agent credentials and cross-session authorization


zj1123581321:
## Problem

Layer 0 exposure control (#1401) can remove `ping_peer`, `inspect_peer`, and `list_peers` from wrapped agent tool catalogs and prompts. It cannot enforce the operator guarantee against a full-auto agent that can run arbitrary shell commands.

Today `POST /api/auth` exchanges the namespace CLI token for the same `{ uid, ns }` JWT used by human clients. Generic session list/read/resume/message routes authorize only that JWT and namespace. A wrapped agent running as the same OS user can read the persisted CLI token, launch a fresh official HAPI package via `npm exec` / `bun x`, exchange it, and call those generic routes directly.

Client hooks, PATH wrappers, tool allowlists, environment hints, `HAPI_SESSION_ID`, and self-declared `principal` fields therefore cannot be a hard boundary.

Related:
- #1401 — Layer 0 exposure control
- #1402 — peer-triggered resume restores elevated permission mode
- #1473 — trusted peer provenance/session capability primitives
- #1486 — operator-approved proof recovery
- zlxlabs/agent-config#164 — production regression and bypass evidence

## Threat model

The hard-boundary mode must assume:

- wrapped agents may execute arbitrary commands;
- package runners and directly downloaded HAPI binaries are available;
- request headers, bodies, environment variables, and executable names are forgeable;
- the agent must not possess or be able to recover an owner-equivalent credential;
- human Web/Phone takeover remains available.

If runner and agent share one UID and the owner token remains readable under the same `HAPI_HOME`, this contract is impossible. Implementations must make the credential-custody or OS-isolation boundary explicit rather than treating obscurity as authorization.

### P1 — trusted caller identities and credential custody

Introduce distinct authenticated principals:

- `human_owner`: namespace-wide operator actions;
- `runner`: only its bound machine/session lifecycle;
- `session_agent`: only its own session and explicitly granted capabilities.

A wrapped child receives only a session-scoped credential. The owner credential must be unavailable to that child. Acceptable foundations include a separate OS UID/container plus socket ACL/peer credentials, or an independently authenticated operator channel with credentials held outside the agent security principal.

The existing CLI-token JWT must not be usable as both agent and human identity in hard-boundary mode.

### P2 — route authorization

Apply one server-side policy to CLI, MCP, and direct HTTP:

| Action | human_owner | runner | session_agent |
|---|---:|---:|---:|
[...]

Issue #1401, opened 2026-08-07
Operator toggle to disable Layer 0 peer tools (ping_peer / list_peers / inspect_peer) per hub


zj1123581321:
## What feature would you like to see?

A hub-persisted settings toggle that lets an operator turn off the Layer 0 peer tools for agent sessions — same pattern as the `AGENT_NOTIFY_SUMMARY` contract-injection toggle added in #1376. When off: the three MCP tools are not registered, and the system-prompt guidance that steers agents toward `list_peers` discovery is not injected. Human-side features (web/phone takeover, session @-mentions in the composer) stay untouched.

Default can be **on** (current behavior preserved); the ask is only that turning Layer 0 off becomes possible without client-side hacks.

## Additional information

**Why (real incident, 2026-08-06).** I run a multi-agent fleet (Claude Code + Codex under HAPI, most sessions in full-auto). Over ~3 hours, codex sessions from three different repos autonomously called `ping_peer` 17 times against a claude session I was actively typing in at my local TUI. The TUI flipped to Remote Mode mid-keystroke, and the session received 196 hub-injected messages. The initiating agents were not malicious — they followed the injected system-prompt guidance ("To discover peers without a citation, call list_peers…") while trying to coordinate CI status across repos.

**Why the existing gate doesn't hold in fleet mode.** #1195 already treats `ping_peer` as dangerous and gates it on the *initiating* side (manual approval, kept out of auto-allow). But fleets run yolo/full-auto, where approval gates auto-approve. Initiator-side approval × full-auto operation = no effective gate. A target-side / operator-side control is the missing half.

**Alignment with the A2A RFC.** This is not an argument against A2A. RFC #1332's own principle is "One Boss routing … so humans retain audit and interrupt control", and Layer 1 plans hub-capability gating precisely because "a contract that agents must infer from prose is not a contract." A Layer 0 toggle is the minimal interim version of that gating until the ledger lands — and stays useful afterwards for operators who want HAPI as human-remote-control only.

**Workaround today** (for anyone else hitting this): client-side denies/hooks (Claude Code `permissions.deny`, Codex `PreToolUse` hook) can block the tool calls, but that's per-client and doesn't stop a token-holding agent from hitting the hub HTTP API directly — the hub is the right layer for this switch.

zj1123581321:
[...]

PR #1537, opened 2026-08-12
feat: add peer tools exposure toggle (#1401)

zj1123581321:
## Summary

- Add an owner-only Hub `peerToolsEnabled` setting, defaulting to enabled for older settings and bootstrap responses.
- Propagate the setting through create/resume bootstrap and remove peer MCP tools, bridge approvals, and peer citation guidance when disabled across Claude, Codex, OpenCode, and ACP paths.
- Add focused coverage and document the exposure boundary.

## Scope

Refs #1401.
Related to #1535; session authorization remains out of scope and M2 is not activated.

This toggle controls agent-tool exposure only. It is not a REST authorization boundary; existing REST/session/message authorization is unchanged.

### DIFF
--- a/GOALS.md
+++ b/GOALS.md
@@ -5,6 +5,7 @@
 - Add owner-only `peerToolsEnabled` Hub setting; default on for old settings and responses.
 - Apply it to new/resumed bootstrap, Claude HTTP MCP, Codex/OpenCode stdio, and ACP tool advertisements/approvals.
 - Remove peer citation/list-discovery steering when off; human session citation text remains usable.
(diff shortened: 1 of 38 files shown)


zj1123581321:
Status: moved back to Draft despite green GitHub checks. An independent incremental review reproduced a fail-open state that the gate missed: after a successful peerToolsEnabled=true read, a later React Query refetch error retains cached data=true, so Web references can still emit inspect_peer steering while settings are unknown. The observed sequence is pending -> success(data=true) -> refetch error(data=true). This needs one final state-aware fix plus a regression test for successful true followed by rejected refetch before the PR is safe to mark ready again.

PR #1563, opened 2026-08-13
fix(pi): settle autonomous agent lifecycles instead of swallowing them

KorenKrita:
## Problem

When a Pi session is driven through HAPI, an agent lifecycle that Pi starts **on its own** — e.g. an async subagent completing and waking the session up, or scheduled/background work — permanently wedges the session:

- the status stays *thinking/brewing* forever even though Pi finished and went idle;
- new messages are accepted into the queue but never delivered (the FIFO pump is blocked by `piIsStreaming`);
- Stop appears to do nothing, because the abort path waits on a settlement that can never arrive;
- the only escape is killing the session.

## Root cause

`deliveredSettlement` in `cli/src/pi/loop.ts` is a per-prompt-lifecycle latch: it is reset only by `beginPromptLifecycle()`, which runs when HAPI sends a prompt. An autonomous lifecycle never goes through `beginPromptLifecycle()`, so when its `agent_start` arrives, the flag is still `true` from the previous prompt's settlement.

`agent_start` unconditionally sets `thinking=true`, but every settlement path is gated on `deliveredSettlement` being `false`:

- `deliverSettlement()` returns immediately;
- `scheduleLegacySettleFallback()` (the `agent_end` grace timer) returns immediately;
- `schedulePromptLifecycleFallback()` is only armed from the prompt-response branch, which never runs.

So the autonomous turn's `agent_end` + `agent_settled` are swallowed. `thinking`/`piIsStreaming` stay `true` forever, `pumpPromptQueue()` refuses to dequeue (`runPi.ts` gates on `piIsStreaming`), and a user-initiated abort blocks on `abortBoundary.promise`, which is resolved by the same `onAgentSettled` that will never fire — which is why Stop is also dead.

**Evidence** (local repro, hapi 0.27.3): CLI log shows `19:24:07 agent_start / turn_start` with **no** `Prompt accepted` (subagent-completion wake-up), then `19:25:40 agent_end` → `agent_settled` — both swallowed. The session stayed *brewing* for 27+ minutes (Pi idle the whole time) until it was killed.

## Fix

When `agent_start`/`turn_start` arrives while `deliveredSettlement` is still `true` (previous cycle fully settled, no HAPI prompt in flight), open a fresh settlement cycle for the autonomous lifecycle by resetting `deliveredSettlement`, `agentEndObserved`, and `activeAgentSettledSeen`. The existing settlement paths (direct `agent_settled`, legacy `agent_end` grace) then apply to the autonomous turn unchanged.

Deliberately **not** reused: `beginPromptLifecycle()` — it would fabricate prompt/FIFO state (`activePromptId`, pending local-ID association) for a turn that has no HAPI message behind it.

Non-impact on existing flows:

- **Prompt-driven lifecycles**: `beginPromptLifecycle()` is called synchronously *before* `transport.send({type:'prompt'})` (`runPi.ts`), so `deliveredSettlement` is already `false` when their `agent_start` arrives — the new branch does not trigger.
- **Mid-cycle retries** (`agent_end willRetry` → `auto_retry_start` → `agent_start`): the cycle has not settled, flag is still `false` — no trigger.
- **Tool-loop `turn_start`s** inside a running cycle: same, no trigger.

## Tests

Two new cases in `cli/src/pi/loop.test.ts` (`Pi settlement compatibility fallbacks`):

1. autonomous lifecycle settles via `agent_settled` after a fully settled prompt lifecycle;
2. autonomous lifecycle settles via the legacy `agent_end` grace when `agent_settled` never arrives.

Validation: `vitest run src/pi/` — 10 files, 222 tests passed; `tsc --noEmit` clean.

### DIFF
--- a/cli/src/pi/loop.test.ts
+++ b/cli/src/pi/loop.test.ts
@@ -1259,6 +1259,48 @@ describe('Pi settlement compatibility fallbacks', () => {
         expect(h.onAgentSettled).not.toHaveBeenCalled();
     });
 
+    it('settles an autonomous agent lifecycle that starts after the previous prompt already settled', () => {
+        const h = setup();
+
+        // A normal prompt lifecycle runs to settlement.
+        h.controller.beginPromptLifecycle('prompt-1');
+        h.emit({ type: 'response', id: 'prompt-1', command: 'prompt', success: true });
+        h.emit({ type: 'agent_start' });
+        h.emit({ type: 'agent_end', willRetry: false });
+        h.emit({ type: 'agent_settled' });
+        expect(h.onAgentSettled).toHaveBeenCalledTimes(1);
+        expect(h.stateSession.piIsStreaming).toBe(false);
+
+        // Pi wakes up on its own (subagent completion, scheduled work) with no
+        // HAPI prompt in flight. Its settlement must not be swallowed by the
+        // already-delivered previous cycle, or thinking stays true forever.
+        h.emit({ type: 'agent_start' });
+        expect(h.stateSession.piIsStreaming).toBe(true);
+        h.emit({ type: 'agent_end', willRetry: false });
+        h.emit({ type: 'agent_settled' });
+        expect(h.onAgentSettled).toHaveBeenCalledTimes(2);
+        expect(h.stateSession.piIsStreaming).toBe(false);
+    });
+
+    it('settles an autonomous agent lifecycle through the legacy agent_end grace when agent_settled never arrives', async () => {
+        vi.useFakeTimers();
+        const h = setup();
+
+        h.controller.beginPromptLifecycle('prompt-1');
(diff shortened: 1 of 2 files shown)


KorenKrita:
Addressed the Major finding in a1c3a793:

- `deliverSettlement()` now captures `lifecycleGeneration` before the async `syncEntries()` and skips `onAgentSettled` when the generation has moved on.
- Reopening a settlement cycle for an autonomous lifecycle (`agent_start` with `deliveredSettlement` still true) now advances `lifecycleGeneration`, so a previous settlement's in-flight callback turns stale and cannot mark the new lifecycle's abort boundary as settled.

Added a regression test that holds `syncEntries` open across the autonomous `agent_start` and asserts the stale callback does not fire `onAgentSettled` (verified it fails without the fix). `vitest run src/pi/`: 223 tests passed; `tsc --noEmit` clean.

PR #1574, opened 2026-08-14
feat(dsh): native DeepSeek Harness integration — full session frontend parity

swear01:
## Summary

Integrates **DeepSeek Harness** (DSH) as a first-class HAPI agent runtime. A HAPI `dsh` session spawns the official DSH host (`dsh --profile web` + host-only overlay: **no web UI served**, GET / → 404) on loopback, and drives it through the exact official wire protocol the DSH web app uses (`@deepseek-ai/dsh-host-apiproxy` — HTTP POST `/api/<endpoint>` + WebSocket event streams). HAPI replaces DSH's outer frontend/session-management experience; DSH keeps all native session semantics.

```
HAPI Web ──REST/SSE──► HAPI Hub ──Socket.IO/RPC──► HAPI CLI/Runner ──official HTTP+WS client──► DSH Host (127.0.0.1, no frontend)
```

## Highlights

- **New `dsh` flavor**: AGENT_FLAVORS, empty permission-mode list (runtime-discovered DSH presets), official wave-mark icon, NewSession picker, runner spawn dispatch, `hapi dsh` command, doctor checks.
- **Host-only runtime**: official `--patch` overlay strips the web surface (verified 404); spawned under node (NAPI/node-pty crashes under Bun); per-session host lifecycle (readiness handshake, crash capture, graceful stop); auto-install under `$HAPI_HOME/dsh-runtime` (pinned `@deepseek-ai/dsh@0.1.0-rc.6`).
- **Durable session mapping**: HAPI session id = DSH session id (official create-as-resume idempotency, verified cross-process); `dshSessionId`/`dshRuntimeVersion`/`dshEventCursor` in metadata.
- **Event bridge + projector**: native events → live streaming text/reasoning (stable ids, reconnect-safe) + durable `dsh_native` journal (tool trees, subagents, workflows, jobs, plans never flattened) + `dsh_state` snapshots (queue/jobs/goal/questions/approvals, higher-seq-wins). Cursor-based at-most-once replay.
- **Allowlisted typed protocol**: `DshActionSchema` (prompt/interrupt/approval/question/queue/model/goal/subagent/presets/history/fork/feedback) with Zod validation; no arbitrary method proxy; host-global surfaces (settings/credentials/preset authoring) not exposed.
[...]

PR #1583, opened 2026-08-15
feat: add native DeepSeek Harness sessions

wu736139669:
## Summary

- add a native `dsh` agent flavor and `hapi dsh` command backed by the DeepSeek Harness Web HTTP/WebSocket RPC
- support Runner create/resume, model and reasoning selection, permission presets and approvals, cancellation, and queued-message steering
- add machine-scoped history discovery plus single-session continue and multi-session import
- preserve native event timestamps and stable IDs, append imports incrementally, and detect rewritten/diverged history
- map native reasoning, model, usage, cache-read, and cache-write events into HAPI session data
- keep existing HAPI message acknowledgement until `session.prompt` is accepted

## Runtime

Start DeepSeek Harness first:

```bash
dsh web --port 3080
hapi dsh
```

The default endpoint is `http://127.0.0.1:3080`. Set `HAPI_DSH_URL` to override it.

This integration uses the DSH Web protocol and does not add a package dependency on DSH. It was validated with `@deepseek-ai/dsh 0.1.0-rc.6`.

## Validation

- CLI DSH-focused tests: 48 passed
- Hub: 1095 passed, 3 skipped
- Web: 2481 passed
- Shared: 264 passed
- CLI typecheck: passed
- Hub typecheck: passed
- production single-executable build: passed
- live smoke test on two macOS arm64 runners: model discovery and history listing passed

## Known limitations

- DSH sessions are remote-control only in HAPI; the DSH Web runtime owns the native client
- structured `question/requested` prompts are not mapped to HAPI's question UI yet and are cancelled with an explicit error

## Existing upstream baseline failures

- full CLI tests on macOS retain 10 existing AGY carrier failures
- Web typecheck retains the existing duplicate `showSessionSummaryInChat` property error in `web/src/components/assistant-ui/markdown-a.test.tsx:64`

Neither baseline failure is in this PR's diff.


tiann:
Thanks for the substantial work here — the implementation quality and validation effort are clearly high. However, after consideration, I don't think HAPI is the right home for a deep DSH integration, for two fundamental reasons:

1. HAPI's core value proposition doesn't apply to DSH.

HAPI exists to fill a gap: terminal-native agents (Claude Code, Codex, Gemini) lose controllability the moment you step away from the machine, so HAPI adds remote control with seamless local↔remote switching mid-session. DSH has no such gap — it is natively a host + web architecture with first-class remote control, and it has no native TUI for HAPI's local mode to wrap. That's why both this PR and #1574 end up as "remote-control only" sessions: not an implementation gap, but a conceptual one. A DSH session inside HAPI would always be a subset of what DSH's own web experience already offers.

2. Embedding DSH one fixed way works against its core design.

DSH's philosophy is extreme flexibility — everything is a plugin, and its value lies in the open composition space of profiles, goals, workflows, and jobs. HAPI's flavor abstraction assumes a convergent capability set (fixed permission modes, turn-based message model) so a unified UI can render any agent. Mapping DSH onto it either reduces DSH to a lowest common denominator (losing what makes it DSH), or forces HAPI's shared models and UI to grow special cases for one agent's open-ended semantics. Neither outcome serves users of either project well.

Suggested path forward: if there's real demand from users running DSH alongside Claude/Codex who want a single control surface, the right integration point is ACP — DSH already has official (@deepseek-ai/dsh-acp) and community ACP servers, and HAPI already has an ACP agent path (cli/src/agent/). That would be a small flavor registration rather than a bespoke protocol layer pinned to an RC-stage wire format. I'd prefer to wait until that need is demonstrated rather than merge a large DSH-specific surface now.

PR #1585, opened 2026-08-15
fix(agy): show Gemini 3.7 Flash in the agy model list

junmo-kim:
## Problem

Gemini 3.7 Flash never appears in the agy model list, even though `agy models` offers it. New Session shows the hardcoded fallback instead of what agy actually reports.

`agyModels.ts` recovers ids and labels from `agy models`, whose table is meant for people to read. That table has changed shape three times across the 1.1.x line:

| agy | piped `agy models` |
|---|---|
| 1.1.1 – 1.1.4 | display names only (`Gemini 3.6 Flash (High)`) |
| 1.1.5 – 1.1.10 | bare ids (`gemini-3.6-flash-high`) |
| 1.1.11 – | `id<TAB>label` |

The parser's two-column pattern accepts two-or-more spaces, so the tab form introduced in 1.1.11 matches none of its branches, the probe returns nothing, and the caller falls back to the hardcoded mirror. That went unnoticed until 1.1.12 added the three Gemini 3.7 Flash entries, which the mirror does not carry.

## Solution

**Read the structured listing instead of the table.** agy publishes the same data as a machine-readable payload:

```
$ agy --output-format=json models
{"status":"SUCCESS", …, "command":{"name":"models","data":{"models":[
  {"id":"gemini-3.7-flash-high","label":"Gemini 3.7 Flash (High)"}, … ]}}}
```

`--output-format` is a global flag, so here it goes before the subcommand and takes its `=` form — after `models`, or in its space-separated form, agy rejects it. (The headless transport passes the same flag without `=` because it has no subcommand to follow.)

Releases that predate the flag ignore it and print the table, so the text parser stays as the fallback and the hardcoded mirror remains the last resort. The fallback also learns the tab-separated shape, so it stays correct on current agy if the structured payload ever can't be read.

Every agy release checked, 1.0.16 through 1.1.13, ignores an unknown `--output-format` rather than rejecting it, so the fallback always has a table to read. A build that rejected it would emit no models at all, so the probe asks once more without the flag when the first invocation yields nothing either parser can read.

The three new ids are also mirrored into `AGY_MODEL_LABELS`, which still backs the fallback list and the composer's model options.

## Tests

- The structured payload parses into ids and labels; the progress line agy writes alongside it is ignored in either order; output that isn't that payload (an older release's table, an empty list, a truncated object) is declined so the caller falls back; an entry without a label keeps its id.
- The probe spawns `agy --output-format=json models` in that argument order, a release that ignores the flag still resolves through the text parser, an invocation that returns no listing at all is retried without the flag, and one that already answered is not repeated.
- The text parser handles the tab-separated shape and does not mistake the `Fetching available models...` status line for a model row. Its existing space-aligned, bare-id and display-name-only cases are unchanged.

Manually verified end-to-end against Antigravity CLI 1.1.13: the probe returns all 14 models with their labels.

---

Opened before the headless transport landed. Rebuilt on top of it: the picker-row commit is gone with `agyModelKeys.ts`, and what remains is the model listing, which that change left untouched.


junmo-kim:
@tiann Friendly ping. This PR fixes the `agy models` parsing, making Gemini 3.7 Flash usable from the model picker. Could you take a look when you have a moment? Thanks!

PR #1598, opened 2026-08-15
feat(web,hub): add opt-in search for session message content

### LINKED ISSUE #1554: feat(web,hub): add opt-in search for session message content
## Problem

HAPI's session search is useful for locating sessions by metadata, but it currently cannot search the actual conversation messages. This makes it difficult to find a session when the user remembers a phrase from the conversation but not the session title, path, agent, machine name, or session ID.

At the same time, searching message content by default could be surprising and more expensive than the existing metadata search. It should therefore be an explicit, opt-in search mode rather than changing the current default behavior.

## Proposal

Add a search-scope control inside or next to the existing session search box, for example:

- **Session metadata** — default
- **Message content** — opt-in

The default **Session metadata** mode should preserve the current behavior and search the existing fields, including session title, path, agent, machine name, session ID, and other supported metadata.

When the user selects **Message content**:

- Search the text of messages across sessions visible to the user.
- Make it clear that the search is content-based, not metadata-based.
- Group matches by session and show a short matching snippet when possible.
- Let the user open the matching session; scrolling to or highlighting the matching message would be ideal.
- Reuse the existing session filters and access boundaries.

## UX requirements

- Metadata search remains the default for backwards compatibility and predictable performance.
- The scope control should be discoverable without making the search box unnecessarily noisy, such as a segmented control, dropdown, or filter button.
- Switching scope should not silently change the meaning of the current search without a visible indication.
- Loading, empty, and error states should distinguish content search from the normal session list search.

## Implementation considerations

- Do not fetch, decompress, or scan all messages while the default metadata search is active.
- Use an indexed full-text search implementation rather than a per-query scan of every stored message. SQLite FTS5 or an equivalent approach could be considered.
- Consider indexing only user-visible user/assistant message text initially; hidden reasoning, internal payloads, and raw tool data should not become searchable or exposed unintentionally.
- Support indexing both historical messages and newly inserted messages, with a safe backfill strategy for existing databases.
- Consider Unicode and CJK matching behavior, as well as result limits and pagination for large installations.
- Preserve namespace/access isolation and existing session visibility rules.

## Acceptance criteria

- [ ] Existing metadata search behavior and default performance remain unchanged.
- [ ] A clearly labeled control enables or disables message-content search.
- [ ] No message-content search request or full message scan occurs when metadata mode is selected.
- [ ] Content mode returns sessions containing the query in searchable message text.
- [ ] Results identify the matching session and provide useful context where possible.
- [ ] Historical and newly created messages are searchable after indexing.
- [ ] Namespace/access boundaries and existing session filters are respected.
- [ ] Relevant web, hub, and search/index tests are added.

## Non-goals

- Replacing metadata search with content search.
- Semantic/vector search or AI-generated search answers.
- Making hidden reasoning or internal tool payloads searchable by default.

## Related work

- [#547](https://github.com/tiann/hapi/pull/547) — sidebar search and per-group preview limit
- [#551](https://github.com/tiann/hapi/pull/551) — session search and ordering polish
- [#1356](https://github.com/tiann/hapi/issues/1356) — collapsed session search query label
- [#1545](https://github.com/tiann/hapi/pull/1545) — session search UI improvements

This proposal keeps the fast, predictable metadata search as the default while making deeper conversation search available when the user explicitly asks for it.

techotaku39:
## Summary

- Add an explicit `Default` / `Content` scope switch to session search.
- Add SQLite FTS5-backed search for visible user and assistant message content.
- Open matching sessions at the target message, including historical messages outside the initial window.
- Add persistent highlighting and previous/next navigation for multiple matches.
- Surface potentially incomplete search coverage instead of silently hiding results, including when matches are returned.
- Show the content-search loading state immediately when a query changes.

## Problem / Motivation

Session search previously covered session metadata only. Users could not find a session from a phrase remembered from the conversation, and opening a content-search result did not reliably navigate to the matching historical message.

Content search remains explicitly opt-in so the existing metadata search stays the default, predictable, and lightweight path.

The bounded indexing strategy can omit parts of very long messages. In particular, two-character queries use a separate bounded short index. Previously, those omissions could be silent.

### Search scope

- Keep metadata search as the default.
- Preserve searching by title, path, agent, machine name, session ID, and existing metadata fields.
- Add a compact `Default` / `Content` scope control.
- Only issue content-search requests after the user explicitly selects `Content`.
- Apply active-only, date, unread, and machine filters before the Hub applies the result limit.
- Preserve namespace, visibility, and access boundaries.
- Return representative snippets and raw message IDs for navigation.

### Content indexing and completeness

- Add a SQLite FTS5 trigram index for normal content queries.
- Add an indexed two-character n-gram table for short queries.
- Bound the main searchable text to 32,768 characters per message.
- Bound the short-query index to 2,048 characters per message.
- Persist separate metadata for main-index and short-index truncation.
- Return per-match `truncated` metadata.
- Return query-aware `hasPotentiallyIncompleteResults` metadata.
- Show an incomplete-result warning when a query may miss content omitted by an indexing bound.
- Keep the one-character query guard.
- Index only visible user and assistant text.
- Exclude system messages, tool payloads, hidden reasoning, queued messages, and unsupported sidechain content.
- Keep derived indexes synchronized with message insertion, import, copy, merge, truncation, and deletion.
- Preserve atomicity between canonical message writes and derived-index writes.

### Historical navigation

- Add bounded message-context loading around a target message.
- Avoid fetching an entire large session just to locate one result.
- Keep the target message alive until its context is loaded and rendered.
- Preserve raw message IDs through streamed-message coalescing.
- Make repeated navigation to historical messages reliable.
- Add locating feedback while historical context is loading.

### UI

- Keep the target match highlighted persistently.
- Add previous/next controls and match position information.
- Support wraparound navigation across multiple matches.
- Use theme variables for highlights, status indicators, and navigation controls.
- Support dark mode.
- Add English and Chinese translations.
- Show loading immediately after the query changes, before the debounce timer completes.
- Show an informational indicator for individual potentially truncated results.

## Migration / Performance

- Bump the database schema to version 30.
- Rebuild existing content indexes with the larger main-text bound and truncation metadata.
- Backfill short-index metadata for existing messages.
- Keep default metadata search independent from message-content indexing.
- Keep content-search work bounded through FTS5 and the indexed short-query path.
- Keep scoped session IDs serialized and size-bounded before SQLite execution.
- Keep historical navigation limited to a bounded context window.

## Validation

- `bun typecheck` — passed after merging the latest `main`.
- Content-search Hub/store/route tests — 108 passed after merge conflict resolution.
- SessionList, SessionRowSummary, message-window, and search-ranking Web tests — 114 passed.
- `bun run test:shared` — 334 passed.
- Root Playwright `terminal-wrap-fidelity.spec.ts` — 2 passed.
- Content-search Live Playwright on the merged HEAD — 6 passed.
- `bun run build` — passed.

## Related Issues

Fixes #1554

## AI Assistance

Implemented with AI-assisted pair programming using OpenAI Codex and GPT-5.6.


heavygee:
@techotaku39 — thanks for the Content search / FTS work in #1598. I dogfooded that and then needed find-in-this-conversation as well (long threads: not only *which* session, but *which turn*).

I opened a stacked follow-up: https://github.com/tiann/hapi/pull/1762 (draft, Fixes #1756). It reuses your session-scoped `content-search` API and jump path; header search + hit cards only.

Delta on top of your tip (easier than the full vs-main file list):

https://github.com/tiann/hapi/compare/techotaku39:feat/session-content-search...heavygee:feat/in-session-content-search

Happy to rebase onto `main` once #1598 lands, or adjust anything you want different in the in-session UI. No rush — wanted you to see it before it looks like a competing indexer.

heavygee:
I've been running this PR at tip on a live fleet since 2 Sep — 10 days (plus a
few local tweaks to snippet rendering; the indexing path is yours unmodified).
On startup the backfill indexed **83,798 messages** spanning ~3.5 months of
history (2026-05-24 → 09-09), and **9,427** have been indexed live since. Two
observations, in case they're useful:

**1. `MAX_INDEXED_MESSAGE_CHARACTERS` fires rarely, but selectively.**
17 of 83,798 rows (0.02%) hit the cap — 14 from backfill, 3 live. They aren't
random: every one is a bulk paste — meeting transcripts, long emails, handoff
dumps. Low volume, but the content I most want to find again later. Because of
the head+tail split the middle is unfindable, and nothing in the response says
so.

**2. The short-gram cost saturates.**
The review finding behind the cap describes O(n) synchronous inserts, but
`getShortSearchGrams` dedupes through a `Set`, so the count is bounded by
alphabet rather than length. Across all 83,798 indexed rows (backfill uses the
same insert path): mean **145** gram inserts per message, max **1,466**, and the
capped 16k messages average **772**. So raising the cap may be cheaper than
feared — the genuinely linear cost is the trigram FTS insert itself.

Two small asks if you agree: a larger cap, and a truncation indicator in the
search response so a null result can be told apart from a dropped middle.

(I understand why indexing is synchronous — the atomicity requirement in
`messages.ts` rules out deferring it without a durable queue, a much bigger
change than this PR should have to carry.)

Happy to share raw numbers or test a patch.

techotaku39:
Thanks for collecting and sharing these production numbers — this is extremely useful.

I agree with both suggestions. The fact that only 17 of 83,798 messages hit the current cap, and that these are precisely the kinds of long-form messages users may want to search later, makes a larger—but still bounded—indexing limit worthwhile.

The short-gram measurements are also helpful. Since the grams are deduplicated, the write cost is much less linear than initially feared; the main remaining cost is the trigram FTS insertion itself.

I’ll prepare a follow-up update that:

- raises the maximum indexed message length while keeping a hard upper bound;
- adds explicit truncation metadata to the search response, so the UI can distinguish a definitive no-match from a result that may be incomplete because part of a message was not indexed;
- adds regression coverage for long bulk-paste messages and matches located in the truncated middle.

I’ll keep the indexing path synchronous to preserve the existing atomicity guarantees, without introducing a durable queue in this PR.

Thanks again for the detailed data and for offering to test a patch.

Issue #881, opened 2026-06-12
feat(web): right-click context menu for project/workspace groups in sidebar


swear01:
## Summary

Currently, the sidebar has a three-level hierarchy (Machine → Project Group → Session), but **only Session items have a right-click context menu**. Project/workspace groups only expose a \"Copy Path\" button and a \"New Session\" button on hover — no context menu at all.

This inconsistency makes bulk operations on a project group awkward and non-discoverable. This issue proposes adding a right-click (and long-press) context menu to project group rows.

## Proposed Context Menu Items

| Action | Behaviour | Availability |
|--------|-----------|--------------|
| **Copy Path** | Copy the full directory path to clipboard | Always |
| **Archive All Sessions** | Archive every active session in this group at once | Always (no-op if already all archived) |
| **Delete Group** | Delete the project group and all its sessions | **Only enabled when every session in the group is already archived** (greyed out otherwise) |

> **Note:** "New Session" stays as the existing hover button — it's a frequent action that benefits from being persistently visible, not hidden in a menu.

## Why "Delete Group" requires all sessions to be archived first

This prevents accidental deletion of running or important sessions. The user must explicitly archive everything first, making deletion a deliberate two-step action.

## Implementation Notes

- Mirror the pattern of `SessionActionMenu.tsx` — create a `ProjectGroupActionMenu.tsx`
- Apply `useLongPress` hook to the project group row (same right-click + long-press detection used for sessions in `SessionList.tsx`)
- "Archive All" can reuse the existing single-session archive API called in a loop
- "Delete Group" needs a bulk-delete API or sequential calls with a guard that checks all sessions are inactive/archived first
- Intelligent menu positioning (flip above/below based on available viewport space) — same logic already in `SessionActionMenu`

## Mockup (text)

```
[Right-click on project group row]
┌──────────────────────────┐
│  📋  Copy Path           │
│  🗄️  Archive All Sessions │
│  ──────────────────────  │
│  🗑️  Delete Group    (grey if any session not archived) │
└──────────────────────────┘
```

heavygee:
Related intake from operator-dock dogfood: project-level default agent should land on this menu surface once designed with per-project auth.

Tracking: https://github.com/tiann/hapi/issues/1827

PR #1610, opened 2026-08-16
feat(cli,web): per-queued-message Steer for Codex and Cursor mid-turn (#888)

### LINKED ISSUE #888: feat(codex,claude): mid-turn steering — deliver queued messages at step boundaries, not only on full-turn completion
## Summary

Today, when the operator sends a message while an agent turn is still running, HAPI **queues it and only delivers it after the entire turn terminates** (full agentic loop, possibly a dozen tool calls). Both native CLIs we wrap instead deliver such "steering" input **at the next step boundary inside the running turn** (after the current model output / next tool call), without a hard interrupt.

This issue proposes adding **mid-turn steering** (opt-in), and documents the per-flavor feasibility I verified. The two flavors are **asymmetric**, so they need different implementations:

- **Codex** — the app-server protocol has a first-class, non-interrupting `turn/steer` RPC. HAPI just doesn't call it. Clean to implement.
- **Claude** — there is **no** non-interrupting steer primitive in `--input-format stream-json`. I verified empirically that mid-turn stdin messages are ignored by the running turn. Steering would require `interrupt` + resubmit (a soft interrupt), which is a different UX.

## Current HAPI behavior (both flavors deliver only at full-turn end)

**Codex** (`cli/src/codex/codexRemoteLauncher.ts`):
- `turnInFlight` is cleared **only** on a terminal event, and `wakeLoop()` is called only there — `codexRemoteLauncher.ts:2159-2172`.
- The main loop parks in `waitForTurnOrRecovery` (`:1797`) until the turn fully ends, then pulls the next queued message via `session.queue.waitForMessagesAndGetAsString` (`:2955`).
- The app-server client exposes `startThread / resumeThread / startTurn / interruptTurn / compactThread / goal` — **no steer/inject** (`cli/src/codex/codexAppServerClient.ts`).

**Claude** (`cli/src/claude/claudeRemote.ts`):
- `scheduleNextMessage()` (which pushes the next queued user message into the streaming-input iterable) is invoked **only** when `message.type === 'result'`, i.e. at full-turn end — `claudeRemote.ts:258-282`.

## What the native CLIs do

- **Codex**: while a turn runs, the queued message UI says *"Messages to be submitted after next tool call"* (openai/codex#17095); there is also an experimental "steer conversation" mode (openai/codex#9524).
- **Claude Code (TUI)**: typed messages "flush at the next LLM pause, not at true end-of-turn, including pauses between tool calls" — note this is *controversial*; many users actually want an end-of-turn-only queue (anthropics/claude-code#49373, #63190), with a separate real-time steering channel requested (#30492).

### Codex — `turn/steer` exists (verified against codex 0.139.0 schema)

`codex app-server generate-ts` emits a dedicated steering request:

```ts
// ClientRequest: { "method": "turn/steer", params: TurnSteerParams }
type TurnSteerParams = {
  threadId: string,
  clientUserMessageId?: string | null,
  input: Array<UserInput>,   // same shape as turn/start: text | image | localImage | skill | mention
  expectedTurnId: string,    // precondition: must equal the active turn id, else the request fails
}
type TurnSteerResponse = { turnId: string }
```

Relevant constraints baked into the protocol:
- Not every turn is steerable: `NonSteerableTurnKind = "review" | "compact"`. Steering one returns `CodexErrorInfo::activeTurnNotSteerable { turnKind }`.
- There is also a lower-level `thread/inject_items` (raw Responses API items) — not what we want for steering.

HAPI does **not** reference `turn/steer` or `thread/inject_items` anywhere.

### Claude — no non-interrupting mid-turn primitive (verified empirically)

`streamToStdin` writes each pushed message to the child `claude` stdin immediately (`cli/src/claude/sdk/utils.ts:188-192`), so a naive "push the queued message as soon as it arrives instead of on `result`" change is trivial on the HAPI side. **But it does not produce mid-turn steering.**

I tested `claude 2.1.175` directly in `--input-format stream-json --output-format stream-json`:
- Turn 1 prompt: run `sleep 3 && echo AAA`, then `sleep 3 && echo BBB`, as two steps.
- At ~2.0s (before the first `tool_use`), injected a second user message: *"STOP now. Ignore everything before. Reply with exactly: ZEBRA"*.
- **Result:** the turn ran `AAA → BBB → "All done." → result`, never said `ZEBRA`, did not pivot at the `AAA→BBB` step boundary, and **no follow-up turn for the injected message was produced** within the session.

So the running turn ignores (effectively drops) mid-turn stdin input. HAPI's existing "hold in our own queue, push on `result`" is in fact the safe pattern for stream-json — you cannot steer Claude mid-turn just by pushing earlier. Achieving "deliver after the current output" for Claude would require the **`interrupt` control request** (already available: `cli/src/claude/sdk/query.ts:171`, `subtype: 'interrupt'`) to halt at the next safe point, then resubmit the queued message — i.e. a soft interrupt, with the in-flight step discarded.

## Proposed direction

Make this **opt-in** (per-session or setting), because "queue strictly until end of turn" is a legitimate preference that the current behavior already serves well.

- **Codex**: add `steerTurn(params)` to `codexAppServerClient` (`turn/steer`). In the launcher, when a queued message arrives while `turnInFlight` and we hold `currentThreadId` + `currentTurnId`, call `turn/steer` with `expectedTurnId: currentTurnId` instead of waiting. Fallbacks:
  - `activeTurnNotSteerable` (review/compact) → fall back to existing queue-until-end path.
  - `expectedTurnId` race (turn just ended) / unknown method on older codex → fall back to `startTurn` / current behavior. Gate on capability or codex version.
- **Claude**: implement via `interrupt` + resubmit behind the same opt-in. Must decide UX for the discarded in-flight step and how to present it. This is closer to a soft-interrupt than to Codex's true steering — set expectations accordingly.

## Open questions
- Minimum codex version that ships `turn/steer`; best capability-detection (catch "method not found" vs. version check).
- For Claude, whether a soft-interrupt steer is desirable at all, or whether we just document the gap and keep queue-until-end as the only mode.
- Surfacing in the web UI: a "steer now" affordance distinct from plain queue, and distinct from the existing Stop/Abort.

---

Verification artifacts (Codex protocol dump via `codex app-server generate-ts`, Claude stream-json probe scripts) available on request.

swear01:
Closes #888. Supersedes the closed #906 (stale 2-month-old PR, 27 commits, conflicts across 20+ files). Re-implemented on current main, which already has the Pi native steer infrastructure (RPC_METHODS.SteerQueuedMessage, hub rpcGateway.steerQueuedMessage, web Steer button) — this PR fills the codex/cursor gap.

## Codex — app-server `turn/steer` (true mid-turn inject)

- `codexAppServerClient.steerTurn({ threadId, input, expectedTurnId })`
- `codexRemoteLauncher` registers `SteerQueuedMessage`: takes the queued row by localId, rejects control commands / mode mismatches / no active turn, injects with `expectedTurnId` so a stale turn is refused.
- `turnInFlight` now mirrors into `agentState.steeringActive` (increments a `steerEpoch` on abort/cleanup so a late steer can't restore into a torn-down queue).

## Cursor ACP — soft steer (concurrent `session/prompt`, no cancel)

- `AcpSdkBackend.beginSoftSteerPrompt` + `softSteerPrompt`: fires a concurrent session/prompt while the main turn stays in flight; prompt-request counting replaces the single boolean so completion signals stay independent.
- `AcpStdioTransport.sendRequestWithDispatch` splits stdin dispatch from completion — the hub's 30s Socket.IO RPC acks at kickoff, not turn end.
- `cursorAcpRemoteLauncher` registers the handler, drains soft-steer waiters before ready / the next prompt, and restores the row when dispatch fails.

## Queue

`MessageQueue2` gains `takeByLocalId` / `restoreReservation` / `beginReservationDispatch` / `commitReservation` so a steered row is reserved against the main loop; cancel/reset/close cancel outstanding reservations.

## Web + shared

- `SessionChat` gates the Steer button with `isSteeringSupportedForSession(metadata)` + `agentState.steeringActive` (reported by both launchers); Pi keeps its existing `thinking`-based gate.
- `shared`: `AgentStateSchema.steeringActive` + `isSteeringSupportedForSession` (legacy Cursor stream-json sessions excluded).
- Drive-by: remove the duplicate `showSessionSummaryInChat` key in `markdown-a.test.tsx` (TS1117 breaks `bun typecheck` on current main — same fix as the #881 PR).

## Verification

- `bun typecheck` → exit 0
- `bun run test` → exit 0: cli 2508 ✓, hub 1089 ✓, web 2474 ✓, shared 266 ✓
- New tests: queue reservations (10), steerTurn wire call, `isCurrentSteerHandler`, cursor launcher steer (3: in-flight inject, no-turn reject, dispatch-failure restore), ACP backend soft steer (3).


swear01:
Temporarily closing to re-trigger the base-repo Test workflow (GitHub dropped the pull_request events for recent heads; the fork-side Test run already passed). Reopening immediately.

swear01:
Closing this superseded PR: the Codex/Cursor queued-message steer functionality is already present in current main.

PR #1627, opened 2026-08-18
feat(web): anchor the composer settings sheet to the clicked value button

KorenKrita:
## Problem

Since #1475 the composer shows `[model]` and `[effort]` value buttons whose captions are the current values. Clicking either one opened the *same* full settings sheet as the gear, always rendered from the top in the fixed order Model → Effort → Permission → other. So clicking **Effort** never took you to Effort — you landed on the Model section and had to scroll past every model row to reach the control you asked for. The label on the button promised a specific destination the sheet did not deliver.

## Changes

`HappyComposer` gains a `settingsSection` state (`'model' | 'effort' | null`) that anchors the open sheet:

- **model button** expands only the Model area (including the Cursor variant drill-down).
- **effort button** expands only the Effort / Reasoning-effort area.
- **gear** keeps the full sheet (`null` anchor).
- Clicking the other value button while the sheet is open **switches sections in place** rather than closing, so model → effort is one click.
- Clicking the gear over an anchored sheet **expands it to the full sheet**.
- Re-clicking the same trigger still closes, preserving the #1475 toggle-close behaviour.

Implementation notes:

- The section gating happens on sheet-local flags derived inside the `overlays` memo (`sheetModelSettings`, `sheetEffortSettings`, …), so the section-divider conditions keep working and a single visible area never renders a leading, trailing, or orphaned divider.
- `handleGearToggle` is a zero-arg wrapper for the DOM `onClick` consumer, so a `MouseEvent` can never leak into the `section` parameter.
- `modelValueOpen` / `effortValueOpen` now highlight only while their own area is expanded, instead of both lighting up whenever the sheet was open.
- `dismissSettings` clears the anchor along with the sheet and the Cursor drill-down, so no stale anchor can survive a dismissal.

Unchanged: the outside-click dismissal exemption for all three triggers, Pi's mid-turn control liveness (`configurationControlsDisabled`, #1442), the Cursor variant drill-down flow, and narrow-viewport behaviour (value buttons hidden, gear opens the full sheet).

## Testing

- `web`: 2558 tests pass, `bunx tsc --noEmit` clean.
- Updated `HappyComposer.modelEffortButtons` coverage: anchored open from each value button (asserting the other sections stay collapsed), in-place section switching, gear opening the full sheet, and the gear expanding an already-anchored sheet.

### DIFF
--- a/web/src/components/AssistantChat/HappyComposer.modelEffortButtons.test.tsx
+++ b/web/src/components/AssistantChat/HappyComposer.modelEffortButtons.test.tsx
@@ -167,13 +167,49 @@ describe('HappyComposer generic model/effort value buttons', () => {
         expect(screen.getByRole('button', { name: 'Settings' })).toBeTruthy()
     })
 
-    it('opens the settings sheet from the model button with Model before Permission', () => {
+    it('opens only the Model section from the model button (anchored open)', () => {
         renderComposer('claude')
         fireEvent.click(screen.getByRole('button', { name: 'Sonnet 4' }))
+        expect(screen.getByText('Model')).toBeTruthy()
+        // Anchored open: the other sections stay collapsed.
+        expect(screen.queryByText('Permission Mode')).toBeNull()
+        expect(screen.queryByText('Effort')).toBeNull()
+    })
+
+    it('opens only the Effort section from the effort button (anchored open)', () => {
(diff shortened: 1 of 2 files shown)


KorenKrita:
@tiann ready review

PR #1646, opened 2026-08-19
docs(macOS): set launchd runner file limit

### LINKED ISSUE #1174: docs(macOS): launchd runner can inherit maxfiles=256 and fail to spawn sessions
## Summary

The macOS `launchd` runner example in `docs/guide/installation.md` does not set `SoftResourceLimits.NumberOfFiles`. A LaunchAgent can therefore inherit the macOS default soft `maxfiles` limit of 256. The runner remains online, but spawning a child session can fail immediately.

I reproduced this on macOS 26.5.2 with HAPI 0.23.3.2 (based on upstream 0.23.3). The current upstream `main` at `6bddc9d044c8083a9397bd4ffcf89087eead8e1a` still has the same launchd example.

## Observed behavior

- Runner heartbeat/state remained healthy and the Web UI showed the machine online.
- A new session failed before reporting metadata:

  ```
  Session process exited before reporting session metadata
  stderr: error: An unknown error occurred (Unexpected)
  ```

- `launchctl print` showed:

  ```
  maxfiles (soft) => 256
  ```

This makes the failure look like a Hub/Web problem even though the runner process itself is healthy.

## Verified workaround

Adding this to the Runner LaunchAgent and reloading it fixed the spawn failure:

```xml
<key>SoftResourceLimits</key>
<dict>
    <key>NumberOfFiles</key>
    <integer>65536</integer>
</dict>
```

After reload:

- `launchctl print` showed `maxfiles (soft) => 65536`
- a disposable session spawn succeeded
- existing detached session processes remained alive

One subtlety: preserved session processes retain the resource limits they were launched with, so validation needs a newly spawned canary rather than only checking existing sessions.

## Suggested change

Update the macOS Runner plist example in `docs/guide/installation.md` to set `SoftResourceLimits.NumberOfFiles` to a practical value such as 65536. A clearer runner preflight/error for an insufficient file-descriptor limit would also make this easier to diagnose, but the documentation fix is the minimal change.

swear01:
## Summary

- Document `SoftResourceLimits.NumberOfFiles=65536` in the macOS runner LaunchAgent example.
- Prevent the documented runner setup from inheriting a low soft `maxfiles` limit that can make detached session children exit before reporting their startup webhook.

The current launchd example lives in `docs/guide/deployment.md` (the issue's former installation path now links there).

## Validation

- `TMPDIR=<isolated temp dir> bun typecheck`
- `TMPDIR=<isolated temp dir> bun run test` — 6,764 passed, 4 skipped, 0 failed
- Both launchd plist examples parse as XML; runner limit assertion passed.
- `git diff --check`

No macOS runtime validation was possible in this Linux worktree; the issue report includes a macOS reproduction and verified canary spawn after applying this setting.

AI-generated code: this documentation change was prepared with AI assistance and reviewed against the current source and Apple launchd documentation.

Fixes #1174

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -203,6 +203,11 @@ Create plist files for automatic startup on macOS.
         <key>HAPI_RUNNER_SUPERVISED</key>
         <string>1</string>
     </dict>
+    <key>SoftResourceLimits</key>
+    <dict>
+        <key>NumberOfFiles</key>
+        <integer>65536</integer>
+    </dict>
     <key>RunAtLoad</key>
     <true/>
     <key>KeepAlive</key>


swear01:
@codex review

PR #1647, opened 2026-08-19
fix: warn before large session exports

### LINKED ISSUE #1641: feat(web,hub): make large conversation export warning-only
## Problem

Session export currently treats `SESSION_EXPORT_MESSAGE_LIMIT = 20_000` as a hard cap. When a session has more than 20,000 exportable messages, the hub returns `413 Session export too large`, and the web UI only shows a generic export error.

This prevents users from downloading their own complete conversation even when they explicitly want the large export.

Relevant current flow:

- `shared/src/sessionExport.ts`: `SESSION_EXPORT_MESSAGE_LIMIT = 20_000`
- `hub/src/sync/messageService.ts`: returns `too-large` instead of an export payload
- `hub/src/web/routes/sessions.ts`: responds with HTTP 413
- `web/src/components/SessionExportDialog.tsx`: displays only the generic error

## Proposed behavior

Treat 20,000 messages as a warning threshold, not an unconditional block:

1. Detect the message count (and, if practical, an estimated/download size).
2. Tell the user how large the export is and that it exceeds the recommended threshold.
3. Offer an explicit confirmation step such as **Download anyway**.
4. If the user confirms, allow the full JSON or Markdown export to download.
5. If the user cancels, do not create or download a file.

Normal exports at or below the threshold should remain one-step downloads.

## Acceptance criteria

- Exports over 20,000 messages are not automatically rejected solely because of the threshold.
- The warning includes the actual message count and useful size information when available.
- The user must explicitly confirm before a large export starts downloading.
- JSON and Markdown formats both support the confirmation flow.
- The server and browser still have an intentional safety strategy for genuinely resource-exhausting exports; the 20,000-message value itself should not be the user-visible hard stop.
- Tests cover the warning/confirmation path, cancel behavior, and successful large export behavior.

swear01:
Closes #1641

## Summary

- Replace the 20,000-message export hard stop with a count/size warning and explicit confirmation.
- Preserve the absolute 100 MiB serialized-export safeguard.
- Support confirmed JSON and Markdown downloads with localized warning text.

## Tests

- `bun typecheck`
- `bun run test`
- `CI=1 PLAYWRIGHT_WEB_PORT=5180 bun run test:e2e -- terminal-wrap-fidelity.spec.ts composer-copy.spec.ts`

AI assistance: OpenAI GPT-5.6-luna via the pi coding agent.

### DIFF
--- a/hub/src/sync/messageService.test.ts
+++ b/hub/src/sync/messageService.test.ts
@@ -175,7 +175,7 @@ describe('MessageService goal status filtering', () => {
         expect(result.payload.messages.map((message) => message.id)).toEqual([normal.id, scheduled.id])
     })
 
-    it('returns too-large instead of truncating an export over the cap', () => {
+    it('returns a warning instead of truncating an export over the threshold', () => {
         const store = makeStore()
         const session = makeSession(store, 'session-export-cap')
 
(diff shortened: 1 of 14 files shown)


swear01:
Superseded by #1648, which implemented the same #1641 scope and was merged into main.

PR #1667, opened 2026-08-21
ci: pin codex-action to v1.11

junmo-kim:
All three Codex workflows (`codex-pr-review`, `codex-mention-response`, `issue-auto-response`) have been failing since codex-action v1.12 landed (~2026-08-20 16:00 UTC): its new protected-run validation rejects the deliberate `model_providers.codex-action-responses-proxy.*` stream-retry overrides in `codex-args` under `drop-sudo`.

This pins all three workflows to v1.12 and moves the retry tuning into a trusted `codex-home` config.toml seeded by an earlier step — the action preserves existing config content, so the settings survive intact and the validation is satisfied. Pinning also stops the workflows from floating, which is what turned an upstream release into a repo-wide outage.

### DIFF
--- a/.github/workflows/codex-mention-response.yml
+++ b/.github/workflows/codex-mention-response.yml
@@ -114,9 +114,18 @@ jobs:
         if: steps.check_skip.outputs.should_skip != 'true'
         run: git fetch --all
 
+      - name: Seed stream retry tuning in codex-home
+        run: |
+          mkdir -p "$RUNNER_TEMP/codex-home"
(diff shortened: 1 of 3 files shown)


junmo-kim:
@tiann This PR fixes the repo-wide `pr-review` failure across all three Codex workflows. Since v1.12, the action rejects the deliberate stream-retry overrides in `codex-args` under `drop-sudo`, and because these run via `pull_request_target`, every open PR shows the same error until this merges.

PR #1670, opened 2026-08-22
ci: drop the codex-home config seed that breaks TOML parsing

junmo-kim:
Follow-up to #1667: its codex-home config seed doesn't work — the action appends its own `[model_providers.codex-action-responses-proxy]` table to `config.toml` instead of merging, so every Codex workflow fails with a TOML duplicate key error (verified on an external PR run). And since v1.12 rejects the flaky-relay retry overrides in `codex-args`, the tuning cannot be expressed on v1.12 at all.

This pins all three workflows to v1.11, the last release that accepts those deliberate overrides under `drop-sudo`, and drops the seed — restoring the original behavior fully intact. Once the action either merges existing config or exposes first-class inputs for proxy tuning (same root cause as openai/codex-action#80), upgrading should be straightforward.

### DIFF
--- a/.github/workflows/codex-mention-response.yml
+++ b/.github/workflows/codex-mention-response.yml
@@ -114,18 +114,9 @@ jobs:
         if: steps.check_skip.outputs.should_skip != 'true'
         run: git fetch --all
 
-      - name: Seed stream retry tuning in codex-home
-        run: |
-          mkdir -p "$RUNNER_TEMP/codex-home"
-          cat > "$RUNNER_TEMP/codex-home/config.toml" <<'TOML'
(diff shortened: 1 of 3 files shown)


junmo-kim:
@tiann Sorry, the fix from yesterday (#1667) didn't hold. The config seed it added collides with the action's own `[model_providers.codex-action-responses-proxy]` table (the action appends rather than merges), so every Codex workflow is now failing with a TOML duplicate key error. This PR drops the seed and pins v1.11 instead, which accepts the original `codex-args` overrides, so review automation is back to its pre-breakage behavior. Details in the body.

PR #1674, opened 2026-08-23
feat(web): expose opencode /compact and /clear builtin slash commands

rzsgsfm:
## Problem

On the hapi **remote web/phone** UI, the `/` slash-command menu is built from `BUILTIN_SLASH_COMMANDS` in `shared/src/slashCommands.ts`. The `opencode` builtin list omits `compact` (and `clear`), so `/compact` never appears in the mobile `/` menu — even though the parsing layer (`resolveOpencodeSlashCommand` in `cli/src/opencode/utils/slashCommands.ts`) already resolves `/compact` to `{ kind: 'compact' }` and triggers native OpenCode compaction, and `/clear` to `{ kind: 'clear' }`.

By contrast, `pi`, `codex`, `grok`, and `claude` all list `compact` in their builtin commands, which is why `/compact` shows up there but not for opencode.

## Change

- Add `compact` and `clear` to `BUILTIN_SLASH_COMMANDS.opencode`.
- Add a web test asserting opencode exposes `compact`, `clear`, `help`, `status`, `plan`, `default`, `init`.

## Verification

Manually confirmed the opencode builtin list now includes `compact` and `clear`. Test added in `web/src/lib/codexSlashCommands.test.ts`.

### DIFF
--- a/shared/src/slashCommands.ts
+++ b/shared/src/slashCommands.ts
@@ -48,6 +48,8 @@ export const BUILTIN_SLASH_COMMANDS = {
         { name: 'plan', description: 'Enable plan mode; use /plan off to return to default', source: 'builtin' },
         { name: 'default', description: 'Return OpenCode permission mode to default', source: 'builtin' },
         { name: 'init', description: 'Generate or refresh AGENTS.md for this project', source: 'builtin' },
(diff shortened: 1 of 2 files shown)


rzsgsfm:
Closing — the upstream design intentionally keeps opencode /compact and /clear out of the autocomplete menu (the resolver still accepts manual input). Thanks for the CI feedback.

PR #975, opened 2026-06-27
feat: add OMP (Oh My Pi) coding agent support

KorenKrita:
## Summary

Adds OMP ([Oh My Pi](https://omp.sh), `@oh-my-pi/pi-coding-agent`) as a first-class Hapi agent flavor. This branch has been refreshed on top of current `main` and aligned with the current Pi/OMP RPC contracts while keeping the existing Claude, Codex, Cursor, Grok, and Pi paths unchanged.

### Shared layer
- Adds `omp` to the supported agent flavors, labels, capabilities, and metadata schema.
- Adds OMP model/session types and OMP-specific RPC method constants.
- Keeps OMP approval mode fixed to `yolo`, matching OMP RPC-mode behavior.

### CLI
- Adds an OMP JSONL RPC transport with the required `ready` handshake, timeout/error handling, and command buffering.
- Supports OMP session start/resume, including exact session-file preference when both a session file and session ID are available.
- Handles OMP model/provider pairs and current model/thinking metadata contracts.
- Maps OMP goal, compaction, and thinking-level events onto Hapi's existing generic event channels.
- Requests OMP's bounded `progress` Subagent subscription when supported, with a silent fallback for older OMP versions.
- Validates and bounds Subagent lifecycle/progress frames, then projects each child onto one coalesced `Agent` card with live activity, model, tool, retry, token, request, and duration state plus terminal cleanup.
- Registers OMP-only RPCs for compaction, steering/interrupt/follow-up modes, model/thinking cycling, and session stats.
- Resolves the OMP executable from PATH or the common Bun global install location.

### Hub
- Adds flavor-guarded OMP RPC forwarding and the OMP models endpoint.
- Recognizes OMP session metadata when extracting agent session identity.

### Web
- Adds OMP model and thinking-level queries/panels, provider grouping, composer wiring, flavor icon, new-session option, and locale strings.
- Uses model-specific thinking efforts and preserves non-standard current values safely.
- Resumes OMP sessions using the exact session file when available, matching the Hub behavior.
- Coalesces repeated OMP Subagent progress updates by child ID and prefers live activity in the existing `Agent` card subtitle.

### Out of scope
- Raw `subagent_event` streaming and child transcripts are intentionally deferred; this PR subscribes only to bounded progress data.
- OMP extension-host requests such as `host_tool_call`, `host_uri_request`, and `extension_ui_request` are not wired because Hapi does not register the corresponding OMP host extensions in this integration.

## Refresh notes

[...]

PR #536, opened 2026-04-27
[codex] restore Codex session history

dsus4wang:
## Summary

Restores existing Codex sessions with transcript history instead of starting with an empty chat when resuming from the web session picker.

## Changes

- Imports Codex transcript messages when resuming with `--hapi-import-history`.
- Restores Codex session id, title, model, reasoning effort, and latest usage metadata from transcript data.
- Improves Codex session title lookup from local index/thread state.

## Validation

- `bun run test:cli -- src/codex/importHistory.test.ts src/modules/common/codexSessions.test.ts src/codex/utils/codexSessionScanner.test.ts src/commands/codex.test.ts src/runner/buildCliArgs.test.ts src/codex/utils/codexUsage.test.ts`
- `bun typecheck`

PR #1677, opened 2026-08-23
feat(opencode): import local session history

junmo-kim:
## Problem

OpenCode stores every conversation locally, but those sessions were invisible once you started using HAPI: there was no way to bring an existing OpenCode history into HAPI, unlike Codex (#536) and Pi (#975), which both already support importing their local transcripts.

## Solution

Mirror the existing Pi import flow for OpenCode:

- The CLI reads OpenCode's SQLite store (`$XDG_DATA_HOME/opencode/opencode.db`, readonly via `bun:sqlite`, no new dependency) and lists sessions over the existing machine RPC channel. Only the SQLite-backed format is supported — installations without `opencode.db` simply report an empty list.
- The hub exposes `GET /api/opencode/sessions` and `POST /api/opencode/import-sessions`. Import creates a HAPI session with `flavor: 'opencode'` + `opencodeSessionId`, appends transcript messages as imported entries keyed by stable local ids (`opencode:<session>:<msg>:<part>`), and only appends the delta on re-import. Diverged or active sessions fail closed instead of being overwritten.
- Resume works out of the box: imported sessions carry the native OpenCode session id, so reopen continues the same conversation through the existing `opencodeSessionId` resume path.
- Web gets an "Import OpenCode history" entry in the new-session picker, mirroring the Pi/Codex dialogs.

## Tests

- Unit tests for the scanner (summaries, envelope conversion, synthetic-text skip, archived-session handling, missing db / missing tables fail-closed).
- Hub route tests covering list annotation, create, delta re-import, divergence guard, and the active-session guard.
- Web component test for the import actions.
- Manually verified end-to-end against a real local OpenCode database through an isolated hub+runner stack (import → messages render → idempotent re-import → delta append on new native messages).


junmo-kim:
Closing — opened by mistake, not ready for review.

PR #1696, opened 2026-08-26
test(cli): stub the Claude binary so the runner integration suite passes without a local Claude install

junmo-kim:
**Problem**

Since e5a8212f, the runner's agent-availability preflight rejects session spawns on machines without a Claude CLI — which includes CI runners. So the four spawn-based tests in `src/runner/runner.integration.test.ts` fail on every recent `main` run, and that red integration check spills onto unrelated open PRs too.

**Solution**

The isolated test env now writes a minimal stub `claude` into the temp home (same pattern as `agentAvailability.test.ts`) and points workers at it via `HAPI_CLAUDE_PATH`, the override the production launcher already honors. Production behavior is untouched; the suite still exercises the real preflight path. Windows skips the stub, and setup only injects the override when the config supplies one.

**Tests**

- Reproduced the exact four CI failures locally with `claude` off `PATH` and no `HAPI_CLAUDE_PATH` (`4 failed | 9 passed`); all pass after the change (`13 passed | 1 skipped`).
- Full `bun typecheck && bun run test` green on the fork.

### DIFF
--- a/cli/src/test/globalSetup.ts
+++ b/cli/src/test/globalSetup.ts
@@ -1,5 +1,5 @@
 import { randomBytes } from 'node:crypto'
-import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join, dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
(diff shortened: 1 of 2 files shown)


junmo-kim:
@tiann heads-up — this also fixes the currently red `integration` check on `main`. Since e5a8212f, the new agent-availability preflight rejects session spawns on machines without a Claude install, which includes CI runners, so every recent Test workflow run fails on the same 4 tests in `src/runner/runner.integration.test.ts` and the red check spills over onto unrelated open PRs. Happy to switch approaches if you'd rather solve it on the CI side (e.g., installing Claude in the workflow).

junmo-kim:
@heavygee could you take a look? Since e5a8212f the `integration` job on `main` has been red (four tests in `src/runner/runner.integration.test.ts` fail without a local Claude install). This small test-infra PR fixes it.

heavygee:
@junmo-kim Merged — thanks for the ping and for the stub. Agreed this was the right place to fix it (test env / `HAPI_CLAUDE_PATH`) rather than installing Claude on CI. Should clear the red `integration` spill on `main` once this lands there.

Issue #842, opened 2026-06-08
metadata.lifecycleState='running' never reconciled against active=0; sessions appear live after host outage


heavygee:
## Summary

`metadata.lifecycleState` and the in-memory `session.active` flag drift apart over time and the hub never reconciles them. Sessions appear as `lifecycleState: 'running'` in the metadata blob (and therefore in the web UI's session-row badge) long after the in-memory `active` flag has been flipped to `false` by inactivity expiration or host disconnect. From the operator's perspective the session looks live; reality is zombie.

A 107-session census on a long-running fleet found 1 session in this state. Pattern: host went offline for days, web UI still showed the session row with the "running" badge.

## Repro

1. Start a session on host A. `metadata.lifecycleState` becomes `'running'` (`cli/src/agent/sessionFactory.ts:78`).
2. Kill host A's runner abruptly (`SIGKILL` the `hapi runner` process so no graceful shutdown runs) OR disconnect host A's network.
3. Wait >30s for the hub's `expireInactive` to flip in-memory `session.active` to `false` (`hub/src/sync/sessionCache.ts:369` and `:389`).
4. Reload the web UI on the operator's host. The session row shows the "running" badge (driven from `metadata.lifecycleState`) but is sortable / filterable as inactive (driven from `active`).
5. Query the DB:
   ```bash
   sqlite3 ~/.hapi/hapi.db "SELECT id, active, json_extract(metadata,'$.lifecycleState') FROM sessions WHERE id='<sid>'"
   ```
   Output: `<sid> | 0 | running` — the divergence.

## Root cause

Two separate state machines for the same concept, no reconciler:

### lifecycleState is metadata (CLI-owned)

`cli/src/agent/sessionFactory.ts:78` initializes new sessions with:

```ts
metadata: {
    ...
    lifecycleState: 'running',
    lifecycleStateSince: now,
}
```

Subsequent metadata writes happen via `cli/src/agent/runnerLifecycle.ts` `archiveAndClose` on graceful close. There is no metadata write on abrupt-kill, host-disconnect, or hub-side inactivity expiration.

### active is in-memory + DB column (hub-owned)

`hub/src/sync/sessionCache.ts` flips `session.active = false` in two paths:

- `markInactive` (line 365-380): explicit transition.
- `expireInactive` (line 384-401): 30s grace after the last `activeAt` heartbeat; runs periodically.

[...]

PR #1706, opened 2026-08-27
fix(hub): verify runner before trusting archiveSession's dead-CLI fallback

### LINKED ISSUE #1705: archiveSession's RpcTargetMissingError fallback marks sessions archived without confirming the runner process is actually dead
## Summary

`SyncEngine.archiveSession` (`hub/src/sync/syncEngine.ts`) only has one way to actually stop the target process — `rpcGateway.killSession(sessionId)`, a **session-scoped socket RPC** keyed on the string `${sessionId}:KillSession`. When that RPC target is missing, the #916 fix (merged) treats it as proof the CLI is already gone and unconditionally marks the row `archived`:

```ts
try {
    await this.rpcGateway.killSession(sessionId)
} catch (error) {
    if (error instanceof RpcTargetMissingError) {
        this.sessionCache.markSessionArchivedFromHub(sessionId, 'Archived from hub (CLI unreachable)')
    } else {
        throw error
    }
}
```

That assumption is false whenever the session-scoped socket registration is stale **but the underlying runner child process is still alive**. Concretely:

- `RpcRegistry` keys handlers on whatever session id the CLI itself registers under when its session socket connects (`hub/src/socket/handlers/cli/rpcHandlers.ts:19`, `rpcRegistry.register(socket, parsed.data.method)`).
- `SessionCache.mergeSessions(oldId, newId, ...)` (`hub/src/sync/sessionCache.ts:1088`) rotates the hub's *canonical* session id during a resume whenever the runner mints a session id different from the one the hub asked it to resume (`syncEngine.ts:3142`, `if (spawnResult.sessionId !== access.sessionId)`). Nothing tells the live CLI socket to re-register its RPC handlers under the new canonical id.
- The session socket can also simply be mid-reconnect (normal churn), independent of any id rotation.

In any of these cases, `killSession(sessionId)` throws `RpcTargetMissingError` even though the OS process is healthy and its runner-level supervision is intact. The current code takes that as "CLI unreachable ⇒ dead" and archives the row. The result: **a live, still-running child process, permanently marked `archived` in the hub, with no further supervision pointed at it** — an orphan that keeps consuming a working directory, API budget, etc., indefinitely, while every operator-facing surface reports it as cleanly stopped.

This is a false-positive that's arguably worse than the original #916 500 — the old behavior at least surfaced an honest error; the current behavior silently reports success.

## Why we know the process can be reachable when the session-socket RPC isn't

The hub already has a second, more robust kill path that isn't wired into `archiveSession`: `rpcGateway.stopRunnerSession(machineId, sessionId)` — a **machine-level** RPC handled by the runner's own `stopSession` (`cli/src/runner/run.ts:910`). That handler resolves the target PID by scanning **both** the requested and confirmed session ids it has recorded for every tracked/persisted child (`cli/src/runner/run.ts:914-990`), so it survives exactly the id-rotation/mismatch case that defeats the session-socket RPC. It's already used this way elsewhere in the codebase to authoritatively confirm a child is gone before touching hub state, e.g. `SyncEngine.reconcilePersistedPtyResumeAttempt` / `reconcilePersistedPiResumeAttempt` (`hub/src/sync/syncEngine.ts:3742-3789`).

`archiveSession` never calls it.

## Relation to other tracked issues

- #916 (closed) — introduced the `RpcTargetMissingError` → archive-from-hub fallback this issue refines. That fix correctly stopped the 500, but traded it for a silent false-positive in the id-mismatch case.
- #842 (open) — `metadata.lifecycleState` vs `active` reconciliation gap. Related but distinct: that issue is about *nothing* ever reconciling a drifted row. This issue is about `archiveSession` *actively* writing a wrong `archived` state while a process is still alive, which then has nothing to reconcile it back either.
- #929 (open) — runner cold start orphaning live-but-detached sessions. Same family of orphan-process problems, different trigger (runner restart vs. archive-time id mismatch).

## Proposed fix

Before trusting `RpcTargetMissingError` as proof of death, have `archiveSession` fall back to `stopRunnerSession(machineId, sessionId)` (when the session's `metadata.machineId` is known) and only proceed to `markSessionArchivedFromHub` if that call also reports the process gone (`already_gone`) — surfacing an error instead of silently archiving if it reports `still_alive`. When there's no known machine, or the machine RPC itself is unreachable, fall back to the current best-effort behavior (nothing stronger to check against).

I have a minimal patch + regression test ready and will open a PR against this issue.

heavygee:
## Summary

- `archiveSession`'s only kill path (`rpcGateway.killSession`, a session-scoped socket RPC) treated any `RpcTargetMissingError` as proof the CLI process was dead and unconditionally archived the row. That RPC target goes stale independently of the process — most notably when `SessionCache.mergeSessions` rotates the hub's canonical session id during a resume — so a missing target does not prove the process is gone.
- Before trusting that fallback, verify with the machine-level `stopRunnerSession` RPC, which resolves the child by PID on the runner side and checks both the requested and confirmed session ids (already used this way by the PTY/Pi resume reconciliation paths). Only archive once that also reports the process gone; if it reports `still_alive`, surface an error instead of silently archiving a live process.
- When the session has no known `machineId`, or the machine RPC itself is unreachable, preserve the prior best-effort behavior (nothing stronger to check against).

Fixes #1705

## Test plan

- [x] New unit tests in `hub/src/sync/syncEngineArchiveSession.test.ts` cover: still-alive (rejects, stays active), confirmed-gone (archives), no-known-machine (falls back to old behavior), machine-unreachable (best-effort archive as before)
- [x] `bun test src/sync/` and `bun test src/web/routes/sessions.test.ts` pass (pre-existing unrelated failures in `loadServerSettings`, `TitleSuggestionService`, `resolveFcmConfig` reproduce identically on `main`)
- [x] `tsc --noEmit` clean in `hub/`

### DIFF
--- a/cli/src/api/apiMachine.ts
+++ b/cli/src/api/apiMachine.ts
@@ -60,7 +60,7 @@ export { normalizeWindowsDriveRoot } from './machinePathPolicy'
 
 type MachineRpcHandlers = {
     spawnSession: (options: SpawnSessionOptions) => Promise<SpawnSessionResult>
-    stopSession: (sessionId: string) => Promise<'stopped' | 'already_gone' | 'still_alive'>
+    stopSession: (sessionId: string) => Promise<'stopped' | 'already_gone' | 'still_alive' | 'unknown'>
     requestShutdown: () => void
 }
 
--- a/cli/src/runner/controlClient.ts
+++ b/cli/src/runner/controlClient.ts
@@ -96,9 +96,9 @@ export async function listRunnerSessions(): Promise<any[]> {
   return result.children || [];
 }
 
-export async function stopRunnerSession(sessionId: string): Promise<'stopped' | 'already_gone' | 'still_alive'> {
+export async function stopRunnerSession(sessionId: string): Promise<'stopped' | 'already_gone' | 'still_alive' | 'unknown'> {
   const result = await runnerPost('/stop-session', { sessionId });
-  return result.status === 'stopped' || result.status === 'already_gone' || result.status === 'still_alive'
+  return result.status === 'stopped' || result.status === 'already_gone' || result.status === 'still_alive' || result.status === 'unknown'
     ? result.status
     : 'still_alive';
 }
--- a/cli/src/runner/controlServer.ts
+++ b/cli/src/runner/controlServer.ts
@@ -19,7 +19,7 @@ export function startRunnerControlServer({
   onHappySessionWebhook
 }: {
   getChildren: () => TrackedSession[];
-  stopSession: (sessionId: string) => Promise<'stopped' | 'already_gone' | 'still_alive'>;
+  stopSession: (sessionId: string) => Promise<'stopped' | 'already_gone' | 'still_alive' | 'unknown'>;
   spawnSession: (options: SpawnSessionOptions) => Promise<SpawnSessionResult>;
   requestShutdown: () => void;
   onHappySessionWebhook: (sessionId: string, metadata: Metadata) => void;
@@ -91,7 +91,7 @@ export function startRunnerControlServer({
         }),
         response: {
           200: z.object({
-            status: z.enum(['stopped', 'already_gone', 'still_alive'])
+            status: z.enum(['stopped', 'already_gone', 'still_alive', 'unknown'])
           })
         }
       }
--- a/cli/src/runner/run.ts
+++ b/cli/src/runner/run.ts
@@ -937,7 +937,7 @@ export async function startRunner(options: { workspaceRoots?: string[] } = {}): 
     }
 
     // Stop a session by sessionId or PID fallback
-    const stopSession = async (sessionId: string): Promise<'stopped' | 'already_gone' | 'still_alive'> => {
+    const stopSession = async (sessionId: string): Promise<'stopped' | 'already_gone' | 'still_alive' | 'unknown'> => {
       logger.debug(`[RUNNER RUN] Attempting to stop session ${sessionId}`);
 
       // Try to find by sessionId first
@@ -1046,8 +1046,15 @@ export async function startRunner(options: { workspaceRoots?: string[] } = {}): 
         logger.debug(`[RUNNER RUN] Session ${sessionId} was previously observed exited`);
         return 'already_gone';
       }
+      // Distinct from 'still_alive': no PID matched this id anywhere (not
+      // tracked, not persisted, no verified-exit tombstone), so there is
+      // nothing here to confirm as alive OR dead. Callers that already have
+      // concrete knowledge of this session (e.g. one they just spawned) should
+      // treat this the same as 'still_alive' defensively; callers reconciling
+      // an old/stale row this runner instance never knew about should not be
+      // blocked forever by a status that never actually confirmed a live process.
       logger.debug(`[RUNNER RUN] Session ${sessionId} not found without verified exit`);
-      return 'still_alive';
+      return 'unknown';
     };
 
     // Handle child process exit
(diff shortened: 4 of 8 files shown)


heavygee:
Follow-up commit: a self-review (`/code-review`) caught that the initial version coerced *any* `stopRunnerSession` failure (including transient ack timeouts / protocol errors on an online machine) into `already_gone`, reproducing this PR's own bug one RPC layer down. Fixed by only treating the fallback as "already gone" when the machine itself isn't currently connected (nothing stronger to check); an online machine whose RPC call fails is now treated as `still_alive` (the same conservative default used elsewhere in this file for the identical RPC). Added a regression test for both branches.

heavygee:
Addressed the Major finding from both review rounds: added a fourth `'unknown'` status distinct from `'still_alive'` for the case where the runner has no PID match and no verified-exit tombstone for a session id at all (cli/src/runner/run.ts's fail-closed default). Threaded through the full runner<->hub RPC contract (run.ts, controlServer.ts, controlClient.ts, apiMachine.ts, rpcGateway.ts). `archiveSession`'s new fallback now treats `'unknown'` like the offline-machine case (best-effort archive — this is exactly the stale-row scenario the fallback needs to unblock); every other existing `stopRunnerSession` call site (fork-child cleanup, OpenCode reserved-clear, PTY/Pi resume termination + reconciliation) keeps its prior conservative behavior by treating `'unknown'` the same as `'still_alive'`, since those all deal with sessions the hub has fresh knowledge of.

Updated the pre-existing `runner.integration.test.ts` assertion for an unknown id (was `'still_alive'`, now correctly `'unknown'`) and added a 6th hub-side regression test. Ran the real CLI integration suite locally via `bun run test:cli:integration` (matches CI's script) — 13/14 pass, 1 skipped; the 4 tests that failed in this PR's CI run (spawn/stop via HTTP, track sessions, webhook metadata, concurrent ops) all pass locally with this branch, and I confirmed the identical 4-test failure signature already exists on `main` at this PR's base commit (unrelated pre-existing CI flake, not a regression from this change).

`tsc --noEmit` clean in both `hub/` and `cli/`.

heavygee:
Addressed the third Major finding: `MachineCache.active` is a 45s heartbeat heuristic (`machineCache.ts` expireInactive), not the same signal as whether the machine's RPC target is actually registered (that only clears on socket disconnect, `rpcRegistry.unregisterAll`). A delayed heartbeat could leave `stopRunnerSession` fully reachable while my `active !== true` pre-check skipped it and archived a live session anyway — the same bug at a third layer.

Dropped the machineCache pre-check entirely. Now it just attempts `stopRunnerSession` and discriminates on the failure itself: `RpcTargetMissingError` (no RPC target at all — genuinely nothing to check) falls back to archiving; any other failure (ack timeout, protocol error — machine answered the transport but this call didn't resolve) is treated as `still_alive`. Rewrote the regression tests to no longer touch machineCache at all, proving the decision doesn't depend on it, and to assert on the RpcTargetMissingError-vs-other-error distinction directly.

heavygee:
Addressed the fourth finding: treating `'unknown'` as archivable was itself unsafe. Terminal-started sessions are tracked only in-memory (`pidToTrackedSession`), never persisted, unlike runner-spawned children — so a still-alive terminal session can return `'unknown'` right after this exact runner process restarts and loses that in-memory tracking, indistinguishable from this RPC alone from a row this runner generation never knew about at all. That ambiguity can't be resolved from the RPC response alone, so `'unknown'` is now treated conservatively — the same as `'still_alive'` — matching every other `stopRunnerSession` call site in this file, exactly per the suggested fix.

The `'unknown'` type distinction stays (it's the runner's accurate contract and fixes the pre-existing integration test's assertion), it just no longer unblocks archival on its own — that stale-row cleanup case is #842's reconciler territory, not this fallback's job.

heavygee:
Production incident umbrella superseding/extending this path: https://github.com/tiann/hapi/issues/1910 (oos-linux 2026-09-23 unreaped driver CLIs). This PR remains relevant as a hub-side piece; the incident also needs runner orphan scan + webhook await kill. Please align or close in favor of #1910 work if stale.

PR #1708, opened 2026-08-28
feat(cli): auto-steer ping_peer messages into an active turn

KevinFun:
## Summary

`ping_peer` messages to a peer that is mid-turn were stored and forwarded immediately, but the peer CLI silently held them until the current turn ended — the nudge looked "sent" yet was ignored by the working agent. This PR delivers peer messages into a running turn automatically (no manual Steer click), while keeping the ordinary queue as the fallback everywhere else.

## Changes

- **`ping_peer` (MCP tool + `hapi ping-peer`)**: sends a generated `localId` (row becomes queue-shaped: visible in the peer's waiting bar, steerable, cancellable) and `deliveryMode: 'steer'` — both are existing protocol fields of `SendMessageRequestSchema`. Tool description documents the steer semantics.
- **Hub** (`messageService.getNormalizedDeliveryMode`): persists steer provenance for every steering-capable flavor via the shared `isSteeringSupportedForSession` predicate (previously pi-only). Claude and unknown flavors still downgrade to an ordinary queue row.
- **Codex CLI** (`codexRemoteLauncher`): the `SteerQueuedMessage` RPC handler is extracted verbatim into `steerQueuedByLocalId` (the web Steer button now delegates to it), and a queue-arrival hook injects a steer-tagged arrival into the active turn via app-server `turn/steer`. Every refusal path restores the queued row, so a refused auto-steer simply delivers at the next turn boundary.
- **Queue** (`MessageQueue2`): `push()` accepts an optional `steerHint` and the `onMessage` handler now receives the queued item.

## Behavior

- Peer idle → unchanged (message is consumed immediately as the next turn).
- Peer mid-turn on codex / pi → message is injected into the running turn; the chat shows the existing "steered" badge (`messages-consumed { steered: true }`). Pi gets this for free: its CLI already honors `meta.deliveryMode === 'steer'`.
- Peer mid-turn on claude / other flavors → unchanged queue-until-turn-end (hub downgrades, CLI unchanged).
- Web composer, Telegram, slash/control commands → unaffected: they never request steer, and the steer routine additionally refuses control commands and mode-mismatched rows.
- Wire protocol unchanged; no fixture-affecting web pipeline changes.

## Testing

- [x] `bun typecheck` (cli / hub / web / relay)
- [x] Full `bun run test` (cli + hub + web + shared + relay)
- New/updated tests:
  - hub: steer provenance stored for codex, downgraded for claude (and flavor-less sessions)
  - cli: `ping_peer` request body carries `localId` (`ping-peer-<uuid>`) + `deliveryMode: 'steer'`
  - cli: `MessageQueue2` passes `steerHint` through to the arrival handler; isolated/clear pushes stay unhinted

### DIFF
--- a/cli/src/codex/codexRemoteLauncher.ts
+++ b/cli/src/codex/codexRemoteLauncher.ts
@@ -2095,147 +2095,165 @@ class CodexRemoteLauncher extends RemoteLauncherBase {
             }
         };
 
-        // Per-message steer from the waiting queue (web "Steer" button).
+        // Per-message steer from the waiting queue (web "Steer" button). Also
+        // driven by the arrival hook below for steer-tagged peer messages.
+        const steerQueuedByLocalId = async (localId: string): Promise<{ steered: boolean; error?: string }> => {
+            if (!localId) {
+                return { steered: false, error: 'Missing localId' };
+            }
+            if (this.abortInProgress || !turnInFlight || !this.currentThreadId || !this.currentTurnId) {
+                return { steered: false, error: 'No active steerable turn' };
+            }
+            // Reserve before awaiting turn/steer so the main loop cannot
+            // collect the same row for turn/start while steer is in flight.
+            const taken = session.queue.takeByLocalId(localId);
+            if (!taken) {
+                return { steered: false, error: 'Message not in queue' };
+            }
+            // An explicit retry supersedes any old reconciliation poll for
+            // the same held row.
+            pendingSteerReconciliations.delete(localId);
+            const isControlCommand = Boolean(taken.item.isolate)
+                || Boolean(parseCodexSpecialCommand(taken.item.message).type);
+            if (isControlCommand) {
+                session.queue.restoreReservation(taken);
+                return { steered: false, error: 'Control commands cannot be steered' };
+            }
+            if (activeMessage?.hash !== taken.item.modeHash) {
+                session.queue.restoreReservation(taken);
+                return { steered: false, error: 'Queued message mode differs from the active turn' };
+            }
+            const batch: QueuedMessage = {
+                message: taken.item.message,
+                mode: taken.item.mode,
+                isolate: Boolean(taken.item.isolate),
+                hash: taken.item.modeHash
+            };
+            const steerEpoch = this.steerEpoch;
+            // Pin the thread this steer targets: reconciliation must look at
+            // the steer's thread, not whichever turn is current later.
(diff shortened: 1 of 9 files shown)


KevinFun:
Thanks for the review. Addressed the Major finding in 73630bbd:

- **Hub gate**: steer provenance is now persisted only for flavors whose CLI consumes the hint on arrival (`pi`, `codex`). Cursor ACP is downgraded to an ordinary queue row until its receive path is wired to the arrival hook — happy to land that wiring as a follow-up if maintainers want Cursor auto-steer.
- **Tool description**: dropped Cursor from the mid-turn injection claim (`codex and pi` only).
- **Regression test**: the hub test now asserts `cursor` + `deliveryMode: 'steer'` is stored as `queue`, alongside codex (steer) and claude (queue).

On the launcher-level test suggestion: `codexRemoteLauncher` has no test harness today (the class boots the app-server client, display, and process spawn), so that is a larger lift than this PR — noted as possible follow-up. The arrival hook is a thin guard (`steerHint` + `turnInFlight`) over the same `steerQueuedByLocalId` routine the existing RPC path uses, and the hint plumbing itself is unit-tested in `MessageQueue2`.

Re: the failing `integration` check — I ran the full runner integration suite locally on this exact branch: **13 passed | 1 skipped**, including the four failing cases (`should spawn & stop a session via HTTP`, `should track both runner-spawned and terminal sessions`, `should update session metadata when webhook is called`, `should handle concurrent session operations`). The diff touches no runner-spawn paths, so this looks like a CI-environment flake rather than a regression — happy to help dig if it reproduces on other PRs.

KevinFun:
Additional data point on the `integration` failure: it predates this PR and affects every recent one. The `test` workflow's last 10 runs (across #1706, #1707, #1695, and this PR) are all `failure`, and the root cause looks like #1696 — the CI runner has no local Claude install, so the runner integration suite cannot spawn real sessions. The same suite passes locally on this branch (13 passed | 1 skipped). Once #1696 or equivalent lands, this check should go green here without any change on my side.

heavygee:
Takeover continuation: https://github.com/tiann/hapi/pull/1825 (branch `heavygee:fix/1708-ping-peer-auto-steer`, rebased on current main).

Addresses the remaining Majors from this PR:
- Cursor ACP inbound auto-steer wiring
- Codex inherited-model first-turn mode-hash mismatch

Credit for the original design remains with @KevinFun — please treat #1825 as the active review surface.

Issue #1709, opened 2026-08-28
feat(workspace): add configurable HAPI Recycle Bin for deleted files


techotaku39:
## Problem / Motivation

HAPI's session Files → Directories view and file preview currently support browsing, reading, copying paths, and downloading files, but they do not provide a recoverable delete action.

The existing `upload/delete` endpoint only removes temporary uploaded attachments; it is not intended for workspace files. When using HAPI remotely from a phone or browser, deleting a file requires returning to the terminal, and permanent deletion creates an unnecessary risk of accidental data loss.

## Proposal

Add a machine-local HAPI Recycle Bin, based on the Recycle Bin design described in #1176.

Normal file deletion should move the item into the HAPI Recycle Bin instead of permanently removing it. The default retention period should be 30 days and configurable through the HAPI configuration file.

The Recycle Bin should be a virtual HAPI-managed surface, not a normal directory exposed inside the workspace tree.

## Initial scope

The initial implementation focuses on moving individual files from the file preview into a machine-local HAPI Recycle Bin, then listing, restoring, and permanently deleting them.

Directory deletion, batch operations, and general copy/move workflows are follow-up features and are not part of this implementation. The implementation may introduce reusable Runner-side file-operation primitives so these future features can build on the same authorization, locking, and task model.

## Proposed UX

- Add a `Move to Recycle Bin` action to the file preview toolbar.
- The trash icon in the file preview should perform the move action; it should not navigate to the Recycle Bin.
- Add a separate `Recycle Bin` button near the file browser toolbar, next to actions such as search, sorting, and refresh.
- The Recycle Bin button opens a machine-scoped Recycle Bin view and may display the number of deleted items.
- Before moving an item, show a confirmation dialog containing:
  - the full original path;
  - the machine/workspace scope;
  - the effective retention period;
  - an explicit statement that the item can be restored before expiration.
- After a successful move, return to the file list or preview origin and refresh directory, search, and Git status data.
- The operation must not automatically stage, commit, or otherwise modify Git history.

## Recycle Bin behavior

Each recycle entry should retain enough metadata to identify and restore the item:

- original authorized path;
- deletion time;
- file type;
- size;
- unique recycle entry ID;
- expiration time.

The Recycle Bin should support:

- listing deleted entries;
- read-only preview where practical;
- restoring to the original authorized path;
- restoring to another authorized destination;
- explicit conflict handling when the original path is occupied;
- permanently deleting a selected entry;
[...]

Issue #1176, opened 2026-07-26
Spec: 完成 HAPI Git、文件与 Managed Provider Profile 管理闭环


zulinliu:
## Problem Statement

HAPI 用户希望在 iPhone Safari/PWA、Telegram Mini App 和桌面浏览器中，安全地管理机器上的 Workspace Root、Session Repository 和 Managed Provider Profile，而不必返回本地终端完成收尾。目前能力分散在 Workspace、Session、Runner 和设置页面：部分 Git 操作已存在但语义与入口不统一；长耗时操作的进度依赖页面本地状态，导航、刷新或断线后难以追踪；文件上传、下载、复制、移动、编辑和删除缺少批量、恢复、冲突处理与移动端闭环；Provider 检测、迁移、默认项和会话恢复也缺少完整安全流程。

这不仅是“补按钮”问题。远程主机操作必须同时解决授权边界、并发写入、状态过期、资源锁、任务恢复、敏感信息保护、破坏性操作确认，以及 iPhone 键盘和窄屏可用性。否则用户可能在 Agent 活跃、仓库状态已变化、目标文件已变化或凭证已失效时执行不可逆操作，却得不到可靠反馈与恢复路径。

## Solution

建立统一的 HAPI 管理控制体系：

- Workspace Root 与 Session Repository 使用同一套 Repository 管理模型和 Repository Panel，但由不同作用域适配器保持各自授权边界。
- 所有长耗时机器操作统一为持久化 Machine Task，跨导航、刷新和重连可见，并提供进度、取消、重试、检查点和脱敏结果。
- 通过 Repository Activity Guard、Repository Snapshot 状态令牌和仓库写锁，检测 Agent/用户并发写入、拒绝过期操作并串行化 Git Mutation。
- 文件管理补齐批量分块上传、流式 ZIP 下载、事务化复制/移动、只读预览与显式编辑、冲突恢复、DestinationPicker 和 HAPI Recycle Bin。
- Managed Provider Profile 补齐草稿检测、分层健康诊断、显式默认、导入导出、加密备份、一次性明文复制和会话恢复替代流程，同时保持只对 HAPI 启动的 Agent 生效。
- 所有复杂流程以 iPhone Safari/PWA 为首要验收环境，并兼容 Telegram Mini App 和桌面浏览器。

## User Stories

1. As a HAPI user, I want long-running machine operations to remain visible after navigation, so that I do not lose progress or results.
[...]

PR #1712, opened 2026-08-29
feat(workspace): add HAPI Recycle Bin for deleted files

techotaku39:
## Summary

- Add a Runner-local, configurable HAPI Recycle Bin for workspace files.
- Add recoverable delete, preview, restore, permanent delete, and empty-bin operations.
- Add authenticated RPC and REST contracts between Web, Hub, and CLI.
- Add Recycle Bin actions to the file preview and Files toolbar.
- Keep recoverable deletion separate from the permanent-delete proposal in #1092.
- Pin destructive filesystem mutations to validated directory identities on POSIX and Windows.

## Problem / Motivation

The workspace file browser can preview, copy, and download files, but it does not provide a recoverable delete action. Remote users must return to the terminal to remove files, while permanent deletion makes accidental data loss unnecessarily difficult to recover from.

This PR adds a local-first Recycle Bin so deleted workspace files can be restored before expiration without persisting file contents in the Hub.

## Implementation

- Store each entry under the Runner's local `$HAPI_HOME/recycle-bin`.
- Store metadata and payload separately, protected by a file lock.
- Record the original path, deletion time, size, file type, expiration time, and content hash.
- Support:
  - moving one file to the Recycle Bin;
  - listing entries visible to the authorized session workspace;
  - read-only preview for bounded-size entries;
  - restoring to the original authorized path;
  - cancel, fail, and restore-with-new-name conflict handling;
  - permanently deleting one entry;
  - emptying visible entries after confirmation.
- Overwrite restore is deliberately disabled for this PR. Requests return `overwrite_unavailable` and leave the existing target unchanged until an atomic identity-bound replacement can be implemented.
- Automatically remove expired entries when the Recycle Bin is accessed.
- Add Web UI actions and English/Chinese localization.
- Do not automatically stage, commit, or modify Git history.

## Configuration

Set the Runner-local retention period in `$HAPI_HOME/settings.json`:

~~~json
{
  "recycleBinRetentionDays": 30
}
~~~

The default is 30 days. Valid values are 1–3650 days. The effective retention period is read by the Runner and stored in each entry's expiration metadata when the entry is created.

## Security and data safety

- Require an active authorized session for Recycle Bin REST operations.
- Bind file operations to the session working directory.
- Canonicalize and revalidate paths at use time.
- Reject out-of-scope paths, symlink escapes, and `.git` paths.
- Use POSIX directory-handle-relative operations and Windows directory handles with native file-information calls for source detach, restore publication, and staging cleanup.
- Preserve target identity checks for non-destructive restore paths; never overwrite an existing target in this PR.
- Protect Recycle Bin storage with a lock and exclusive file operations.
- Verify payload integrity with a SHA-256 content hash.
- Keep file contents on the Runner; the Hub only forwards metadata and RPC results.

## Scope and follow-ups

This implementation supports individual files and restores them to their original authorized paths.

Atomic identity-bound overwrite restore, arbitrary destination restore, directory deletion, batch operations, general copy/move workflows, and a full destination picker remain follow-up work.

The permanent-delete behavior proposed in #1092 is not modified or replaced. The Recycle Bin's purge and empty actions are separate, explicit, irreversible operations for entries already inside the Recycle Bin.

## Validation

- `bun typecheck` using Bun 1.4.0 — passed.
- `bun run build` using Bun 1.4.0 — passed.
- CLI Recycle Bin, secure filesystem, and handler tests — 49 passed; native FFI tests are covered by separate Bun smoke tests.
- Web Recycle Bin dialog tests — 8 passed.
- Native Bun filesystem smoke tests for directory-relative move, restore, overwrite rejection, and staging cleanup — passed.
- Full real Bun move, restore, and safe conflict flows — passed.

## Related Issues

- Refs #1709 — configurable HAPI Recycle Bin for deleted workspace files.
- Refs #1176 — HAPI Recycle Bin design.
- Refs #1092 — separate proposal for permanent delete, move, and copy operations.

## AI Assistance

Implemented and reviewed with OpenAI Codex using GPT-5.6. Tests, builds, integration checks, and the final diff review were performed locally.


gavinc:
Does it happen a lot that files are getting deleted accidentally?

techotaku39:
Thanks for asking. I don’t have telemetry showing that accidental deletions happen frequently. The goal of this PR is not to address a high-volume problem, but to provide a safety net for occasional mistakes, especially when deleting files from the workspace file preview.

The current deletion behavior is irreversible, while recovering a mistakenly deleted file can be much more costly than the deletion itself. The Recycle Bin adds an explicit, configurable recovery window—30 days by default—while keeping the existing permanent-deletion behavior unchanged.

PR #1715, opened 2026-08-29
fix(pi,agents): tool-result images + provider-qualified set-session-config + live model catalogs

HongChenGG:
## Summary
Fixes broken inline model switching from the web UI and keeps in-session model catalogs in sync with the live agent config. Includes the earlier `fix(pi): render images returned in pi tool results` commits.

### 1. set-session-config used the wrong wire format per agent (the crash)
The hub normalizes a picked model into a `{provider, modelId}` object, but every agent handler collapsed it to a bare modelId:
- **opencode** requires the provider-qualified `provider/modelId` wire format (its ACP resolver treats a bare id as a provider name → `model not found`): added `modelProviderQualified` option;
- **claude / codex / kimi / copilot** handlers crashed on the object payload (TypeError / unhandled throw): extract the bare modelId they expect;
- **hub session cache** now normalizes per flavor (opencode keeps the qualified id, others the bare id);
- a failed inline switch rolls `sessionModel` back via `onModelRollback` so keepalives and the hub display converge on the model the backend actually kept.

### 2. In-session model catalogs drifted from the create-session form
The session picker served frozen startup snapshots while the create-session form probes live config:
- **pi**: `ListPiModels` re-queries the running process on every request (cache kept only as fallback);
- **opencode / grok / copilot**: probe live (shared with the create-session form, 60s cache) instead of serving the frozen `session/new` ACP snapshot; `currentModelId` stays session-scoped so inline switches / rollbacks display accurately;
- **web**: in-session model pickers refresh every 15s.

### Note on conflicts with current main
Main gained `onModelRollback` wiring (d4fe6b9..) that overlaps `cli/src/opencode/opencodeRemoteLauncher.ts` / `runOpencode.ts` here — the merge resolution is to keep upstream's `rollbackModel` (`model ?? undefined`) plus this branch's `modelProviderQualified`, probe fallback, and comments. Happy to rebase if that's easier for review.

### DIFF
--- a/cli/src/pi/loop.ts
+++ b/cli/src/pi/loop.ts
@@ -2,11 +2,11 @@ import { logger } from '@/ui/logger';
 import { convertAgentMessage } from '@/agent/messageConverter';
 import { createNativeSessionTitleMetadataSync } from '@/agent/nativeSessionTitle';
 import { PiTransport } from './piTransport';
-import { convertPiEvent, convertPiTurnUsage } from './piEventConverter';
+import { convertPiEvent, convertPiTurnUsage, extractPiGeneratedImages } from './piEventConverter';
 import { PiMessageAccumulator } from './piMessageAccumulator';
 import { PiExtensionUiHandler } from './extensionUiHandler';
 import { parsePiModels, parsePiCommands, parsePiContextUsage, PiAgentEndEventSchema, PiAgentSettledEventSchema, PiExtensionUiRequestSchema, PiLifecycleEventSchema, PiResponseEventSchema, PiSessionInfoChangedEventSchema, PiStateDataSchema, PiSetModelDataSchema } from './schemas';
-import type { PiContextUsage, PiResponseEvent, PiRpcCommand, PiThinkingLevel, PiTurnEndEvent } from './types';
+import type { PiContextUsage, PiResponseEvent, PiRpcCommand, PiThinkingLevel, PiToolExecutionStartEvent, PiTurnEndEvent } from './types';
 import type { PiSession } from './session';
 import type { PiConversationHistory } from './conversationHistory';
 
@@ -533,6 +533,9 @@ export function wireTransportEvents(
     const lifecycleTimeline = new PiLifecycleTimeline();
     let latestContextUsageRequest = 0;
     let deliveredSettlement = false;
+    // Bounded tool-args cache so tool_execution_end can attribute registered
+    // media (e.g. images read by the `read` tool) to their source path.
+    const piToolArgsByCallId = new Map<string, unknown>();
     let legacySettleTimer: ReturnType<typeof setTimeout> | null = null;
     let promptLifecycleTimer: ReturnType<typeof setTimeout> | null = null;
     let compactionRetryTimer: ReturnType<typeof setTimeout> | null = null;
@@ -849,6 +852,19 @@ export function wireTransportEvents(
             for (const message of messages) {
                 const converted = convertAgentMessage(message, session.currentModel);
                 if (converted) session.sendAgentMessage(converted);
+            }
+            if (event.type === 'tool_execution_start') {
+                if (piToolArgsByCallId.size > 500) piToolArgsByCallId.clear();
+                const startEvent = event as PiToolExecutionStartEvent;
+                piToolArgsByCallId.set(startEvent.toolCallId, startEvent.args);
+            } else if (event.type === 'tool_execution_end') {
+                const endEvent = event as { toolCallId: string };
+                const mediaMessages = extractPiGeneratedImages(event, (toolCallId) => piToolArgsByCallId.get(toolCallId));
+                piToolArgsByCallId.delete(endEvent.toolCallId);
+                for (const message of mediaMessages) {
+                    const converted = convertAgentMessage(message, session.currentModel);
+                    if (converted) session.sendAgentMessage(converted);
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


HongChenGG:
Rebased onto current `main` (`10042f79`) — thanks for the merge-resolution note, it was accurate.

**Conflict resolutions (3 hunks, 2 files):**

1. **`rollbackModel`** — kept upstream's version exactly (`batch.mode.model = model ?? undefined`); this branch's earlier variant and its doc comment are gone (the `d6dccf10`/`caa204b4` pair already converged there).
2. **Duplicate `onModelRollback` / `onReasoningEffortRollback` wiring** — dropped, because current `main` already has both the `OpencodeLoopOptions`/`OpencodeRemoteLauncherOptions` declarations and the `runRemote` pass-through. `cli/src/opencode/loop.ts` therefore **no longer appears in the PR diff**; the rest of the opencode probe/fallback work is unchanged.
3. Fixed a stale comment above the opencode probe: it still claimed the code "falls back whenever it did not produce a non-empty catalog", but `dce83174`/`0ab2281d` deliberately treats a *successful* empty catalog as authoritative. Comment now matches the code.

**Re-verified on the rebased head:**
- `bun run typecheck` — cli / web / hub / relay all clean
- `bun run test` — cli 234 files / 2628 passed, hub 1225 passed, web 2947 passed, shared 299 passed, relay 88 passed
- Diff is now 22 files, +511/−50 (`cli/src/opencode/loop.ts` dropped as above).

HongChenGG:
Both findings fixed in `df77180c` (on top of the rebase in `0ab2281d`).

**1. [Major] `runPi.ts:858` — polling reapplied the startup model.** Correct, and this is exactly what the new 15s poll made severe: `ListPiModels` now re-queries `get_available_models`, so its response re-entered the startup `set_model` branch on every poll and reverted an inline switch to the launch model. Fixed by making startup-model application a **one-shot bootstrap**, not a per-response action:
- `PiSession.initialModel` is consumed on the first response that can act on it, and the branch is additionally gated on `session.explicitModelSelection` (set by the `SetSessionConfig` and `/model` paths), so a user selection always wins.
- The gate is re-checked *inside* the `runRuntimeMutation` lease, because the outer check runs before a queued user switch is granted the lease.

These are the three commits from #1815 (`fix(pi): make user model selection sticky` + tests), cherry-picked here, since #1715 is what makes the bug reachable every 15 seconds. **#1815 is the standalone version of the same guard against current `main`** — if you merge this PR first, #1815 can be closed as redundant (otherwise they will conflict on `cli/src/pi/loop.ts`; tell me which one you'd rather keep and I'll drop the other).

**2. [Minor] `runPi.ts:864` — dedup ran after the metadata write.** Fixed as suggested: the cache write and the change-detected `updateMetadata` now happen in one place in `handleResponse`, for every response *including successful empty catalogs*, and the dead duplicate in the `ListPiModels` handler is gone. Added a regression test that asserts two identical polls version the metadata once and that an empty catalog clears the cache and versions it once (fails before the fix: 2 calls, passes after: 1).

**Tests** — `cli`: typecheck clean, 234 files / 2635 passed / 2 skipped. `web`/`hub` are untouched by these commits and were green on `0ab2281d`.

**Still untested** — the bot's third coverage note (Pi tool-result image delivery through real transport) is not unit-testable in this suite: `piEventConverter` is mock-converted in `loop.test.ts` and the whole suite mocks Pi. That path is exercised by the integration job.

HongChenGG:
Decision on the overlap mentioned earlier: **#1815 is closed as redundant**, its three commits live here as `3b2911ca`/`a6299f08`/`ce3fb321`, so there is nothing left to arbitrate — just merge this one.

The branch is preserved at `HongChenGG/hapi:fix/pi-sticky-user-model` in case you'd rather take the small standalone fix first; then I'll rebase this PR instead.

HongChenGG:
### Re: "image test does not establish end-to-end hub/client rendering"

Acknowledging the review note on 4e45e84/eeab6570 — replying with evidence rather than more mocks, because we believe the residual gap it describes is already closed layer-by-layer, and the Pi path deliberately adds **no Pi-specific hub or client code**:

1. **Pi-side registration & delivery** (`cli/src/pi/piEventConverter.ts:118` → `wireTransportEvents`): the new `piGeneratedImageDelivery.test.ts` exercises the *real* converter and the *real* generated-image registry (only logger/accumulator are stubbed), asserting the `generated_image` AgentMessage on `sendAgentMessage` and byte-exact retrieval via `getGeneratedImage`.
2. **AgentMessage → wire block** (`cli/src/agent/messageConverter.ts:123` `generated_image → generated-image` with imageId/fileName/mimeType/source): covered by the pre-existing `messageConverter.test.ts:243` — unchanged by this PR, which is precisely why the delivery test mocks it.
3. **Hub**: the message flow is type-agnostic (no `generated_image` branching in `hub/src/`); retrieval is the established route `GET /sessions/:id/generated-images/:imageId` (`hub/src/web/routes/git.ts:161`) → `syncEngine.readGeneratedImage` → RPC `ReadGeneratedImage`. Nothing Pi-specific to test there.
4. **CLI retrieval handler** (`cli/src/modules/common/handlers/files.ts:62`) reads from the same registry instance the delivery test asserts against — the byte comparison in step 1 is what that handler would serve.
5. **Web**: renders `generated-image` blocks without distinguishing flavors (`ToolMessage` aligns only on user/assistant source), already covered by `ToolMessage.generatedMedia.test.tsx` (labels, lazy fetch, blob retrieval) — shared with the Codex flow this mirrors.

So every Pi-specific new behavior has direct coverage; each shared hop keeps its own layer's test; what genuinely remains unverified is real-agent runtime rendering, which we consider out of scope for unit tests in this PR. Happy to add an integration fixture if a maintainer wants it.

PR #1719, opened 2026-08-30
feat(pi): auto-title Pi sessions via bundled hapi_change_title extension

HongChenGG:
## Problem

Pi sessions in HAPI never get automatic titles:

- HAPI launches Pi with `PI_RPC_EMIT_TITLE=1`, but Pi does not implement that env var (verified against pi 0.84.2 and the latest 0.84.4) — it is a no-op.
- Unlike the Claude/Codex/OpenCode launchers, the Pi bridge neither registers a `change_title` tool nor injects a title instruction, so nothing ever sets the session title. Sessions fall back to path/id names forever.

Related: #1669 (hub-side auto title-provider approach), #1672 (change_title availability for other launcher types).

## Approach

Give Pi sessions the same treatment the other launchers already get — a tool plus an instruction — using Pi's own extension mechanism:

1. `titleExtension.ts` embeds a small Pi extension source (plain JS as a string) and materializes it under `<happyHomeDir>/runtime/<version>/pi/hapi-title-extension.ts` at launch, then passes it to Pi via `--extension` (available since pi 0.35.0).
2. The extension registers a namespaced `hapi_change_title` tool (naming mirrors the OpenCode launcher's `hapi_change_title`, so user extensions registering their own tools cannot collide) and injects a first-turn system-prompt instruction with the same wording used for Claude/Codex.
3. The model calls the tool; the extension applies the title via `ctx.ui.setTitle()`, which flows through the existing Pi extension UI bridge (`extensionUiHandler` → `syncTitle`) into session metadata. No extra title-provider round-trip and no changes to the title-suggestion service.

Also removes the dead `PI_RPC_EMIT_TITLE` env var.

## Behavior notes

- Titles are generated by the session's own model, so there is no additional provider call (vs. the hub-side approach discussed in #1669). Both approaches can coexist; this one needs no configuration.
- Titles appear once the objective is clear (first real task message), matching the other launchers' semantics; pure greetings do not trigger naming.
- Requires pi >= 0.35.0 (when `--extension` landed, 2026-01).

## Testing

- Unit: new `titleExtension.test.ts` (source sanity + materialization idempotency); updated `runPi.test.ts` startup assertions (transport args now include `--extension <path>`).
- `bun run typecheck` passes; `vitest run src/pi/titleExtension.test.ts src/pi/runPi.test.ts` → 61 passed.
- Manual end-to-end with a real Pi RPC session: the model called the tool on its first turn and the `setTitle` extension UI request reached the hub, which persisted it as the session title.

## AI-generated code disclosure

Per the Vibe Coding policy: code authored with AI assistance (GLM-5.3-flash via the pi coding agent), reviewed and tested as described above.


HongChenGG:
Both findings are addressed in ec6d2797:

**[Major] atomic publication** — `materializePiTitleExtension` now writes to a unique temp file (`<file>.<pid>.<uuid>.tmp`) and moves it into place with `rename`, so a concurrent launcher can never observe a half-written extension. If `rename` loses a race against a launcher that already published the file, the existing copy is kept and the temp file is cleaned up.

**[Minor] persistent instruction** — the `titled` short-circuit is removed; `before_agent_start` now injects the instruction on every turn, preserving the objective-change retitle rule from the persistent Claude/Codex instruction.

Tests added per the review's request:
- concurrent materialization: 8 parallel launches must all resolve to the same path with complete content and leave no `.tmp` files behind;
- handler test that loads the materialized source as a module, executes the title tool, and asserts a later turn still carries the instruction (`Rename only when...`).

Full suite: 63 passed, `bun run typecheck` clean.

HongChenGG:
Addressed in eb5d0f02 — the extension source is now backward-compatible with the Pi releases HAPI accepts:

- **TypeBox specifier**: imports `Type` from `'@sinclair/typebox'`, which the legacy loaders expose and the current loader also resolves (it aliases both `'typebox'` and `'@sinclair/typebox'` to its bundled module — verified in pi 0.84.2 `dist/core/extensions/loader.js`).
- **execute() argument order**: the context is normalized positionally (`fifth` if it carries the `ui` bridge, else `fourth`), covering both current `(id, params, signal, onUpdate, ctx)` and legacy `(id, params, onUpdate, ctx, signal)`.
- **hasUI guard removed**: `ctx.ui.setTitle()` is now called directly (guarded only against a missing `ctx.ui`), since this extension only runs under HAPI's RPC bridge and older RPC releases report `hasUI: false` even though `setTitle` emits the event.

Tests added per the review request: both call orders are exercised, including `hasUI: false`, asserting `setTitle` fires in each case. Also verified on a live pi 0.84.2 RPC session that the extension loads cleanly with the `'@sinclair/typebox'` specifier. Suite: 64 passed, `bun run typecheck` clean.

PR #1722, opened 2026-08-30
fix(web): allow retrying failed attachment uploads

techotaku39:
## Summary

- Add an inline refresh-style retry control for failed composer attachments.
- Reuse the existing attachment control geometry while hiding the failed-state text and preserving the upstream drag/remove layout.
- Create a fresh `File` for retries so restored draft attachments receive a new upload ID.
- Delegate retry eligibility to the active attachment adapter so deterministic size, MIME, and quota failures do not expose a retry action while transient failures remain retryable.
- Preserve the original attachment position as soon as the replacement appears, including during the upload.
- Balance the remove-button spacing when a long filename is truncated, including when an attachment transitions into the error layout.
- Keep non-retryable failures accessible with a localized hidden failure label.
- Add localized accessible labels and regression coverage for the retry flow, adapter behavior, ordering, and layout.
- No API, database, dependency, or migration changes.

## Validation

- `bun typecheck` — passed.
- `pwsh -NoProfile -File .\scripts\Invoke-HapiTaskPlaywright.ps1 -Name attachment-upload-retry -Suite Root -TestArgs 'terminal-wrap-fidelity.spec.ts'` — 2 passed.
- `bun --cwd web test src/components/AssistantChat/AttachmentItem.test.tsx src/components/AssistantChat/SortableComposerAttachments.test.tsx src/lib/attachmentAdapter.test.ts src/lib/scratchlistAttachmentAdapter.test.ts src/lib/attachmentOrder.test.ts` — 42 passed.
- `bun run build` — passed.

## Related Issues

None

## AI Assistance

OpenAI Codex with GPT-5.6 assisted with implementation, testing, and validation.

### DIFF
--- a/web/src/components/AssistantChat/AttachmentItem.test.tsx
+++ b/web/src/components/AssistantChat/AttachmentItem.test.tsx
@@ -12,6 +12,7 @@ const mocks = vi.hoisted(() => ({
     composer: {
         addAttachment: vi.fn(async (_file: File) => {}),
         getState: vi.fn(() => ({ attachments: [] })),
+        subscribe: vi.fn((_listener: () => void) => () => {}),
     },
     attachmentRuntime: {
         remove: vi.fn(async () => {}),
@@ -41,6 +42,7 @@ beforeEach(() => {
     mocks.composer.addAttachment.mockClear()
     mocks.composer.getState.mockReset()
     mocks.composer.getState.mockReturnValue({ attachments: [] })
+    mocks.composer.subscribe.mockClear()
     mocks.attachmentRuntime.remove.mockClear()
 })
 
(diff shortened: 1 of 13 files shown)


techotaku39:
Addressed in 48cde1da:

- Clone the `File` before removing the failed attachment so restored draft metadata cannot reuse a cancelled attachment ID.
- Re-run truncation measurement when the error layout mounts.
- Hide Retry for deterministic oversized-file failures and retain the error indicator.
- Added component and adapter regression coverage for all three cases.

The updated targeted tests, typecheck, build, and terminal Playwright check pass.

techotaku39:
Addressed in 478adfca:

- Retry eligibility now comes from the active attachment adapter instead of a hard-coded chat upload limit.
- Normal and scratchlist adapters mark deterministic size/MIME/quota failures as non-retryable while leaving transient failures retryable.
- Added adapter and component regression coverage for deterministic and transient failures.

Validation on the new HEAD: 32 targeted tests passed, `bun typecheck` passed, and `bun run build` passed.

techotaku39:
Addressed in 8ee0e98d:

- Preserve the failed attachment's original index when replacing it with the fresh retry attachment ID.
- Thread the shared attachment-order ref through the sortable attachment components.
- Added regression coverage for the replacement helper and retry callback.

Validation on the new HEAD: 42 targeted tests passed, `bun typecheck` passed, and `bun run build` passed.

techotaku39:
Addressed in 694c7e0c:

- Subscribe to composer changes so the replacement ID is moved into the original order slot as soon as its first pending state appears.
- Avoid late order callbacks overwriting a reorder made while the upload is in flight.
- Add a localized screen-reader-only “Upload failed” label for non-retryable failures.

Validation on the new HEAD: 42 targeted tests passed, `bun typecheck` passed, and `bun run build` passed.

PR #1741, opened 2026-09-01
fix(web): keep streamed reasoning/text block ids stable across snapshot rows

HongChenGG:
## Problem

While a pi/codex session streams, the web UI reasoning panel re-renders from scratch on every snapshot tick (~every 250ms): the panel's open/collapse animation replays and the smooth typewriter restarts, which looks like constant flashing for the entire duration of the thinking phase.

### Root cause

The chain is:

1. CLI emits full-text snapshots for one reasoning stream every ~250ms, each under the **same stream id** but as separate message rows (`pi-…-reasoning-N` / `rs_…`).
2. The hub retires older live snapshot rows, and the web window store (`dropSupersededReasoningSnapshots`) keeps only the newest row per stream — so the timeline reducer sees a **different underlying message row on every pass**.
3. `reducerTimeline.ts` folded snapshots into a single block by stream id (good), but derived the block `id` from whichever row was first encountered: `` id: `${msg.id}:${idx}` ``.
4. So the block id — and the `threadMessageId` (`agent-reasoning:<id>`) built from it — **changed on every snapshot**. The React key changed, the whole reasoning message + collapsible group remounted, the open animation (`max-h`/`opacity` transition) replayed and the smooth streaming restarted from empty text. Every 250ms.

## Fix

When a content block carries a `streamId` (unique per stream, stable across its snapshot rows), use it as the block id; keep `` `${msg.id}:${idx}` `` as the fallback for content without a stream id. Applied to both `agent-reasoning` and `agent-text` (text has the same churn with `streamSnapshot` rows).

With a stable id, the timeline/assistant-ui pipeline updates the block in place: no remount, no animation replay, and the smooth typewriter appends only the delta — which is the desired streaming UX.

This also makes block ids more stable across history reloads: after a reload the surviving snapshot row is the newest one, so a row-derived id would differ from the id seen live, while the stream id does not change.

## Testing

- Updated the two stream-collapse tests in `reducerTimeline.test.ts` for the new id shape.
- Added a regression test: two reduce passes over different snapshot rows of the same stream produce the same block id (this mirrors the production window-retirement flow).
- Reran `bun run gen:fixtures` — only the two fixtures carrying streamed reasoning/text change (id field only).
- `cd web && vitest run` — 2860 tests pass (incl. the refreshed golden fixtures); `tsc --noEmit` clean.

## AI disclosure (per CONTRIBUTING)

Code authored with AI assistance (GLM, Z.ai) under human direction, based on a root-cause debugging session of the flashing reasoning panel.

### DIFF
--- a/android/core/protocol/src/main/kotlin/app/hapi/protocol/chat/ReducerTimeline.kt
+++ b/android/core/protocol/src/main/kotlin/app/hapi/protocol/chat/ReducerTimeline.kt
@@ -25,6 +25,14 @@ class TimelineResult(
 
 private fun getEventString(event: JsonObject, key: String): String? = asString(event[key])
 
+/**
+ * Stream identity must match the wire-level semantics in shared/src/messages.ts
+ * (blank ids are not streams): a blank value falls back to row-derived ids
+ * instead of colliding every blank-id row onto one block identity.
+ */
+private fun nonBlankStreamId(value: String?): String? =
+    value?.takeIf { it.isNotBlank() }
+
 private fun getEventNumber(event: JsonObject, key: String): Double? = asNumber(event[key])
 
 private fun getAgentRunStartedAt(event: JsonObject): Double? =
@@ -775,7 +783,7 @@ fun reduceTimeline(
                             )
                             continue
                         }
-                        val streamId = c.streamId
+                        val streamId = nonBlankStreamId(c.streamId)
                         if (streamId != null) {
                             val existing = textBlocksByStreamId[streamId]
                             if (existing != null) {
@@ -825,7 +833,7 @@ fun reduceTimeline(
                     }
 
                     is NormalizedAgentContent.Reasoning -> {
-                        val streamId = c.streamId
+                        val streamId = nonBlankStreamId(c.streamId)
                         if (streamId != null) {
                             val existing = reasoningBlocksByStreamId[streamId]
                             if (existing != null) {
--- a/android/core/protocol/src/main/kotlin/app/hapi/protocol/chat/ReducerTimeline.kt
+++ b/android/core/protocol/src/main/kotlin/app/hapi/protocol/chat/ReducerTimeline.kt
@@ -789,7 +789,11 @@ fun reduceTimeline(
                         }
 
                         val block = AgentTextBlock(
-                            id = "${msg.id}:$idx",
+                            // Stream-stable id (mirrors web/src/chat/reducerTimeline.ts):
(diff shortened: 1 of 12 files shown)


HongChenGG:
The CI failure in `ios / package-tests` was the native golden-fixture conformance suites: the Swift (HapiKit) and Kotlin (protocol) chat pipelines are ports of the web `reducerTimeline` and project the same fixtures, so they still emitted row-derived block ids for streamed reasoning/text.

Pushed ec19c6f6 to mirror the streamId-first id derivation in both ports (same fallback for content without a stream id). Local Swift/Android toolchains aren't available here, so relying on CI to confirm — the change is a strict mirror of the web-side logic.

HongChenGG:
Addressed the review finding in 96c9fc05: blank (`''` / whitespace-only) stream ids are now rejected before being used as block identity, in all three pipelines (web, Swift, Kotlin) — matching the wire semantics of `readReasoningStreamId` (trim before accept). Blank-id rows keep their distinct row-derived ids instead of colliding onto one empty identity. Added a web regression test covering both empty and whitespace-only ids.

HongChenGG:
Addressed in c295b7e6: the Swift text/reasoning branches now hoist `nonBlankStreamId(...)` into a local and reuse it for lookup, block identity, **and** map insertion (construction previously re-read the raw optional, so blank ids slipped through as a shared blank identity). Also added `ChatStreamIdentityTests.swift` covering stream-id derivation for both block kinds and the blank/`""`-vs-whitespace fallbacks.

HongChenGG:
Addressed in 852d0b93, covering both the finding and the residual test gap in one go:

- Updated the two stale fixture descriptions: the stream-snapshot case now states blocks are keyed by the stable stream id (`data.id`), and the reasoning case now states the stream id **is** the projected id.
- Added a generated fixture `codex-blank-stream-ids.json` (empty + whitespace-only `data.id`, both reasoning and text): pins the fallback contract — blank ids are not stream identities, each payload keeps its own row-derived block id — as conformance data shared by web, iOS, **and Android** (closing the Android coverage gap).

`bun run gen:fixtures` rerun; web suite 2864 tests green.

HongChenGG:
Addressed in HEAD push: added `HAPI_TITLE_PROVIDER_MAX_TOKENS` / `HAPI_TITLE_PROVIDER_TIMEOUT_MS` to the title-provider configuration table in `docs/guide/installation.md`, and extended the provider tests to directly exercise the timeout abort path (in-flight request rejects when the configured signal fires).

PR #1753, opened 2026-09-02
fix(runner): reject a spawn model missing from the machine's catalog

### LINKED ISSUE #1752: bug(cli): bad --model on spawn (e.g. cursor gpt-5) archives peer silently; wrapper OK is remit-only
## Symptom

`hapi spawn-peer` / `hapi-spawn-peer` with `--agent cursor --model gpt-5` creates the session, delivers the remit, prints success — and the agent **never runs**. The session goes inactive/archived with the remit as the only message and zero agent output.

The wrapper's success check is misleading:

```
hapi-spawn-peer: OK <sessionId>  name="…"  messages>=1
```

`messages>=1` is true because the *remit* landed. It does not assert that the agent started or produced output.

## Reproduction (operator fleet)

Failed twice (~90 min apart), different session ids, both with `--model gpt-5`:

```bash
hapi-spawn-peer --dir <path> --name "…" \
  --agent cursor --model gpt-5 --session-type worktree --message-file brief.md
hapi-spawn-peer --dir <path> --name "…" \
  --agent cursor --model gpt-5 --session-type simple --message-file brief.md
```

Control — identical command with `--model` omitted — succeeds (`lifecycle`/active running, real agent output).

Isolated variable: `--model gpt-5`.

Verified on hub after the fact: failed sessions store `model: "gpt-5"` and `active: false`; control has `model: null` and `active: true`.

## Why it is confusing rather than simply wrong

`cursor-agent --help` documents `gpt-5` as an example model string:

```
--model <model>   Model to use (e.g., gpt-5, sonnet-4-thinking).
```

But on the same host:

- `cursor-agent --list-models` has **no** bare `gpt-5` (current ids look like `gpt-5.5`, `gpt-5.6-sol-high`, …)
- `GET /api/machines/:id/cursor-models` also has no bare `gpt-5` (e.g. `gpt-5.5[fast=…]`, `gpt-5.6-sol[fast=…]`)

So an operator following the CLI help example picks a string the help advertises; spawn accepts any string (`SpawnSessionRequestSchema.model` is optional unconstrained `z.string()`), and the runner/agent dies silently — no loud spawn-time error.

## Expected

1. **Fail closed at spawn** when `model` is not accepted for the resolved agent flavor (non-zero exit + accepted list). Prefer validating against the flavor's catalog (`cursor-models` / equivalent), not trusting CLI help examples alone.
2. **Strengthen the spawn-peer success assertion**: remit-only `messages>=1` is insufficient. Assert agent started / lifecycle running / at least one assistant (or non-user) message before printing OK — or clearly label remit-only as "delivered handoff, agent not verified".
3. Document accepted model ids per flavor where they diverge from the underlying CLI's `--help` examples.

(1) is the important fix. A spawn that fails closed and still says OK is worse than one that errors — operators wait on a dead peer.

## Related

- Estate memory / prior bites: bad model param leaves peer silent; overseer must verify — still no product guardrail
- Empty-shell postmortem (remit missing): different failure mode; this is remit present, agent never runs
- `#1509` / spawn-peer first-class surface — adjacent; does not cover model validation
- `#1427` — Claude DeepSeek / model=auto UI gap — related "bad/default model" family, different flavor

## Acceptance

- [ ] `--agent cursor --model gpt-5` (or any unknown id) fails at spawn with a clear error listing accepted models (or equivalent catalog check)
- [ ] Known-good catalog ids still spawn and run
- [ ] `hapi spawn-peer` / wrapper does not print unconditional OK on remit-only when the agent never became active

heavygee:
Fixes #1752

## Problem

Spawning with a model the agent does not have — `--agent cursor --model gpt-5` — booted a child that died at the agent handshake. The hub row was already created by then, so the result was an archived session with zero turns and no explanation anywhere: no error on the spawn call, nothing in the session, nothing in the UI. `gpt-5` is not in `cursor-agent --list-models` (the catalog has `gpt-5.1` … `gpt-5.6-sol`), but nothing on the spawn path checked.

`SpawnSessionRequestSchema.model` is an unconstrained `z.string().optional()` and the runner forwards it straight to `--model`.

## Fix

A runner preflight beside the existing agent-availability check: when the machine already has a cached model catalog for the resolved flavor, an id that is not in it fails the spawn with `code: 'model_unavailable'` and a message naming the accepted ids. Nothing is spawned, so no zombie row is created.

```
Model 'gpt-5' is not in the cursor model catalog on this machine.
Accepted: gpt-5-mini, gpt-5.1, gpt-5.2, gpt-5.3-codex, gpt-5.4, gpt-5.4-mini,
gpt-5.4-nano, gpt-5.5, gpt-5.6-luna, gpt-5.6-sol, gpt-5.6-terra, claude-fable-5, … (36 total)
```

Near-misses are listed first (a rejected `gpt-5` is most usefully answered with the `gpt-5.x` ids), and Cursor entries collapse to the base slug so the list is not spent on `gpt-5-mini[fast=false]`-style variants.

## Deliberately conservative

The preflight rejects only what it can prove wrong:

- **Never probes.** It reads the ACP snapshot, the in-process cache and the on-disk shared cache — nothing else. A spawn must not block on `agent --list-models`, which can take 30s and contends with the ACP spawn lease.
- **Empty catalog means "unknown", not "no models".** Cursor is the only flavor whose catalog is readable without a probe today; every other flavor reports `[]` and is never rejected. Adding a flavor is one line in `getCachedModelCatalog`.
- **Stale catalogs go dormant.** A cached catalog older than 24h is dropped rather than trusted: a runner can stay up for weeks, and a catalog predating a Cursor upgrade would otherwise reject an id that is now valid.
- **Matching is as permissive as the spawn path.** ACP wire ids and CLI skus collapse to their base slug, renamed bases (`grok-4.5` → `cursor-grok-4.5`) resolve through `resolveCursorLegacyModelBase` the same way `cursorStaleModelRemap` does, comparison is case-insensitive, and `auto` / `default` / `default[]` are always allowed.

## Tests

- `shared/src/spawnModelCatalog.test.ts` — 11 cases over the matching rules: rejection message and its ordering, exact/base/sku/legacy-alias/case-insensitive acceptance, wildcards, unknown catalog, and that Cursor sku-suffix stripping is not applied to other flavors.
- `cli/src/runner/spawnModelPreflight.test.ts` — 7 cases over the cached-catalog reader: the `#1752` repro, ACP snapshot and on-disk sources, the 24h freshness bound, flavors with no catalog, and no model at all.

Red-green verified: stubbing `validateSpawnModelAgainstCatalog` to always return `ok` fails 4 shared + 2 CLI tests.

`bun typecheck` clean. `test:cli` 2473 pass, `test:shared` 309 pass, `test:web` 2859 pass, `test:relay` 80 pass. `test:hub` has 2 failures (`TitleSuggestionService`, `resolveFcmConfig`) that reproduce identically with this change reverted and pass when those files run in isolation — pre-existing cross-file pollution, untouched here.

Also verified against this machine's real `~/.hapi/cache/cursor-models.json`: `gpt-5` rejected; `gpt-5.5`, `GPT-5.5`, `grok-4.5`, `composer-2.5`, `default[]` and no-model all allowed; `claude --model gpt-5` unaffected.

## Note

Agent CLI `--help` examples can drift from `--list-models` and from the hub catalog — this is what sent `gpt-5` down the spawn path in the first place. This change makes that drift fail loudly at spawn instead of silently at handshake.


heavygee:
Good catch — confirmed and fixed in d6cd148.

`listCursorModels` was feeding the shared cache file's own contents back through `applyInMemoryCache`, which writes it out again, so startup prewarm re-stamped an arbitrarily old snapshot as current on every boot. The 24h fail-open could never fire.

Applied the suggested shape: `applyInMemoryCache(response, persist = true)`, with `persist: false` in both branches that loaded `response` off the shared file (`listCursorModels` and `listCursorModelsWhileAcpActive`). Only a live source — ACP snapshot or probe — may stamp the file now. Added a comment at the function stating that invariant so a future edit doesn't quietly reintroduce it.

Regression test added as requested (`cli/src/runner/spawnModelPreflight.test.ts`), modelling a fresh runner: cold in-process cache, leftover shared file, aged 2 days, then the prewarm path.

```ts
writeSharedCursorModelsCache({ success: true, availableModels: [{ modelId: 'composer-2.5' }], currentModelId: 'composer-2.5' })
ageSharedCache(2 * 24 * 60 * 60 * 1000)
await listCursorModels()
expect(checkSpawnModel('cursor', 'gpt-5')).toEqual({ ok: true })
```

Red-green verified: reverting just the two `persist: false` arguments fails exactly this test and nothing else.

`bun typecheck` clean. cli 2474 pass, shared 309, web 2859, relay 80. The 2 hub failures (`TitleSuggestionService`, `resolveFcmConfig`) are pre-existing cross-file pollution — they reproduce identically with this whole PR reverted and pass when those files run alone.

heavygee:
Fixed in 410be98 — but not with the suggested form, which I have to push back on.

**The finding is real.** Base collapsing let an explicit sku through whenever *any* variant of its base existed, so `gpt-5.5-high-fast` passed a catalog offering only `gpt-5.5-medium`. `remapStaleCursorModelId` refuses to downgrade that pair by design and `resolveCursorSpawnModel` forwards the original, so the handshake failure landed anyway. Exactly as described.

**The suggested fix would have been much worse than the bug.** Treating `remapStaleCursorModelId(...) === null` as a rejection signal conflates "I will not rewrite this id" with "this id is invalid" — and null is the common case. Measured against this machine's real `~/.hapi/cache/cursor-models.json` (74 entries, all ACP wire ids, **zero** CLI sku variants):

```
gpt-5.5-high-fast     variant=true  remap=null   → would reject
gpt-5.5-medium        variant=true  remap=null   → would reject
cursor-grok-4.5-fast  variant=true  remap=null   → would reject
cursor-grok-4.5-high  variant=true  remap=null   → would reject
claude-opus-5-high    variant=true  remap=null   → would reject
```

Every one of those is a valid Cursor sku. A wire-only cache is a different *naming* of the same models — it carries no `effort=` params to match a sku against — so the remap declines, and the guard would have blocked essentially every suffixed spawn on this machine. That is a far bigger regression than the one it closes.

**What landed instead:** the exact-variant requirement applies only when the catalog actually *enumerates* CLI sku variants of the requested base. `['gpt-5.5-medium']` is authoritative about gpt-5.5's skus, so `gpt-5.5-high-fast` is rejected and the message names the siblings. A wire-only catalog is not authoritative about skus, so those keep matching on base.

Tests added as requested, including the false-rejection guard that pins why the suggested form was rejected:

```ts
// authoritative sku catalog → exact variant required
validateSpawnModelAgainstCatalog('cursor', 'gpt-5.5-high-fast', ['gpt-5.5-medium'])
// → not an available cursor variant of 'gpt-5.5'. Accepted: gpt-5.5-medium
validateSpawnModelAgainstCatalog('cursor', 'gpt-5.5-high', ['gpt-5.5-medium','gpt-5.5-high']) // ok
validateSpawnModelAgainstCatalog('cursor', 'gpt-5.5',      ['gpt-5.5-medium','gpt-5.5-high']) // ok

// wire-only catalog says nothing about skus → base matching stands
validateSpawnModelAgainstCatalog('cursor', 'cursor-grok-4.5-high', ['cursor-grok-4.5[fast=true]']) // ok
```

Red-green verified: stubbing the new rule to a no-op fails exactly the new rejection test.

Re-verified end to end against the real cached catalog — `gpt-5` still rejected; `gpt-5.5`, `gpt-5.5-high-fast`, `composer-2.5-high`, `cursor-grok-4.5-high`, `claude-opus-5-high`, `default[]` and no-model all allowed.

`bun typecheck` clean. cli 2474 pass, shared 311, web 2859, relay 80. The 2 hub failures are the pre-existing cross-file pollution noted earlier.

heavygee:
Round 3: **the second finding is right and I have reverted my own previous commit for it. The first one I am declining, with measurements.**

## Partial SKU caches — correct, reverted in d5fa896

You're right, and `cursorModels.test.ts` says so in a test name: *"unions shared partial cliModelSkus with probe results when lock is inactive"*. A shared row holding only `gpt-5.5-medium` is a **supported intermediate state**, not a complete catalog — and `enrichCursorModelsWithCliSkus` skips the probe entirely while an ACP session holds the CLI lock, so that state can persist indefinitely. My 410be98 would have rejected a valid `gpt-5.5-high` throughout.

That is a worse failure than the handshake error it pre-empted, so 410be98 is reverted rather than patched. I did not take the `authoritativeCursorSkus` + persisted-completeness-bit route: it is a new on-disk schema field and a structured catalog type threaded through three call sites, to gate a case the handshake already reports. Not proportionate for a low-impact fix.

## ACP wire params — declining, same trap as round 2

`remapStaleCursorModelId(...) === null` still is not an availability signal. Measured against this machine's real cache (74 wire ids):

```
catalog wires = 74,  remap→null = 0          ← nothing already cached would break
composer-2.5[thinking]        remap = null   ← would be REJECTED
gpt-5.5[reasoning=high]       remap = null   ← would be REJECTED
composer-2.5[fast=true]       remap = composer-2.5[fast=true]
claude-opus-5[fast=false]     remap = claude-opus-5[fast=false]
```

`composer-2.5[thinking]` is a partial param set for a model that plainly exists here — it is the id used throughout this PR's own fixtures. The null means "I will not rewrite this", not "this does not exist"; `resolveCursorSpawnModel` forwards it deliberately. Rejecting on null would block valid spawns whenever a user supplies fewer params than the cache happens to store.

## The contract, now stated in the code

Both findings are the same shape, so I have pinned the scope rather than keep re-narrowing it:

> **This guard rejects a model only when its base appears nowhere in the catalog. Variant-level availability is left to the agent handshake, because a cached catalog is not authoritative about variants** — sku rows are partial and mid-union, wire-only caches carry no sku naming, sku-only caches carry no param sets.

Two tests pin it so the scope is not re-narrowed by accident:

```ts
validateSpawnModelAgainstCatalog('cursor', 'gpt-5.5-high-fast', ['gpt-5.5-medium'])            // ok
validateSpawnModelAgainstCatalog('cursor', 'claude-opus-4-8[thinking=false,effort=high]',
    ['claude-opus-4-8[thinking=true,context=300k,effort=high,fast=false]'])                    // ok
```

This is the scope #1752 actually reported: `gpt-5`, a base present in the catalog in no form at all. That still fails closed, with the `gpt-5.x` ids named.

`bun typecheck` clean. cli 2474 pass, shared 310, web 2859, relay 80. Real-catalog check: `gpt-5` rejected; `gpt-5.5`, `gpt-5.5-high-fast`, `composer-2.5`, `composer-2.5[thinking]`, `cursor-grok-4.5-high`, `grok-4.5`, `default[]` and no-model all allowed.

swear01:
Heads-up on a gap this preflight cannot close, found while triaging a live Cursor failure.

The preflight validates `model` against the machine's cached catalog, but it cannot detect a case where **the catalog itself is wrong**. Current `listCursorModels` on `cursor-agent 2026.09.10-fd3934a` returns ids like `grok-4.6[fast=false]` — present in the catalog, and still rejected by Cursor at ACP `session/new`:

```
Cannot use this model: grok-4.6[fast=false]. Available models: auto, ..., cursor-grok-4.6-high, ...
```

Checked directly against the 76-entry catalog: only 2 ids (`composer-2.5[fast=false]`, `composer-2.5[fast=true]`) actually reach `session/new`. Cursor accepts a bare base, but once a `[` is present it requires that model's **complete** parameter set; the synthesis in `cli/src/cursor/utils/cursorAcpModelsSnapshot.ts:60-72` writes only `fast` and drops the `thought_level` effort.

Two practical consequences for this PR:

1. Collapsing Cursor entries to base slugs (as the accepted-ids message does) is the same normalization that would make a catalog-derived-but-unspawnable id look valid, so the preflight's `model_unavailable` path will not fire for it. That is fine for scope — it just should not be read as "spawn model validation is now covered for Cursor".
2. Worth keeping in mind for the "empty catalog means unknown" rule: a non-empty catalog is also not proof of spawnability for Cursor, so this preflight and a catalog-correctness fix are complementary rather than redundant.

Tracked in #1818 with the per-id verification and the `_meta.parameterizedModelPicker` A/B (without that capability Cursor returns complete wires, which is how the root cause was isolated).

PR #1780, opened 2026-09-06
feat(codex): support backend-authorized Luna Reserve fallback

### LINKED ISSUE #1779: Support Codex Luna Reserve fallback with usage shown only while active

Support Codex’s official Luna Reserve fallback in HAPI remote sessions, with contextual usage visibility.

When ordinary Codex usage is exhausted and OpenAI enables Luna Reserve for the account, users should be able to continue the same conversation with GPT-5.6 Luna using the separate, limited Reserve allowance. HAPI should reflect the official eligibility and activation state and keep its session settings synchronized with Codex.

**Luna Reserve must stay hidden from Codex Usage during normal usage.** An available Reserve bucket alone must not create a permanent row, badge, or model-picker option.

Expected behavior:

- **Normal usage:** Keep the selected model and existing usage display. Do not show Luna Reserve, even if the backend reports an unused Reserve allowance.
- **Official Reserve activation after ordinary usage is exhausted:** Continue through Codex’s supported fallback flow in the same conversation. Show a concise transition notice and the effective Luna model, then reveal the Luna Reserve allowance in Codex Usage.
- **Reserve active:** Show its own remaining percentage and any backend-provided reset time. Keep it separate from ordinary usage windows; do not overwrite the ordinary weekly allowance with a Reserve weekly allowance. Use `100 - usedPercent` only for a valid reported percentage; missing data is unknown, not 0% or 100%. Do not hardcode the quota-window duration or reset schedule.
- **Reserve exhausted:** Show the exhausted Reserve state and let Codex report the usage block. Do not loop retries or replay already-executed user messages/tool actions.
- **Ordinary usage restored and Reserve exited:** Synchronize the official recovery/model state and hide the Reserve-specific usage row and badge again. Preserve the task’s previous model/reasoning settings where the supported recovery flow requires them.
- **Reconnect, resume, or local/remote handoff:** Reconcile actual Codex state so stale HAPI model settings cannot undo the fallback or show an inactive Reserve mode as active.

Reserve is a conditional allowance for Luna, not a separately selectable product model. No additional public model option, global default, custom inference proxy, or user-facing routing configuration is needed.

### Additional information

OpenAI documents Luna Reserve as a separate, limited fallback allowance after ordinary usage is exhausted, available only to selected accounts:
https://help.openai.com/en/articles/20001499-luna-reserve-in-codex-and-chatgpt-work

The integration boundary matters: HAPI remote mode starts `codex app-server` and submits `turn/start` with model settings. In the official implementation inspected, the TUI handles backend-authorized fallback and calls `thread/settings/update`; upgrading Codex alone should not be assumed to give every app-server client that TUI behavior:
https://github.com/openai/codex/blob/main/codex-rs/tui/src/app/backend_banner_fallback.rs
https://github.com/openai/codex/blob/main/codex-rs/tui/src/chatwidget/backend_banners.rs

Latest upstream protocol source includes `supportsLunaReserve`, `ordinaryUsageAllowed`, and per-limit snapshots. Activation/recovery must use authoritative backend signals, not inferred eligibility from percentages or reset timestamps. Follow the official supported protocol and pass required identifiers only at the integration boundary:
https://github.com/openai/codex/blob/main/codex-rs/app-server-protocol/src/protocol/v2/account.rs

These references describe upstream source, not a verified minimum released Codex version. Confirm the supported release/protocol and validate an eligible account’s ordinary → Reserve → exhausted/recovered transitions before claiming full support. Clients without the required capability should retain accurate existing behavior without claiming Reserve activation.

swear01:
When Codex authorizes Luna Reserve after ordinary usage is exhausted, HAPI now updates the same thread's settings and shows its separate allowance. An unused Reserve bucket stays hidden; exiting Reserve hides it again. No Reserve model is added to the picker, and Reserve transitions never replay submitted messages.

The CLI owns the app-server client flow: fresh account reads authorize fallback/recovery, accepted thread settings establish activation, and idle resume/reconnect reconciles native settings before another turn. Explicit queued settings survive ordinary resume; quota failures do not retry, while existing transient/context recovery remains available. Hub/Web receive normalized usage only. The task-local return record follows the official TUI format to preserve the previous model/reasoning settings across handoff and native forks, preserving the parent record. Reserve limits are identified by the backend's limit name, independently of ordinary windows; missing/invalid percentages remain unknown.

Protocol evidence: [official Codex source ac192cd7937](https://github.com/openai/codex/tree/ac192cd7937b0d73edc6dffe009940ae53782dd4), including `backend_banner_fallback.rs`, `chatwidget/backend_banners.rs`, and `v2/account.rs`. Capability opt-in follows protocol discovery. Codex 0.153.4 requires null parameters for the initial usage read and lacks `ordinaryUsageAllowed`; it keeps ordinary usage without activating Reserve.

Validation:
- Full `bun typecheck && bun run test`: exit 0; 6,980 passed, 1 skipped.
- Regression coverage for conditional visibility, separate windows, unknown percentages, recovery/account identity, stale reads, rejected settings, idle resume/reconnect, native forks, explicit queued settings, transient/context recovery during Reserve, and quota failures without replay.
- Real installed Codex 0.153.4: generated protocol schema and successful app-server initialize + account/rateLimits/read with null parameters.

Validation limit: no eligible-account ordinary → Reserve → exhausted/recovered inference run. No minimum released Codex version is claimed; the required capability is present in the pinned upstream source, not the installed 0.153.4 release. UI behavior is covered by jsdom component tests.

Fixes #1779

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -211,3 +211,25 @@ bun run build:single-exe
 
 - `../hub/README.md`
 - `../web/README.md`
+
+### Codex Luna Reserve
+
+Remote sessions read Codex's account usage through app-server. Luna Reserve is
+shown only while the task is using the backend-authorized Reserve route; an
+unused Reserve allowance never adds a row or a model option. Ordinary and
+Reserve windows remain separate, and unavailable percentages stay unknown.
+
+HAPI applies the official fallback with `thread/settings/update` and restores
+the task's saved model and reasoning effort after a fresh account read permits
+ordinary usage. It does not replay the blocked turn. Resume first reconciles
+Codex's task settings; the task-local return record uses Codex's
+`CODEX_HOME/tui-luna-reserve/<thread-id>.json` format for TUI handoff.
+
+This requires the newer app-server protocol with `ordinaryUsageAllowed`,
+`supportsLunaReserve`, the hidden Reserve catalog entry, and thread settings
+updates. Verified against official source
+[`ac192cd7937`](https://github.com/openai/codex/tree/ac192cd7937b0d73edc6dffe009940ae53782dd4).
+Codex 0.153.4 does not expose the required usage capability; no minimum released
+version is claimed. Older servers keep their ordinary usage display without
+advertising Reserve activation. Real eligible-account exhaustion, Reserve
+exhaustion, and recovery still require account-level validation.
--- a/cli/src/codex/appServerTypes.ts
+++ b/cli/src/codex/appServerTypes.ts
@@ -32,6 +32,7 @@ export interface InitializeResponse {
 }
 
 export interface ModelListParams {
+    cursor?: string;
     includeHidden?: boolean;
 }
 
@@ -385,4 +386,12 @@ export interface ExperimentalFeatureEnablementSetParams {
 export interface ExperimentalFeatureEnablementSetResponse {
     enablement: Record<string, boolean>;
     [key: string]: unknown;
+}
+
+export interface ThreadSettingsUpdateParams {
+    threadId: string;
+    model: string;
+    effort?: string | null;
+    serviceTier?: string | null;
+    collaborationMode?: CollaborationMode;
 }
--- a/cli/src/codex/codexAppServerClient.test.ts
+++ b/cli/src/codex/codexAppServerClient.test.ts
@@ -62,6 +62,26 @@ describe('CodexAppServerClient process cwd', () => {
         spawnMock.mockReset();
     });
 
+    it('uses null for legacy usage reads and opts in only when Reserve is supported', async () => {
+        const child = fakeChild();
+        child.stdin.write = vi.fn((_data: unknown, callback?: (error?: Error | null) => void) => {
+            callback?.();
+            return true;
+        });
+        spawnMock.mockReturnValue(child);
+        const client = new CodexAppServerClient({ cwd: '/neutral-home' });
+        await client.connect();
+        for (const supported of [false, true]) {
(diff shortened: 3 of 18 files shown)


tiann:
I am reverting this feature.

An explicitly selected model is part of a task's capability and quality requirements. Backend authorization to use Reserve allowance is not user consent to replace that model. Automatically moving a task from, for example, `gpt-6-astra` to `gpt-5.6-luna` may keep it running, but does not preserve the quality assumptions under which it was started. Quota exhaustion already produces a visible Codex error; HAPI does not need to add an automatic fallback or another quota-handling workflow.

There is also a concrete regression: each idle session polls account usage every 60 seconds and republishes agent state even when the usage is unchanged. The hub treats that write as session activity, so long-idle session cards keep showing `now`. The polling and state updates also multiply with the number of open sessions.

The revert removes the Reserve fallback/recovery, account-usage polling, and related UI, protocol fields, tests, and documentation, including the later shared-runtime integration. Generic model pagination and method probing needed by the current shared-session architecture are retained. No replacement quota logic is added.

Validation: `bun typecheck` passed; the full test suite passed with 7,435 tests and 10 skipped, using `NODE_OPTIONS=--no-experimental-webstorage` to avoid the local Node 26/jsdom WebStorage conflict. An idle-session regression test verifies that heartbeats continue without usage polling or agent-state churn.

Local revert commit: `7841f5e892896c729b9ad7a51f534e655f1ff54e` (not pushed yet).

PR #1798, opened 2026-09-09
feat(web): show unavailable Agents on demand

techotaku39:
## Summary

- Keep Create Session focused on Agents available on the selected machine by default.
- Add an inline `More Agents` button after the available Agent options.
- Reveal unavailable supported Agents on demand as disabled options with availability reasons.

## Problem / Motivation

Availability filtering can make supported Agents invisible when they are not installed on a particular machine.

This change keeps the normal Create Session flow compact while making the complete supported Agent catalog discoverable at the point of use.

## Validation

- `bun run --cwd web test -- src/components/NewSession/AgentSelector.test.tsx src/components/NewSession/index.test.tsx` — 44/44 passed
- `bun typecheck` — passed
- `bun run build` — passed
- Full live browser smoke through the task-isolated Playwright environment — 1/1 passed
  - Available Agents remain visible by default.
  - `More Agents` appears inline after the available Agent options.
  - Unavailable Agents expand on demand and remain disabled with reasons.

## Related Issues

Refs #1451

## AI Disclosure

Implemented with OpenAI Codex using GPT-5.6.


tiann:
Thanks for addressing Agent discoverability. Keeping the default Create Session UI focused on available Agents makes sense.

My concern is that a global Display setting is itself hard to discover: users who do not know HAPI supports additional Agents are unlikely to look for this toggle. Once enabled, it also keeps unavailable options in every session-creation flow, even though exploring supported Agents is usually an occasional task.

Could we instead add a lightweight “View other supported Agents” disclosure directly below the Agent selector? It could expand unavailable Agents with their reasons and, where useful, links to installation/configuration docs. This would address discoverability at the point of use without adding a persistent global preference.

I support the underlying problem being solved; I would prefer that contextual, on-demand interaction over the settings toggle.

techotaku39:
Thanks for the thoughtful feedback. I agree that a contextual disclosure is more discoverable and better suited to an occasional exploration flow.

I’ve updated PR #1798 to use this approach:

- Available Agents remain visible by default.
- A compact `More Agents` button appears inline after the available Agent options.
- Clicking it reveals supported but unavailable Agents as disabled options with their availability reasons.
- The button changes to `Hide other Agents` when expanded, and the layout wraps naturally when needed.

I chose the shorter `More Agents` label so the disclosure fits naturally alongside the Agent selector.

Validation has been completed with the targeted Web tests, full typecheck, build, and a live browser smoke test.

PR #1809, opened 2026-09-10
fix(web): render cold sessions from a small initial page

### LINKED ISSUE #1806: bug(web): slow first paint when opening long sessions without a cached message window
## Environment

- HAPI version: `origin/main` at `e2518b6` (reference reproduction)
- Client: HAPI Web/PWA with Chromium via Playwright
- Host: Windows x64 isolated task environment
- Agent flavor: Codex
- Network: tested against both the local Hub and the task's public FRP URL
- Test data: 1,200 Codex-shaped messages in one session

## What issue are you seeing?

When opening a long-running Codex session for the first time in a browser context without a persisted message-window cache, the chat area remains blank or shows a loading skeleton for several seconds before the latest messages become visible.

The session shell, header, and composer may already be visible, but the conversation itself is not immediately usable. The delay is especially noticeable when accessing the Hub through FRP.

This report is specifically about a cold or uncached first open. Reopening a session with an existing message-window cache is a different path and is related to the latest-tail re-entry work in #1549 and #1560.

## Steps to reproduce

1. Open a long-running Codex session with substantial message history.
2. Use a fresh browser context, or open the session before a persisted `hapi:message-window:v2:<sessionId>` entry exists.
3. Navigate to the session.
4. Observe that the latest message is not shown immediately.
5. Wait several seconds and observe that the latest messages eventually appear.
6. Repeat through the FRP public URL and compare the perceived delay.

## Expected behavior

As soon as the latest tail is available, the latest usable message or rendered conversation block should become visible without waiting for the complete initial client-side message processing and rendering path to finish.

Older messages may continue loading or hydrating in the background.

If the user immediately scrolls upward, that gesture should still trigger the normal older-history loading path. It should not stop history loading entirely.

Loading additional history should:

- Preserve the normal chronological display order.
- Avoid forcing the user back to the bottom.
- Preserve the current scroll position when older messages are inserted.
- Avoid duplicate or missing messages.
- Preserve queued, optimistic, and live SSE message behavior.

## Measurements

In an isolated Chromium measurement using 1,200 Codex-shaped messages:

- The cold path requested `GET /api/sessions/:id/messages?limit=200`.
- The response contained 200 messages and was approximately 245 KB after decoding.
- Browser resource timing measured approximately 13 ms for the local Hub response.
- After the message response completed, the first message became visible approximately 3 seconds later.
- Through the public FRP URL, the network portion was variable and one browser sample took approximately 1.8 seconds; the post-response rendering phase remained approximately 3 seconds.
- Reducing the API limit from 200 to 20 greatly reduced the decoded response size, but only reduced the warm local request time by a few milliseconds.

The test payload used repetitive synthetic text, so its compressed wire size is not representative of every real conversation. The decoded message size and browser-side rendering path are still relevant to the client behavior.

This suggests that changing the page size alone may not fully address the problem. The client-side JSON handling, message-window merge, message normalization, chat-block reduction, assistant-ui mounting, and initial scroll/layout work should be profiled.

## Relevant code

- `web/src/lib/message-window-store.ts:45` — defines `PAGE_SIZE = 200`.
- `web/src/lib/message-window-store.ts:670-696` — cold tail synchronization waits for the latest message response before applying the window.
- `web/src/components/AssistantChat/HappyThread.tsx:1627` — displays the message skeleton while tail synchronization is active and no messages have been rendered.
- `web/src/components/SessionChat.tsx:1312-1373` — filters, normalizes, reduces, and reconciles the loaded messages before rendering.
- `hub/src/sync/messageService.ts:302-393` — returns the latest page and pagination cursor.

## Related work

- #1549 / #1560 — cached session re-entry prioritizing the latest tail.
- #1539 / #1542 — preserving scroll position and loaded older history while a session streams.
- #1730 — automatically hiding primary session scrollbars.

This issue is limited to the slow first meaningful paint when opening a long session without a usable cached message window.

## Possible direction

The implementation is open to alternatives. Possible directions include:

- Profile the browser-side stages after the message response completes.
- Render the latest usable block before hydrating the complete initial window.
- Process or render the remaining messages incrementally.
- Keep the existing cursor/epoch pagination and scroll-anchor guarantees.
- Keep the DOM in normal chronological order; literal reversed DOM order is not required.

techotaku39:
## Summary

- Request only the latest 20 messages for a genuinely cold session window across Web, iOS, and Android.
- Keep the existing 200-message page size for resets, re-entry with cached cursors, forward sync, and older-history loading.
- Add cold-start regression coverage, update shared pagination fixtures, and document the pagination contract.

## Problem / Motivation

Opening a previously unseen session can remain blank while the client waits for the full latest page to arrive, decode, and render. This is especially noticeable over a slow or remote connection.

The initial page only needs to establish a useful conversation tail. Older history can continue using the existing 200-message pagination path when the user requests it.

In the existing 1,200-message Web test scenario, the initial decoded payload decreased from approximately 244.7 KB to 24.7 KB. First-message visibility measured approximately 0.46–0.69 seconds across three optimized runs, compared with approximately 1.59 seconds for the previous 200-message cold request.

## Implementation

- Add `INITIAL_PAGE_SIZE = 20` to the Web, iOS, and Android window implementations.
- Use the smaller page only when there is no cached newest cursor and no structural latest reset.
- Preserve `PAGE_SIZE = 200` for structural resets, cached re-entry, forward synchronization, and older-page requests on all clients.
- Extend the Web history-load fixture with an explicit cold-start mode and deterministic loading probe.
- Add Web, iOS, and Android regression coverage for cold and reset request sizes.
- Regenerate the affected shared pagination fixtures so all client conformance suites use the new cold-page expectation.
- Update the client pagination documentation and the CI E2E whitelist.

No new dependencies or database migrations are introduced.

## Validation

- `bun typecheck` — passed.
- `bun run --cwd web test -- src/lib/message-window-store.test.ts src/lib/message-window-store.fixtures.test.ts` — 83 tests passed.
- `pwsh -NoProfile -File .\\scripts\\Invoke-HapiTaskPlaywright.ps1 -Name investigate-session-first-paint -Suite Root -TestArgs 'cold-initial-tail.spec.ts'` — 1 test passed after adding an explicit initial-render wait.
- `pwsh -NoProfile -File .\\scripts\\Invoke-HapiTaskPlaywright.ps1 -Name investigate-session-first-paint -Suite Root -TestArgs 'terminal-wrap-fidelity.spec.ts'` — 2 tests passed.
- `bun run build` — passed, including CLI typecheck, Web build, embedded assets, and Hub build.
- `bun run gen:fixtures` followed by pagination fixture regeneration comparison — no fixture drift.
- Public smoke validation against three sanitized real Codex sessions — passed; initial windows rendered successfully and API pagination returned 20 latest messages and 200 older messages.
- iOS package tests and Android Gradle tests were not runnable locally because this Windows environment has no Swift toolchain and no JDK 17; the PR CI workflows compile and test the affected native targets.

## Risk / Rollback

A genuinely cold session initially contains 20 messages instead of 200, so older content is fetched when the user loads history. Reset and cached re-entry paths retain the existing full-page behavior.

Reverting this change restores the previous 200-message cold request on all clients.

## Related Issues

Fixes #1806

## AI Disclosure

Implemented and validated with Codex (GPT-5.6).


tiann:
Reducing cold-start work makes sense. My concern is that **20 raw messages do not necessarily produce 20 meaningful UI items**. HAPI reduces and groups messages before rendering, so a tool/reasoning-heavy tail may collapse into very little visible content, with the relevant conversation context outside the initial page.

The new text-only fixture verifies the request size and basic rendering, but does not cover this distinction.

Could you validate cold opens with realistic mixed-message tails and add regression coverage for:
- Useful visible content after reduction/grouping.
- Tool calls and results spanning the initial-page boundary.
- Actual scroll-triggered history loading when the reduced initial content is short.

I’m not saying 20 is necessarily too small. I’d like the choice validated against the **post-reduction user experience**, rather than raw message count alone.

techotaku39:
Thanks for the follow-up. I completed the additional validation against the current PR build, focusing on the post-reduction UI and the actual behavior of the newest page.

Results:

- In a mixed synthetic Agent tail containing user messages, reasoning, tool calls, tool results, and assistant text, the initial `limit=20` response reduced to 7 visible UI blocks. The rendered content was 1173 px tall in a 720 px viewport.
- I tested a boundary case where a tool call was just outside the initial 20-message page and its result was inside it. After loading the older page, the call and result remained correctly paired.
- In a compact case where the initial messages reduced to only one visible block, the existing HappyThread viewport-coverage safeguard automatically requested an older page with `limit=200`.
- I also tested three sanitized real sessions using HAPI’s Codex integration. Each initially requested `latest?limit=20`. The browser-level upward-scroll path then successfully requested `before?limit=200`, and older content rendered afterward, with 7, 3, and 13 visible rows respectively. The pagination behavior described here is part of HAPI’s shared Agent message path; Codex was used as the available real-world workload for these sessions.

One important detail about the pagination semantics:

- `latest?limit=20` selects the newest 20 raw message rows by their compound position. The response is returned in ascending display order, so these are effectively the bottom 20 rows of the conversation.
- A long final assistant response is still one raw message row; its length does not consume multiple page slots.
- In the normal pattern where tool calls and results appear above the final assistant text at the end of a completed Agent turn, that final text is one of the newest rows and is included in the initial page.
- If the newest page ends with tool records, that normally means the Agent has not produced its final response yet or the turn is still in progress. It is not a case where earlier tool calls have pushed a completed final response out of the page.

The validation did not reproduce the specific failure mode under discussion: a completed Agent turn whose final assistant text was missing from the newest 20 rows because there were many preceding tool records. The tests do show that raw message rows and rendered UI blocks are not equivalent, while also confirming that the existing viewport-coverage and scroll-triggered history loading work as a fallback.

Based on these results, `limit=20` appears to be a reasonable first-paint default for the cold latest-page path. I would keep #1809 focused on that optimization.

No code changes were made as part of this validation.

PR #1815, opened 2026-09-11
fix(pi): make user model selection sticky — do not revert to startup model on list refresh

HongChenGG:
## Problem

Launching Pi with a startup model (`hapi pi --model <m>`) captures that model as `initialModel`, meant to be applied once Pi's available models are confirmed. But `initialModel` was `readonly` and never cleared, so **every** subsequent `get_available_models` response re-issued `set_model` for the startup model and silently reverted whatever model the user had switched to.

Those responses are not rare: the web model picker polls `ListPiModels` (`/api/sessions/:id/pi-models`), and Pi re-emits the model list on window focus and reconnect. Net effect: switch the model in the web UI, and the next list refresh snaps it back to the launch model.

## Fix

1. `cli/src/pi/session.ts` — make `initialModel` mutable and add `explicitModelSelection`.
2. `cli/src/pi/loop.ts`:
   - the startup model is now consumed (nulled out) on the first discovery that can act on it, so it gets exactly one chance;
   - application is additionally guarded by `!session.explicitModelSelection`;
   - an accepted `set_model` response sets `explicitModelSelection = true`.
3. `cli/src/pi/runPi.ts` — mark `explicitModelSelection = true` on both the web picker (`SetSessionConfig`) and the CLI (`/model`) paths *before* the switch round-trip, so a `get_available_models` response landing in between cannot re-apply the launch model over the user's choice.
4. `cli/src/pi/loop.test.ts` — 3 new unit tests: repeated list refreshes do not revert the user's selection, a selection made before the first discovery is preserved, and a `set_model` response marks the flag.

## Testing

- `cd cli && bun run typecheck` — clean
- `cd cli && bun run test` — 234 files, 2619 tests passed, 2 skipped

## AI assistance

Drafted with an AI coding agent (Vibe Coding); reviewed, tested and verified locally.

### DIFF
--- a/cli/src/pi/loop.test.ts
+++ b/cli/src/pi/loop.test.ts
@@ -844,6 +844,87 @@ describe('wireTransportEvents', () => {
         }
     });
 
+    it('applies startup model only once across repeated get_available_models responses (does not revert user switch on list refresh)', async () => {
+        vi.useFakeTimers();
+        try {
+            session = createMockSession('startup-model');
+            const transport = createMockTransport();
+            wireTransportEvents(transport, session, []);
+
+            // First discovery: applies startup model
+            emitEvent({
+                type: 'response',
+                command: 'get_available_models',
+                success: true,
+                data: { models: [{ id: 'startup-model', provider: 'provider' }, { id: 'user-model', provider: 'provider' }] },
+            });
+            await vi.advanceTimersByTimeAsync(0);
+            expect(transport.send).toHaveBeenCalledTimes(1);
+            expect(transport.send).toHaveBeenCalledWith(expect.objectContaining({
+                type: 'set_model', provider: 'provider', modelId: 'startup-model',
+            }));
+
+            // Simulate user explicitly switching to another model
+            session.explicitModelSelection = true;
+            session.currentModel = 'user-model';
+
+            // Subsequent discovery (e.g. web UI refetch on window focus): must NOT re-apply startup model
+            (transport.send as any).mockClear();
+            emitEvent({
+                type: 'response',
+                command: 'get_available_models',
+                success: true,
+                data: { models: [{ id: 'startup-model', provider: 'provider' }, { id: 'user-model', provider: 'provider' }] },
+            });
+            await vi.advanceTimersByTimeAsync(0);
+            expect(transport.send).not.toHaveBeenCalled();
+            expect(session.currentModel).toBe('user-model');
+        } finally {
+            vi.useRealTimers();
+        }
+    });
+
+    it('does not apply startup model if user already selected a model before first discovery', async () => {
+        vi.useFakeTimers();
+        try {
+            session = createMockSession('startup-model');
+            const transport = createMockTransport();
+            wireTransportEvents(transport, session, []);
+
+            // User selected model before discovery returned
+            session.explicitModelSelection = true;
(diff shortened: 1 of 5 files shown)


HongChenGG:
Addressed the residual coverage gap from the first review: the regression tests no longer set `explicitModelSelection` directly on the loop path — they now exercise the real request ordering in `runPi`:

- **web picker** (`SetSessionConfig`): Pi's first `get_available_models` response lands while the picker's `set_model` is still unconfirmed. Asserts the launch model is not queued behind it (1 `set_model` total, the user's).
- **`/model`**: a later list refresh arrives before the switch confirms. Asserts no launch-model re-application (2 `set_model` total, last one the user's).

Both fail on the pre-fix code (2 and 3 `set_model` calls respectively) and pass with the fix. `cd cli && bun run typecheck` clean, `bun run test` 225 files / 2471 passed.

HongChenGG:
The residual gap from the last two reviews was a real bug, not just missing coverage — thanks, the suggested reproduction was accurate.

**The hole:** the startup-model gate is evaluated when `get_available_models` lands, but `explicitModelSelection` is recorded *inside* the `runRuntimeMutation` body, i.e. only once the lease is granted. So a picker/\`/model\` selection that was merely **queued** behind another runtime mutation had not set the flag yet; discovery scheduled the startup bootstrap behind that pending selection, and the bootstrap then ran after it and reverted the user's model.

**Fix (`656fe9d9`):** re-check `explicitModelSelection` inside the lease before applying the startup model (the outer gate is unchanged), and log the skip instead of the apply. Either the user's selection acquired the lease first and the bootstrap skips, or the bootstrap ran first and the queued selection applies after it — the user's choice always wins.

**Regression test:** hold the lease with an unrelated config change, queue a picker selection, deliver discovery, release the lease, assert only the user's `set_model` is sent. It fails on the previous head (2 `set_model` calls, the last one the launch model) and passes now.

Rebased onto current `main` (`10042f79`) as well: `bun run typecheck` clean, `bun run test` 234 files / 2622 passed.

HongChenGG:
Note: these same three commits (the one-shot startup model + `explicitModelSelection` gate + the lease re-check, plus their tests) are now also in #1715, because #1715's new 15s `get_available_models` poll is what makes the reversion reachable every few seconds.

This PR stands on its own against current `main` (where the reversion is only reachable via reconnect/focus/picker refreshes), so either can land first — **just not both**: merging one makes the other conflict on `cli/src/pi/loop.ts`. Happy to close either one, whichever you prefer.

HongChenGG:
Closing this as superseded — #1715 now carries these same three commits (`fix(pi): make user model selection sticky`, its tests, and the lease re-check).

The guard has to ship with #1715 rather than separately: #1715's new 15s `get_available_models` poll is what makes the reversion reachable every few seconds, so that PR is not correct without it. Splitting it only creates a guaranteed conflict on `cli/src/pi/loop.ts` and makes the maintainer arbitrate which landed first.

Nothing is lost — the branch is kept at `HongChenGG/hapi:fix/pi-sticky-user-model` (`64a65b92`, CI 4/4 green, bot: no reportable issues). It stands alone against current `main` (where the reversion is only reachable via reconnect/focus/picker refreshes), so if #1715 does not land I can reopen this one as-is.

PR #1826, opened 2026-09-11
fix(pi): keep imported transcript titles out of metadata.name

luodeb:
## What

Pi session auto-titling wrote the transcript title into **`metadata.name`** — the slot the web title helper (`getSessionTitle`) prefers and the one hub `renameSession` uses for **manually chosen** names. So an agent-set title (the `hapi_set_title` extension writes `metadata.summary.text`) never surfaced in the web session list; the imported first-message title kept winning.

The hub already documents its intended split (`sessionCache.ts`): *"Keep the generated/native title separate from `metadata.name`. A manually chosen name must continue to win in the Web title helper, while the summary remains available as the agent-authored fallback."*

## Fix (fix the source, not the consumer)

- **`hub/src/web/routes/piSessions.ts`**: `buildPiMetadata` no longer claims `metadata.name` from `transcript.title`. An existing manual `name` is preserved; the imported title rides in the existing `summary` fallback (which already prefers `lastUserMessage`, then `title`).
- The Pi extension `setTitle` path stays on `metadata.summary.text` — it must not overwrite a manually chosen name.

## Tests

- `hub/src/web/routes/piSessions.test.ts` (new, x2): imported transcript title stays out of `metadata.name` when the session has no manual name (summary carries `lastUserMessage`); and a session with no user messages falls back to the transcript title in summary.
- `cli/src/pi/extensionUiHandler.test.ts`: regression — with a manual `metadata.name` set, a `setTitle` delivers the title to the summary slot and the name survives unchanged.
- `bun run typecheck` green across all packages (relay/hub/cli/web); cli + hub suites pass.

## Before/after

| state | `metadata.name` | web list shows |
|---|---|---|
| before | transcript.title (auto) | first-message title, agent titles invisible |
| after | manual name only | manual name wins; else agent/import title from summary |

---

*AI disclosure: this change was implemented with assistance from the AI model `deepseek/deepseek-v4-flash`.*

### DIFF
--- a/cli/src/pi/extensionUiHandler.test.ts
+++ b/cli/src/pi/extensionUiHandler.test.ts
@@ -6,6 +6,7 @@ type PermissionHandler = (response: unknown) => Promise<void>;
 function createHarness() {
     let permissionHandler: PermissionHandler | null = null;
     let state: Record<string, unknown> = { requests: {}, completedRequests: {} };
+    let capturedUpdater: ((metadata: Record<string, unknown> | null) => Record<string, unknown>) | null = null;
     const session = {
         rpcHandlerManager: {
             registerHandler: vi.fn((_method: unknown, handler: PermissionHandler) => { permissionHandler = handler; }),
@@ -13,8 +14,11 @@ function createHarness() {
         updateAgentState: vi.fn((updater: (current: never) => unknown) => { state = updater(state as never) as Record<string, unknown>; }),
         sendAgentMessage: vi.fn(),
         sendSessionEvent: vi.fn(),
-        getMetadata: vi.fn(() => null),
-        updateMetadata: vi.fn(),
+        getMetadata: vi.fn((): Record<string, unknown> | null => null),
+        updateMetadata: vi.fn((updater: (metadata: Record<string, unknown> | null) => Record<string, unknown>) => {
+            capturedUpdater = updater;
+            updater(null);
+        }),
     };
     const sendResponse = vi.fn();
     const handler = new PiExtensionUiHandler({ session: session as never, sendResponse });
@@ -23,6 +27,7 @@ function createHarness() {
         session,
         sendResponse,
         state: () => state,
+        lastUpdater: () => capturedUpdater,
         respond: async (response: unknown) => permissionHandler?.(response),
     };
 }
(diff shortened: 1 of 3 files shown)


luodeb:
Thanks for the catch — you are right, and the fix now goes the other way.

**Updated approach (6203d0d8):** keep the Pi extension `setTitle` on the summary slot (no `name` targeting), and fix the generated-name **source** instead:

- `hub/src/web/routes/piSessions.ts` (`buildPiMetadata`): imported `transcript.title` no longer claims `metadata.name`; an existing manual `name` is preserved, and the import title rides in the existing `summary` fallback.
- Regression tests: Pi import keeps `metadata.name` empty for unnamed sessions (summary carries `lastUserMessage`, falling back to `transcript.title`); `extensionUiHandler.setTitle` preserves a manual `metadata.name` while updating the summary.

So: manual names survive every auto-title path, and import titles no longer shadow agent-set titles in the web list.

luodeb:
Good catch on the last-user-message precedence (e0d5a8a3).

Now `buildPiMetadata` computes `summaryText = transcript.title ?? transcript.lastUserMessage`, so a native Pi title (including explicit `session_info.name` values) survives the import and displays in the web list, while a manual `metadata.name` still wins over everything. The import summary regressions now assert: unnamed session → `metadata.name` absent + summary carries the transcript title; native title + last message `thanks` → title wins for display.

luodeb:
Also closed the remaining combination gap (2317b332): a hub test now covers import → later extension-style `summary` write landing through the metadata merge, asserting the manual-name slot stays empty so the web title helper keeps preferring the agent title.

luodeb:
**Behavior before/after this change** (for review convenience)

The whole fix is about who owns `metadata.name`: before, Pi imports claimed it with the transcript title; after, the slot belongs to manual renames only, and generated/native titles live in `metadata.summary.text` (the web title helper already prefers `name > summary.text`).

### Field placement on import

| field | before | after |
|---|---|---|
| `metadata.name` | `existing.name ?? transcript.title` (auto title takes the slot) | `existing.name ?? undefined` (manual names only) |
| `metadata.summary.text` | `lastUserMessage ?? transcript.title` | `transcript.title ?? lastUserMessage` (native title wins) |
| cli `hapi_set_title` landing | `summary.text` (unchanged) | `summary.text` (unchanged) |
| web shown title (`getSessionTitle`) | `name > summary` (unchanged) | `name > summary` (unchanged) |

### Scenario comparison

| scenario | before | after |
|---|---|---|
| **Pi session without a manual name; agent calls `hapi_set_title("new title")`** | title written to `summary`, but the web kept showing the import-claimed auto name → **agent title invisible** | `name` slot empty → **agent title visible right away** (the point of this PR) |
| imported session's list title | `transcript.title` (via `name`) | `transcript.title` (via `summary`) — visually the same |
| user renames in the web app | manual `name` wins, agent titles never override | **same: manual name always wins** (unbroken hub principle) |
| native title vs last user message ("fix login" vs "thanks") | summary got clobbered by "thanks" (masked by `name` = title) | summary directly holds "fix login" (the follow-up Minor fix) |
| peer labels (`list_peers` / `inspect-peer`) | `name` first → auto title | manual name > summary (native/agent title) — consistent with web |

### Unchanged

- Manual rename (web → `renameSession` → `metadata.name`) keeps priority, always.
- The `hapi_set_title` extension semantics are untouched.

### One caveat (existing data)

The change only affects **newly imported** sessions. Sessions already imported before this change keep the old auto title in `metadata.name` (it is not cleared by re-import reconcile, which preserves `existing.name`). For those, a manual rename is the way to update what the web shows.

PR #1831, opened 2026-09-12
feat(pi): inject hapi_display_image and render inline tool-result images

### LINKED ISSUE #1830: feat(pi): inject hapi_display_image and render inline tool-result images
## Problem

Pi sessions have no way to present a local image in the chat. Claude/Codex sessions get a `display_image` / `display_media` MCP tool from the injected HAPI MCP server; Pi sessions only get the `hapi_change_title` extension. Sending an image as a web attachment to a Pi session hits `(image omitted: model does not support images)` when the model has no vision support, so there is no reliable image-presentation path at all.

## Proposal

Follow the existing Pi extension injection pattern (`pi --mode rpc --extension <file>`, same file that already registers `hapi_change_title`):

1. **Inject a `hapi_display_image` tool** into Pi sessions: reads a local image (png/jpeg/gif/webp, magic-byte detection, 25 MiB cap) and returns it as a Pi `ImageContent` block in the tool result (`cli/src/pi/titleExtension.ts`).
2. **Parse inline image blocks in tool results**: `piEventConverter` already receives structured results (`{ content: [...] }`); route `image` blocks through `registerGeneratedImageFromAcpBlock` (shared with the ACP backend) into `generated_image` AgentMessages, so the image renders as a card in the web chat — independent of whether the model supports vision (`cli/src/pi/piEventConverter.ts`, `loop.ts` consumption).

This mirrors the ACP backend's existing generated-image pipeline (same shared function, zero adaptation), and the parsed images also cover *any* future Pi tool returning ImageContent (e.g. a `pi-mcp-adapter` screenshot server), not just the injected tool.

Scope: images only (Pi's `ToolResultContent` is `TextContent | ImageContent`); video/media stay out of scope. Verified end-to-end on an isolated hub (real Pi session calling the injected tool, image card rendered in the web chat).

## Screenshots

Web chat after the injected `hapi_display_image` tool runs (real Pi session, isolated hub):

pi display web

Same flow reached through the happy-server MCP `display_image` tool on a Claude session (for comparison):

mcp display web

luodeb:
Closes #1830

## What

Pi sessions get a working image-presentation path, following the existing extension-injection pattern (`pi --mode rpc --extension`, the same file that already registers `hapi_change_title`):

1. **`hapi_display_image` tool injected into Pi sessions** (`cli/src/pi/titleExtension.ts`): reads a local image (png/jpeg/gif/webp, magic-byte detection, 25 MiB cap) and returns it as a Pi `ImageContent` block in the tool result.

2. **Inline tool-result images now reach the chat** (`cli/src/pi/piEventConverter.ts` + `loop.ts`): `tool_execution_end` results are parsed as structured content — text blocks become the tool_result output, `image` blocks flow through `registerGeneratedImageFromAcpBlock` (the exact function the ACP backend already uses, so zero adaptation) into `generated_image` AgentMessages. The image renders as a card in the web chat **independent of whether the model supports vision** — this is the fix for web attachments to a Pi session ending up as `(image omitted: model does not support images)`.

Legacy string results keep the pass-through behavior; malformed/unsupported image payloads are dropped gracefully.

This also unlocks *any* Pi tool returning ImageContent (e.g. a future `pi-mcp-adapter` screenshot server), not just the injected one.

## Scope

- Images only: Pi's `ToolResultContent` is `TextContent | ImageContent` — video/media stay out of scope.
- Images enter the model context (tool result), unlike the Claude MCP `display_image` which is display-only; the tool description says so.

## Tests

- `piEventConverter` (27): inline image block → `generated_image` message, text-block joining, malformed-image drop, structured-result fallbacks, legacy string pass-through. `convertPiEvent` is now async; `loop.ts` consumes it fire-and-forget with error logging.
- `titleExtension`: extension source registers `hapi_display_image` with base64 + magic-byte mime detection.
- `bun run typecheck` green; cli full suite 2672 passed.

## End-to-end verification (isolated hub, real Pi session)

Real Pi session called the injected tool; the image card rendered in the web chat (blob URL, `GET /api/sessions/:id/generated-images/:id` → 200 `image/png`):

pi display web

Same flow through the happy-server MCP `display_image` tool on a Claude session, for comparison:

mcp display web

---

*AI disclosure: this change was implemented with assistance from the AI model `deepseek/deepseek-v4-flash`.*

### DIFF
--- a/cli/src/pi/loop.test.ts
+++ b/cli/src/pi/loop.test.ts
@@ -25,7 +25,7 @@ vi.mock('./piEventConverter', async (importOriginal) => {
     const actual = await importOriginal<typeof import('./piEventConverter')>();
     return {
         ...actual,
-        convertPiEvent: vi.fn(() => []),
+        convertPiEvent: vi.fn(async () => []),
     };
 });
 
@@ -1050,7 +1050,7 @@ describe('sendPiRpcAndWait', () => {
 });
 
 describe('Pi lifecycle timeline', () => {
-    it('synchronizes authoritative get_state streaming and deduplicates compaction/retry timeline events', () => {
+    it('synchronizes authoritative get_state streaming and deduplicates compaction/retry timeline events', async () => {
         let listener: ((event: Record<string, unknown>) => void) | null = null;
         const transport = {
             onEvent: vi.fn((handler: (event: Record<string, unknown>) => void) => { listener = handler; }),
@@ -1075,6 +1075,9 @@ describe('Pi lifecycle timeline', () => {
         emit({ type: 'compaction_end', reason: 'threshold', aborted: false, willRetry: false });
         emit({ type: 'auto_retry_start', attempt: 1, maxAttempts: 3, delayMs: 10, errorMessage: '429' });
         emit({ type: 'auto_retry_end', attempt: 1, success: true });
+        // Lifecycle notices publish through the serialized chat queue; flush the
+        // microtask chain before asserting what the session emitted.
+        await new Promise((resolve) => setTimeout(resolve, 0));
         expect(stateSession.client.sendSessionEvent).toHaveBeenCalledWith({ type: 'message', message: '📦 Compaction started' });
         expect(stateSession.client.sendSessionEvent).toHaveBeenCalledWith({ type: 'message', message: '📦 Compaction completed' });
         expect(stateSession.client.sendSessionEvent).toHaveBeenCalledTimes(8);
--- a/cli/src/pi/loop.ts
+++ b/cli/src/pi/loop.ts
@@ -691,6 +691,12 @@ export function wireTransportEvents(
     };
     const flushAccumulator = (): void => sendMessages(assistantMessageAccumulator.flush());
 
+    // Serializes chat publishing so tool results (which may await image
+    // registration) can never overtake the assistant snapshot of the next
+    // event in the same stdout chunk; JsonLineParser.feed() dispatches a chunk
+    // synchronously line-by-line, so without this chain an assistant
+    // `message_end` arriving after a `tool_execution_end` would publish first.
+    let publishChain: Promise<void> = Promise.resolve();
     transport.onEvent((event) => {
         // Legacy Pi emitted auto_compaction_*; normalize it before every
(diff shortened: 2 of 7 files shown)


luodeb:
Both Majors addressed in the same commit (`8e20ed2d`):

1. **Enforce the file limit before reading** (`titleExtension.ts`) — the injected tool now `stat`s the path first, rejects non-regular files (`/dev/zero` etc.) and anything over 25 MiB before any `readFile` allocation. Regression assertions added in `titleExtension.test.ts`.

2. **Preserve transport order when publishing** (`loop.ts`) — chat publishing is now serialized through a `publishChain`: assistant snapshot messages and converted tool events (including image registration) are published in event order, so a `tool_execution_end` in the same stdout chunk can no longer be overtaken by the next `message_end`. `convertPiEvent` stays async; the chain awaits it per event.

luodeb:
Minor addressed (`f12fd7cb`): lifecycle notices now publish through the same serialized queue, so a retry/compaction timeline event arriving later in a stdout chunk cannot overtake earlier assistant or image output.

Added a transport ordering regression test (`cli/src/pi/loopPublishOrder.test.ts`) using the **real** accumulator/converter path (not the per-file mocks): tool-result → image card → lifecycle notice ordering is asserted for both a text result and an inline image result feeding through `wireTransportEvents`.

Note on the failing `test` job: `AgentPicker.test.ts` fails on `main` itself (another open PR hits the identical failure) — it is an upstream flaky, unrelated to this diff.

luodeb:
**Note on the preview-related review comments**

Two inline comments in this thread (`hub/src/web/previewRoutes.ts` [Blocker], `cli/src/preview/proxyClient.ts` [Major]) reference files that are **not part of this PR's diff** (7 files under `cli/src/pi/*` only). They were generated while the head temporarily pointed at a `feat(preview)` commit (`d83c3093`) — an accidental push during a branch realignment. The head has since been corrected to `0ab08bfc` (single pi commit, no preview files), so those findings should be treated as stale/outdated.

The actual review findings for this PR (2 Major + 1 Minor, all addressed) are the `titleExtension.ts` and `loop.ts` ones below; the bot's latest review reports **no reportable issues** in the current diff.

PR #1772, opened 2026-09-05
feat(web): add combined session list filters

techotaku39:
## Summary

- Add a namespace-scoped batch endpoint for retrieving sessions with scratchlist entries.
- Replace the unread-only control with combined unread, date, and scratchlist filters.
- Allow multiple session filters to be selected together using intersection semantics.
- Unify session and machine filtering in the mobile filter menu.
- Improve responsive menu sizing, viewport clamping, and date-picker positioning.
- Add regression tests and update API, Web, and Hub documentation.

## Problem / Motivation

Previously, filtering was split across separate controls and did not provide a shared multi-select model. Users could not directly build combined queries such as “unread sessions updated within a date range that also contain scratchlist entries”, or consistently combine session criteria with machine filtering from one filter surface.

Scratchlist data was already persisted in the Hub and synchronized across clients, but the session list had no way to identify sessions containing scratchlist entries without opening them individually. Long date ranges and filter menus also needed more reliable positioning on narrow mobile viewports.

## Implementation Notes

- Added `GET /api/sessions/scratchlist-status`, returning distinct scratchlist-bearing session IDs within the authenticated namespace.
- Added Web API client, query hook, cache key, and SSE invalidation support.
- Added independent unread, date, and scratchlist filter state with AND/intersection filtering.
- Kept the selected session visible while applying filters.
- Unified session and machine filters in the mobile menu while preserving the desktop machine filter bar.
- Made filter menus size to their longest content and clamp only when viewport space is insufficient.
- Centered the date picker within the session-list panel.
- No new dependencies or database migration were introduced.

## Validation

- `bun run typecheck` — passed for CLI, Web, Hub, and Relay.
- `bun run --cwd hub test -- src/web/routes/sessions-scratchlist.test.ts src/store/migration-v12.test.ts` — 38/38 passed.
- Focused Web tests for session filters, machine filters, date filtering, and directory actions — 70/70 passed.
- Final machine-filter label regression tests — 32/32 passed.
- `.\scripts\Invoke-HapiTaskPlaywright.ps1 -Name investigate-draft-session-filter -Suite Root e2e/terminal-wrap-fidelity.spec.ts` — 2/2 passed.
- Real browser acceptance passed for date-picker centering at 393px and 1024px, dual-runner machine filtering, and Chinese UI labels.
- `bun run build` — passed.

## Related Issues

None

## AI Disclosure

OpenAI Codex (GPT-5.6)

PR #1842, opened 2026-09-12
feat(web): rank session list search by field weight and IDF

### LINKED ISSUE #1841: Session list search ranks by field weight + IDF (not recency alone)
## Problem
Session list metadata search is boolean match → sort by `updatedAt`. Ubiquitous path segments (`/home/…`, `coding`, …) match hundreds of sessions equally, so distinctive title hits like **Home Assistant** are buried under recently touched path matches.

## Proposal
- Keep paths fully searchable (no OS-home stripping).
- Rank matches by field weight × smoothed IDF; recency is tie-break only.
- Boundary affinity is a **score bonus**, not an exclusion gate (substring inclusion preserved for as-you-type, @-mentions, share picker).

## Out of scope
- Score cutoff / result capping (separate UX choice).
- Content (FTS) search ranking.
- PR #1772 combined filters (orthogonal).

heavygee:
## Summary
- Rank session-list metadata search by **field weight × smoothed IDF**, with **recency as tie-break** (was boolean match → recency only).
- Keep paths fully searchable (no OS `/home/` stripping).
- Boundary affinity is a **score bonus** (`×1.75`), not an exclusion gate — substring inclusion preserved for as-you-type, `@`-mentions, and share picker.
- Multi-word queries AND across terms (each token must hit some field).

### Field weights
| Field | Weight |
| title / \`getSessionTitle\` | 10 |
| worktree label | 5 |
| summary | 3 |
| flavor | 2 |
| machine label | 2 |
| path parts | 1 |
| session id | 0.5 |

IDF: \`log(1 + N/(1+df))\` over the time-scoped candidate corpus. Ubiquitous terms (\`home\` in \`/home/…\`) contribute little; distinctive terms dominate.

### Honest limitation
Ranking sorts ~800 equal-looking matches — it does **not** reduce them to a handful. A score cutoff would be a separate deliberate UX choice.

### Notes
- Orthogonal to #1772 (combined filters). Does not touch that PR.
- Fork-stage draft: heavygee/hapi#147. Local Claude cold review Ready YES after boundary-as-bonus fix (substring inclusion restored for shared \`sessionMatchesQuery\` callers).

## Test plan
- [x] \`web\` vitest: \`sessionListSearch\`, \`sessionReference\`, \`sharePickerSessions\`, \`SessionList\`, \`SessionList.machine-filter\`
- [x] Cold review (Claude) — Ready YES
- [ ] Dogfood: type \`Home\` in session list; **Home Assistant** should float above recent \`/home/…\` path-only hits

## Issues
Fixes #1841

### DIFF
--- a/web/src/components/SessionList.tsx
+++ b/web/src/components/SessionList.tsx
@@ -1,8 +1,14 @@
 import { useEffect, useMemo, useRef, useState } from 'react'
 import { SessionListScrollAnchor } from './SessionListScrollAnchor'
 import type { SessionSummary } from '@/types/api'
-import { isWildcardSearch, matchesSearchQuery } from '@hapi/protocol'
 import type { ApiClient } from '@/api/client'
+import {
+    buildSessionSearchScoreIndex,
+    compareSessionsBySearchRelevance,
+    rankSessionGroupsBySearchRelevance,
+    sessionMatchesQuery,
+    sortSessionsBySearchRelevance,
+} from '@/lib/sessionListSearch'
 import { useLongPress } from '@/hooks/useLongPress'
 import { usePlatform } from '@/hooks/usePlatform'
 import { useSessionActions } from '@/hooks/mutations/useSessionActions'
@@ -514,30 +520,10 @@ function SessionPreviewArrowIcon(props: { direction: 'up' | 'down'; className?: 
 }
 
 export { getSessionTitle } from '@/lib/sessionTitle'
+export { sessionMatchesQuery } from '@/lib/sessionListSearch'
 
 export function normalizeSearch(value: string | null | undefined): string {
     return (value ?? '').trim().toLowerCase()
-}
-
-export function sessionMatchesQuery(session: SessionSummary, query: string, machineLabel: string): boolean {
-    if (!query) return true
-    const searchableParts = [
-        getSessionTitle(session),
-        getWorktreeSessionLabel(session),
-        session.id,
-        session.metadata?.path,
-        session.metadata?.worktree?.basePath,
-        session.metadata?.worktree?.worktreePath,
-        session.metadata?.name,
-        session.metadata?.summary?.text,
-        session.metadata?.flavor,
-        machineLabel,
-    ]
-        .filter((part): part is string => typeof part === 'string' && part.length > 0)
-    if (isWildcardSearch(query)) {
-        return searchableParts.some((part) => matchesSearchQuery(part, query))
-    }
-    return searchableParts.join('\n').toLowerCase().includes(query)
 }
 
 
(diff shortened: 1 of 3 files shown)


heavygee:
Addressed the Minor: `bestWeightedHitForTerm` now compares post-bonus contribution (`weight × boundaryBonus`) when picking the best field, not raw weight. Tip `61615200f`.

tiann:
The direction makes sense. I found one reproducible UI regression to fix before merging, plus two design suggestions.

1. **Fix project pin dividers during relevance sorting.**

   `rankSessionGroupsBySearchRelevance` can interleave project-pinned and ordinary sessions, while the renderer still draws a divider at every pinned-to-unpinned transition.

   Reproduction: in one directory, create `Home pinned` (pinned), `Home ordinary`, `homelab pinned` (pinned), and `homelab ordinary`, with descending timestamps in that order. Searching `home` produces pinned → ordinary → pinned → ordinary, with **two pin dividers** in the same group. Before this PR, pinned rows remain together with one divider.

   Hide the divider during relevance sorting, or preserve a coherent pinned section, and add a component regression test.

2. **Clarify and test the benefit of IDF.**

   For a single-term query such as `Home`, IDF is the same multiplier for every matching session, so it cannot change their relative order. Field weights and the boundary bonus solve that example.

   IDF can help with multi-term queries, but the current “distinctive second term” test has only one session satisfying the AND filter; it does not demonstrate a ranking benefit. Consider simplifying the initial implementation, or add a case with at least two qualifying sessions where IDF improves their order.

3. **Consider a full-title / phrase bonus.**

   For query `Home Assistant`, titles `Home Assistant` and `Assistant for Home` currently receive identical scores, so the latter ranks first if newer.

   Cross-field AND matching is reasonable, but exact titles and contiguous phrases could retain priority. The shared matcher also changes multi-word inclusion in mentions and the share picker; dedicated coverage would make that behavior explicit.

Validation on `bcd9c89`: 139 relevant tests passed (`bun run test:web sessionListSearch sessionReference sharePickerSessions SessionList`), and `bun run typecheck:web` passed. An additional component reproduction for the divider issue passed against merge base `092a2259` and failed on the PR head: expected at most one divider, received two.

heavygee:
Apologies for merging ahead while your review was open — that was accidental on my part, and it shouldn't have happened.

On your three points:

**1. Pin dividers.** Confirmed — your reproduction fails on the merged head. Fixing it and adding the component regression test you describe, in a follow-up PR immediately.

**2. IDF.** Understood, and a great point. For a single-term query it's a constant multiplier across all matches and can't affect their order — field weights and the boundary bonus are doing that work. Our multi-term test doesn't demonstrate a benefit either, since only one session survives the AND filter. We'll either add a case with two qualifying sessions where IDF genuinely improves the order, or drop it and keep the simpler implementation.

**3. Phrase / exact-title bonus.** Agreed — `Home Assistant` losing to `Assistant for Home` on recency is wrong. We'll add it, plus the dedicated coverage for mentions and the share picker, since the shared matcher changes their multi-word behaviour too.

PR #1851, opened 2026-09-14
perf(hub): cache prepared statements and memoize per-socket session access

miuiadmin:
## Problem

Every CLI socket event — message-stream deltas, keep-alives, consumed acks, terminal output — runs the session-authorization gate (`resolveSessionAccess` in `hub/src/socket/handlers/cli/index.ts`). Each resolution paid a fresh `db.prepare()` + SELECT + session-metadata JSON parse. A streaming CLI emits dozens of events per second against the same session, and one runner socket multiplexes many concurrent sessions (agents driving parallel sessions on one machine), so session ids interleave on the same socket.

On a hub serving sustained event traffic this dominated the CPU profile: **~60% of a saturated event loop** (prepare 25.8%, metadata parse 16.3%, sqlite step 15.8%) — spent re-deriving an authorization answer that is stable for a given session id.

## Fix

Two independent, composable changes:

**1. Prepared-statement cache (`hub/src/store/statementCache.ts`)**

`WeakMap<Database, Map<sql, Statement>>`; `prepareCached(db, sql)` returns the compiled statement. Statements re-bind per call, so reuse is semantically identical to inline `db.prepare`. Swapped into `sessions.ts` (18 sites) and `messages.ts` (61 sites). `messageStore.ts` keeps inline prepares intentionally: its `iterate()` callers `finalize()` the returned statement, which a shared cache cannot survive.

**2. Per-socket session-access memo (`handlers/cli/index.ts`)**

Positive `resolveSessionAccess` results are memoized per socket for 1s, **keyed by session id** in a bounded `Map` (64 entries, expired-first then oldest-first eviction):

- Keyed by session id, not a single slot: one runner socket multiplexes concurrent sessions, and a single-slot memo thrashes to a miss on every event under A-B-A interleaving — profiled as seconds-long event-loop saturation while message floods alternate session ids.
- Denials stay uncached, so clients see a fresh outcome on every event while a rename / namespace switch settles.
- Namespace is immutable for a session id and metadata consumers tolerate ≤1s staleness, so the bounded window cannot change an access decision that would otherwise differ.

## Tests

- 5 new tests in `handlers/cli/index.test.ts`: burst memoization (one store read for many events), denials re-resolve per event, TTL expiry, A-B-A interleave survival, 64-entry cap eviction (oldest first)
- `statementCache.test.ts` covers cache identity / re-bind behavior
- Full hub suite passes (1298 pass); `bun typecheck` clean

## Notes

- No public API, protocol, or schema changes; behavior is visible only as CPU/latency under load.
- Validated on a self-hosted hub under sustained multi-session streaming: hub CPU during event floods dropped from full event-loop saturation to ~10-15%, and the authorization path disappeared as a profile hotspot.
- Independent of #1844 (different subsystems; the two PRs do not overlap).

## AI disclosure

Per the AI-Generated Code Policy: this change was developed with AI assistance (Claude Code agent, powered by GLM by Z.ai) and validated against the existing hub test suite; the TTL / eviction / denial-uncached semantics were forced out by the regression tests.


miuiadmin:
Both review findings addressed:

- **[Major] Read current metadata before preserving hub-owned fields** → `resolveSessionAccess` gained a `fresh` bypass (skips the memo read, still write-throughs the cache); `update-metadata` now resolves fresh, since its result feeds `preserveHubOwnedMetadata` as a write base. Confirmed the store-level merge only carries forward parse/routing/resume fields, so the handler-side preserve reading a ≤TTL-stale snapshot was indeed the only line of defense — 9b4d197, with a regression test that warms the memo, lands a hub-owned key via a hub-side write inside the window, and asserts it survives the next client metadata update.
- **[Minor] Avoid retaining every variable-length SQL statement** → per-database retention capped at 512 entries (all current call sites pass one of ~80 static strings, so the cap is defensive); beyond the cap statements are prepared per call — b4e2aaf, with a flood-past-the-cap test.

miuiadmin:
Addressed both follow-up findings:

**[Major] Invalidate cached access when a session is deleted** — `deleteSession` now bumps a process-wide monotonic epoch (`store/sessionInvalidation.ts`); memo entries stamp the epoch they were filled under and a mismatch forces one re-resolve on the next event. Every deletion path (webapp delete, session merge, codex desktop) funnels through the store's `deleteSession`, so all of them invalidate. The hit path gains a single integer compare — no extra DB read, bumps are rare, so the hot-path win is unchanged.

**[Minor] Read current capabilities for native queue events** — `native-queue-message` and `session-end` read `capabilities.concurrentClients` to decide whether the queue ledger is written/swept; both now resolve the session fresh (same treatment `update-metadata` already gets). Both events are human-paced, not stream hot paths, so the fresh read is free relative to the memo's purpose.

Tests: added one per fix (deletion drops a warm memo mid-TTL and denies; native-queue picks up a capabilities write that landed after the memo was warmed). Full suite passes, `tsc --noEmit` clean.

